PR #18, opened 2021-10-01
Feat: Throw Exception when reading not qbeast formatted data.

### LINKED ISSUE #13: Throw new exception when loading non qbeast-formatted data
**What went wrong?**
Trying to load data not stored in qbeast format currently throws a `java.lang.NullPointerException`. We could add a new Exception to be thrown in that case, which tells the user something like "This is not qbeast-formatted data".

**How to reproduce?**
1. Code that triggered the bug, or steps to reproduce:
```scala
val df = spark.read.format("qbeast").load(nonQbeastFormatPath)
```

2. Branch and commit id:
`main` branch, 15667c2

3. Spark version:
3.1.1

4. Hadoop version:
2.7.4

5. Are you running Spark inside a container? Are you launching the app on a remote K8s cluster? Or are you just running the tests in a local computer?
N/A

6. Stack trace:
```
java.lang.NullPointerException
  at io.qbeast.spark.sql.qbeast.QbeastSnapshot.<init>(QbeastSnapshot.scala:35)
  at io.qbeast.spark.table.IndexedTableImpl.snapshot(IndexedTable.scala:151)
  at io.qbeast.spark.table.IndexedTableImpl.load(IndexedTable.scala:146)
  at io.qbeast.spark.sql.sources.QbeastDataSource.createRelation(QbeastDataSource.scala:94)
  at org.apache.spark.sql.execution.datasources.DataSource.resolveRelation(DataSource.scala:354)
  at org.apache.spark.sql.DataFrameReader.loadV1Source(DataFrameReader.scala:326)
  at org.apache.spark.sql.DataFrameReader.$anonfun$load$1(DataFrameReader.scala:306)
  at scala.Option.map(Option.scala:230)
  at org.apache.spark.sql.DataFrameReader.load(DataFrameReader.scala:266)
  at org.apache.spark.sql.DataFrameReader.load(DataFrameReader.scala:240)
  ... 47 elided
```

eavilaes:
This PR resolves #13.
After calling `tableFactory.getIndexedTable()` we must check if the table really exists:
https://github.com/Qbeast-io/qbeast-spark/blob/4f37b584065f99b48e848eeaaa6304bc92223feb/src/main/scala/io/qbeast/spark/table/IndexedTable.scala#L85-L86

### DIFF
--- a/src/main/scala/io/qbeast/spark/sql/sources/QbeastDataSource.scala
+++ b/src/main/scala/io/qbeast/spark/sql/sources/QbeastDataSource.scala
@@ -91,7 +91,11 @@ class QbeastDataSource private[sources] (private val tableFactory: IndexedTableF
       parameters: Map[String, String]): BaseRelation = {
     val path = getPath(parameters)
     val table = tableFactory.getIndexedTable(sqlContext, path)
-    table.load()
+    if (table.exists) {
+      table.load()
+    } else {
+      throw AnalysisExceptionFactory.create(s"'$path' is not a Qbeast formatted data directory.")
+    }
   }
 
   private def getPath(parameters: Map[String, String]): String = {


osopardo1:
Probably we need to change the test for this solution. Loading a table that does not exist should throw an error.

PR #20, opened 2021-10-04
Store metadata in the schema of columns that are indexed

### LINKED ISSUE #2: Investigate a way of saving columnsToIndex in schema or table metadata
**What went wrong?**
As Eric pointed out, there's no easy way to retrieve the columns indexed on a dataframe. Even if you know the insights, you still have to look through the commit log to know which are the ones selected. 
Either if they are chosen manually (with `columnsToIndex` option) or automatically (in the future), user and developer should consult the information in a simpler manner.

eavilaes:
Resolves #2: We need an easy way to obtain the indexed columns from a DataFrame.

We can store this information in the metadata field of the "columnsToIndex" columns. My concern with this solution is the modification of the attribute `data` from the QbeastWriter class. This attribute was originally a `val`, while with this solution I had to change it to a `var` to be able to modify the schema. I have the feeling this is not a good practice, so I would appreciate your comments about it.

With this solution, you can check the metadata for every column:
```scala
df.schema.foreach(field => println(s"${field.name}: metadata=${field.metadata}"))

/*
 foo: metadata={}
 bar: metadata={"isQbeastIndexedColumn":true}
*/
```

### DIFF
--- a/src/main/scala/io/qbeast/spark/sql/qbeast/QbeastWriter.scala
+++ b/src/main/scala/io/qbeast/spark/sql/qbeast/QbeastWriter.scala
@@ -14,6 +14,7 @@ import org.apache.spark.sql.delta.{DeltaLog, DeltaOptions, OptimisticTransaction
 import org.apache.spark.sql.execution.datasources.OutputWriterFactory
 import org.apache.spark.sql.execution.datasources.parquet.ParquetFileFormat
 import org.apache.spark.sql.functions.col
+import org.apache.spark.sql.types.MetadataBuilder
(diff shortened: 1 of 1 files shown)


cugni (changes requested):
the filed data can still be final, we just need to create a new apply method with the companion object. Also, to add the columns in the method write can be dangerous, as we might add the metadata multiples times.

osopardo1:
Converting to Draft Pull Request due to reworking on few things: 

1. Add the SpaceRevision information to the `Metadata` of the table instead of saving it on each record (only maintain an identifier there)
2. Investigate `stats` struct for giving extra statistics information to the `add` files

PR #21, opened 2021-10-05
19 delta log should use relative paths instead of absolute

alexeiakimov:
This PR makes the BlockWriter to use relative names for the data files.

In addition it also changes QbeastSnapshotTest, because it was failing on my machine due to rounding errors while merging the weight estimations of the partitions.


cugni (approved):
All good! I also saw that you tried to align the description of the parameters. It looks better, I guess we should do it everywhere then, no?

alexeiakimov:
I just used mechnically scalafmtAll in sbt console. And I like this style, i.e code first then format with the build tool. We also used the same approach in qbeast-keeper.

osopardo1 (approved):

Issue #28, opened 2021-10-14
Filter Pushdown on indexed columns


osopardo1:
Right now we are using the **whole space for querying**. 

https://github.com/Qbeast-io/qbeast-spark/blob/7edbfc61e64b5bf687098350440c4c9405130ee5/src/main/scala/io/qbeast/spark/sql/files/OTreeIndex.scala#L92

https://github.com/Qbeast-io/qbeast-spark/blob/7edbfc61e64b5bf687098350440c4c9405130ee5/src/main/scala/io/qbeast/spark/sql/files/OTreeIndex.scala#L94

https://github.com/Qbeast-io/qbeast-spark/blob/7edbfc61e64b5bf687098350440c4c9405130ee5/src/main/scala/io/qbeast/spark/sql/files/OTreeIndex.scala#L100

This means that every search starts on the top of the tree and goes down until it finds all cubes that match the `From` / `To` precision. But if the user is interested only in some part of the space, this is filtered on memory instead of skipping the files before.

For example, imagine a table indexed with columns `col1` and `col2`. A query like:

```sql
SELECT * FROM qbeast_table WHERE col1 > 2 and col1 < 7 and col2 > 3 and col2 < 10
``` 

will read all data and then filter by the column value. However, the index should allow filtering cube/blocks/files that have data only on the (2, 7),(3,10) range. 

We should add two main functionalities:
- Push down filters `col1 > 2 and col1 < 7 and col2 > 3 and col2 < 10` 
- Processing ranges of query space

osopardo1:
They are all present as a single Expression:

`DataFilters: [(((user_id#1762 >= 537764969) OR ((user_id#1762 < 666666666) AND (product_id#1757 >= 6789009)))`

PR #35, opened 2021-10-26
Filter cubes by query range 

osopardo1:
This PR is for solving #28.

We need to ensure that the user filters get processed by the read algorithm, so we can filter out those cubes that are not in the query domain.
For that:
- Add a new container class `QbeastFile` with the main properties (metadata) of the parquet file written with qbeast format. 
- Add a new container class `QuerySpec` with the main properties of the query.
- Add a new class `QuerySpecBuilder` to transform/build the `QuerySpec` from the sequence of `Expression` received by Spark.
- Add new classes `QueryExecutor` and `QueryIndexStatusExecutor` which are in charge to execute a query against a `QbeastSnapshot` or an `IndexStatus`. 
This allows the code of the **query** to be decoupled of `delta` packages. `QueryExecutor` is instantiated by the `OTreeIndex` and uses `QbeastFile` as an abstraction of the file metadata. 
- New tests are added


src/main/scala/io/qbeast/spark/sql/utils/QbeastExpressionUtils.scala
```diff
@@ -0,0 +1,163 @@
...
+package io.qbeast.spark.sql.utils
+
+import io.qbeast.spark.index.Weight
+import io.qbeast.spark.model.{Point, QuerySpace, QuerySpaceFromTo, Revision}
+import io.qbeast.spark.sql.rules.QbeastMurmur3Hash
+import org.apache.spark.sql.SparkSession
+import org.apache.spark.sql.catalyst.expressions._
+import org.apache.spark.sql.types._
+
+object QbeastExpressionUtils {
```
alexeiakimov, line 13: Scaladocs?

src/main/scala/io/qbeast/spark/sql/files/RevisionFiltering.scala
```diff
@@ -0,0 +1,99 @@
...
+   * @param files the data files
+   * @param cubeWeights the cube weights
+   * @return the files with sample data
+   */
+  def findSampleFiles(
+      space: QuerySpace,
+      fromWeight: Weight,
+      toWeight: Weight,
+      startCube: CubeId,
+      files: Vector[AddFile],
```
alexeiakimov, line 67: Vector or IndexedSeq?

src/main/scala/io/qbeast/spark/sql/utils/QbeastExpressionUtils.scala
```diff
@@ -0,0 +1,166 @@
...
+      expression: Expression,
+      indexedColumns: Seq[String],
+      spark: SparkSession): Boolean =
+    isQbeastWeightExpression(expression) || hasQbeastColumnReference(
+      expression,
+      indexedColumns,
+      spark)
+
+  /**
+   * Analyzes the data filters from the query
```
alexeiakimov, line 45: Analyzes or extracts?

src/main/scala/io/qbeast/spark/sql/files/RevisionFiltering.scala
osopardo1: Following actual implementation, it is Vector. Once #39 is in place we are going to change it.

src/main/scala/io/qbeast/spark/sql/utils/QbeastExpressionUtils.scala
osopardo1: Sorry, extracts :+1:

cugni (changes requested):
This needs to be aligned with the current version of the main. There are some classes that changed, and also the overall code organization is different (e.g. different packages, different class names). It should be merged with the main before we can review it.

osopardo1:
> This needs to be aligned with the current version of the main. There are some classes that changed, and also the overall code organization is different (e.g. different packages, different class names). It should be merged with the main before we can review it.

Already merged! Give it a look whenever you can :pray:

core/src/main/scala/io/qbeast/core/model/QuerySpace.scala
```diff
@@ -45,17 +45,10 @@ case class AllSpace(dimensionCount: Int) extends QuerySpace {
...
+ * @param to   exclusive ending query range
  */
-case class QuerySpaceFromTo(originalFrom: Point, originalTo: Point, revision: Revision)
-    extends QuerySpace {
-
-  require(originalFrom <= originalTo, "from point must be < then to point")
-  require(originalFrom.dimensionCount == originalTo.dimensionCount)
-  val from: Point = Point(revision.transform(originalFrom.coordinates))
-  val to: Point = Point(revision.transform(originalTo.coordinates))
+case class QuerySpaceFromTo(from: Point, to: Point) extends QuerySpace {
```
cugni, line 51: we shouldn't expose from and to point. QuerySpace should not expose any detail about how we implement the intersectsWith methods. 

Second point, using the Point class is sub-optimal, as we might have 1 or many dimensions unbounded and using from is inconvinient. We should think about a better private structure to use, something that use either optional filtering or something similar

src/main/scala/io/qbeast/spark/utils/QbeastExpressionUtils.scala
```diff
@@ -0,0 +1,147 @@
...
+  LessThan,
+  Literal,
+  SubqueryExpression
+}
+import org.apache.spark.sql.types._
+
+/**
+ * Object with utility methods to work with Spark expressions.
+ */
+object QbeastExpressionUtils {
```
cugni, line 24: What's the goal of this utility class? If this class converts the SQL predicate to QuerySpace, we should use a more meaningful name. Also, if it is an internal class, better to limit its visibility.

cugni (changes requested):
I think we should simplify the QuerySpace interface (we do not need from to, for example), and also move the findSampleFiles algorithm to another class, eg QueryEngine or something similar.

cugni:
@alexeiakimov the implementation of findSampleFiles is recursive, could you rewrite it in an iterable and hopefully faster way? thanks! ☺️

src/main/scala/io/qbeast/spark/utils/QbeastExpressionUtils.scala
osopardo1: Moved to QuerySpec class on ae8b2c829e180b48870f0fc62ef32d3cf5769dfe

core/src/main/scala/io/qbeast/core/model/QuerySpace.scala
osopardo1: Made some changes to the creation of the QuerySpace on last commits ae8b2c829e180b48870f0fc62ef32d3cf5769dfe 2acdae12b22ada61689ddc98bad4b773f0f08801

src/test/scala/io/qbeast/spark/utils/QueryExecutorTest.scala
```diff
@@ -0,0 +1,140 @@
...
+    val allFiles = allDeltaFiles.map(a => QbeastFile(a.path, a.tags))
+
+    val matchFiles = queryExecutor.execute(allFiles)
+
+    matchFiles.size shouldBe <(allFiles.length)
+    matchFiles.foreach(file => allFiles should contain(file))
+
+  })
+
+  it should "find all files in different revisions" in withSparkAndTmpDir((spark, tmpdir) => {
```
cugni, line 109: I don't understand how to this works. The function writeTestData is set to overwrite the dataset, so it should delete the old one, therefore we should have only 1 revision and only the last files. How is it possible this test was even working?

cugni (changes requested):
We have to review how we manage the revision, both in the indexing lifecycle, but also if we overwrite a dataset. If we overwrite a dataset, we should start from RevisionID 1, right?

src/test/scala/io/qbeast/spark/utils/QueryExecutorTest.scala
osopardo1: Thanks for noticing. I think it's a problem with the Metadata update, I will check it

core/src/main/scala/io/qbeast/core/model/QuerySpace.scala
```diff
@@ -45,24 +32,48 @@ case class AllSpace(dimensionCount: Int) extends QuerySpace {
...
  * (inclusive, exclusive).
  *
- * @param originalFrom inclusive starting range
- * @param originalTo   exclusive ending query range
- * @param revision revision applied on this space
+ * @param from inclusive starting range
+ * @param to   exclusive ending query range
  */
-case class QuerySpaceFromTo(originalFrom: Point, originalTo: Point, revision: Revision)
+case class QuerySpaceFromTo(from: Seq[Option[Double]], to: Seq[Option[Double]])
```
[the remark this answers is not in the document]
osopardo1: I don't think so. Neither we do any pattern matching on it. Thank you!

src/test/scala/io/qbeast/spark/utils/QueryExecutorTest.scala
osopardo1: Ok, have seeing the changes now. I will take a look to the lifecycle of revision.

I agree on the revision ID, and of course, overwrite should drop all previous configuration. Delta maintains the schema and everything, but better to drop Qbeast keys before updating.  
In any case, RevisionChanges should notice the overwrite mode as well and set the ID to 1.

src/main/scala/io/qbeast/spark/index/OTreeDataAnalyzer.scala
```diff
@@ -182,35 +180,35 @@ object DoublePassOTreeDataAnalyzer extends OTreeDataAnalyzer with Serializable {
 
   override def analyze(
       dataFrame: DataFrame,
-      oldIndexStatus: IndexStatus,
+      indexStatus: IndexStatus,
```
osopardo1, line 183: The use of `indexStatus` in the initialization of `IndexStatusChanges`, causes conflicts when we append a new revision. 

When we add new data and the revision changes, the previous `CubeWeights` to merge should be **empty**. In this case, as we use the superseded index status to do the merge in `IndexStatusChanges`, CubeWeights has data on it.  Then we are merging two index structures with different revisions. 

This was the reason for using 

```
// The indexStatus to use
    val indexStatus = spaceChanges match {
      case Some(revisionChange) => IndexStatus(revisionChange.newRevision)
      case None => oldIndexStatus
    }
```

cugni: But that part is a mistake. You should not create a new revision and pack as a superseded one. The idea of the revision classes is to make clear the life cycle of the meta, and that was a hack. I also saw that in the code, we recreate the new revision or index status multiple times, this can cause many problems. We have to review all the uses of these classes and make sure the way to use them is cristal clear for everybody.

cugni (approved):

PR #51, opened 2021-12-08
Avoiding Out of memory when running the Quick start

cugni:
We were experiencing some Out of Memory errors caused by the CubeWeightBuilder. This might relate to issue #9.
The CubeWeightBuilder keeps in memory a priority queue of all entries when it scans the data, and then generates the resulting list of NormalizedWeights only at the end, once the result() method is called. However, suppose Spark indexes a dataset with a low number of parallelism. In that case, we might have a single instance of the CubeWeightBuilder running in a single mapPartition, and this causes an OOM error. 

As a simple solution, I've created a new class, GroupedCubeWeightsBuilder, which takes an additional parameter, maxGroupSize. When the buffer/priority queue in memory reaches the threshold size, I call the method result internally, clear the queue and keep only the final result, instead of the buffer. ( If you check the code, it is probably clearer 😅)

Ideally, having smaller maxGroupSize equals having more partition, should it should not influence the correctness of the index, but of course, might worsen the quality of the index structure estimation.
You can configure the maxGroupSize with the parameters qbeast.index.maxGroupSize  is set by default at 10 000 000. 

We are adding a lot of parameters to fix this part of the code, with the boostSize and the maxGroupSize. They all interconnect, even though they have different effects in the execution and the index result. We should study a better way to mange these things. 

To reproduce the error I run this code:
```bash
 $SPARK_HOME/bin/spark-shell --master "local[4]" \
--conf spark.driver.memory=1g --driver-java-options="-Dqbeast.index.maxGroupSize=100000" --jars ./target/scala-2.12/qbeast-spark-assembly-0.2.0.jar \
--conf spark.hadoop.fs.s3a.aws.credentials.provider=org.apache.hadoop.fs.s3a.AnonymousAWSCredentialsProvider \
--packages io.delta:delta-core_2.12:1.0.0,\
com.amazonaws:aws-java-sdk:1.12.20,\
org.apache.hadoop:hadoop-common:3.2.0,\
org.apache.hadoop:hadoop-client:3.2.0,\
org.apache.hadoop:hadoop-aws:3.2.0 -I ~/testOOM.scala

```
where testOOM.scala is
```scala
(spark.read.format("parquet").load( "s3a://qbeast-public-datasets/store_sales").na.drop()
    .write
    .mode("overwrite")
    .format("qbeast")     // Saving the dataframe in a qbeast datasource
    .option("columnsToIndex", "ss_cdemo_sk,ss_cdemo_sk")      // Indexing the table
    .option("cubeSize", 300000) // The desired number of records of the resulting files/cubes. Default is 100000
    .save("/tmp/qbeast-test-data/qtable"))

System.exit(0)
```
With the main branch it fails on my mac book air m1 16gb. In the new branch, if fails if I set a maxGroupSize too high, but if a smaller value (like 100000), it works. 

Things to do:

- [x] decide if use the GroupedCubeWeightsBuilder, or just put the same logic directly in the CubeWeightsBuilder
- [x] The maxBlockSize also influences the boostSize somehow, we should understand how to managed them together in a way that makes sense. 
- [x] Find a reasonable default value
- [x] Should this parameter part of the option API?
- [x] Add tests 
- [x] Document how to configure the trade-off between less memory pressure or better index prediction.


core/src/main/scala/io/qbeast/core/model/CubeWeights.scala
```diff
@@ -78,6 +86,50 @@ class CubeWeightsBuilder(
...
+   *               the container cube if available
+   * @return this instance
+   */
+  override def update(
+      point: Point,
+      weight: Weight,
+      parent: Option[CubeId]): CubeWeightsBuilder = {
+    super.update(point, weight, parent)
+    groupCount += 1
+    if (groupCount > maxGroupSize) {
```
alexeiakimov, line 117: > or >=?

src/main/resources/reference.conf
```diff
@@ -7,5 +7,6 @@ qbeast {
     index {
         defaultCubeSize = 100000
         minPartitionCubeSize = 100
+        maxGroupSize = 100000
```
alexeiakimov, line 10: This is name cryptic in the config context.

alexeiakimov:
General opinionated comment on class design:
It seems that the memory problem will relevant forever, thus I can hardly imagine that in some cases we choose a grouped implementation and in other cases we choose a simple one. It means we do not need both classes to have the same interface or to be subclasses. I would use a slightly different approach, I would keep the original name and interface for the new class, and I would rework the existing one like the following:
```scala
class CubeWeightsBuffer(capacity: Int) {
  private val queue = ...

  def add(point: Point,  weight: Weight, parent: CubeId) {
  ...
  }

  def isFull: Boolean = {
      queue.size() >= capacity;
  }

  def flush(): Map[CubeId, Weight] = {
      ...
      queue.clear()
      result
  }
}
```
Then
```scala
CubeWeightsBuilder(bufferCapacity: Int) {
  private val buffer = CubeWeightsBuffer(bufferCapacity)
  private val result = mutable.Map.empty()

  def update(point, weight, parent): CubeWeightBuilder = {
      buffer.add(point, weight, parent)
      if (buffer.isFull) {
        result ++= buffer.flush()
      } 
      this
  }

  def build(): ... = {
    result ++= buffer.flush()
  }
}
```
In the config the property can be names as `cubeWeightsBufferCapacity`.

alexeiakimov:
General opinionated comment on naming:
Prefix `Grouped` reveals how the builder is implemented. I believe we do not need to have two implementations if builder, so the prefix can be dropped. Even if we need prefix, then `Buffered` sounds better for me because the way how the data is processed looks like classic buffering in streams. The last but not least, the name of the config property should clearly point to the CubeWeightBuilder or one of its internal mechanisms, simple `maxGroupSize` is too general when apears in the config.

core/src/main/scala/io/qbeast/core/model/CubeWeights.scala
eavilaes: Solved in f881e18

src/main/resources/reference.conf
eavilaes: Changed it to `cubeWeightsBufferCapacity` in f881e18

src/main/scala/io/qbeast/spark/index/OTreeDataAnalyzer.scala
```diff
@@ -46,25 +48,30 @@ object DoublePassOTreeDataAnalyzer extends OTreeDataAnalyzer with Serializable {
...
-          "The DataFrame is empty, why are you trying to index an empty dataset?")
+    val columnNames = revision.columnTransformers.map(_.columnName)
+
+    val mins = dataFrameStats.filter(_.get(0) == "min").head.getValuesMap(columnNames)
+    val maxs = dataFrameStats.filter(_.get(0) == "max").head.getValuesMap(columnNames)
+
+    val newTransformation = revision.columnTransformers.map(ct => {
+      val minVal: String = mins(ct.columnName)
+      val maxVal: String = maxs(ct.columnName)
+      val columnStats: ColumnStats = {
```
osopardo1, line 62: I think it would be useful to build these ColumnStats from the companion object. Also, instead of initializing min and max values outside the map, we can retrieve all the stats for the column at the same time

It would be something like:

```scala
val newTransformation = revision.columnTransformers.map(ct => {
      val stats =
        dataFrameStats.map(row => (row.getAs[String](0), row.getAs[String](ct.columnName))).toMap // a map to get the stats by their name
      val columnStats = ColumnStats(stats, ct.dataType) // here in the apply method we can do all this matching with the data types
      ct.makeTransformation(columnStats)
    })
```

osopardo1: Having the `stats` in `Map[String, String] ` can help to get the information we want through: 

```scala
val min = stats("min")
val max = stats("max")
```

And we can as well collect the other properties like `stddev`, `mean` and `count`

osopardo1:
A little update on the tasks:

- _Decide if use the GroupedCubeWeightsBuilder, or just put the same logic directly in the CubeWeightsBuilder_ 
We use the same logic inside the CubeWeightsBuilder

- _The maxBlockSize also influences the boostSize somehow, we should understand how to managed them together in a way that makes sense._
The next formulas are used in order to manage the different parameters. 
  - `numGroups = MAX(numPartitions, (numElements / cubeWeightsBufferCapacity))` . This defines the number of groups in which we divide the dataset to compute the CubeWeights
  - `groupCubeSize = desiredCubeSize / numGroups`. This defines the number of elements in each of the groups

- Find a reasonable default value for the different variables.
Here is a glossary on the default values:
  - `defaultCubeSize = 100000`  The default cube size for all datasets written with the application. If the user does not explicitly define any "cubeSize" in the options, this value is used. 
  - `minPartitionCubeSize = 100` . The minimum cube size per partition allowed for the distributed computation of CubeWeights
  - `cubeWeightsBufferCapacity = 100000`. The maximum capacity of the buffer used to store the intermediate results of the CubeWeightsBuilder. We have seen that bigger numbers than 100000 causes memory errors

- _Should this parameter be part of the options API?_ 
Since they are advanced configurations, we decided to keep them as internal and don't expose them to the options API on the DataFrame.

src/main/scala/io/qbeast/spark/index/OTreeDataAnalyzer.scala
```diff
@@ -115,18 +122,21 @@ object DoublePassOTreeDataAnalyzer extends OTreeDataAnalyzer with Serializable {
...
-      val desiredPartitionCubeSize =
-        Math.ceil(desiredCubeSize.toDouble / numPartitions + 1) // Round to the next value
-      if (desiredPartitionCubeSize < minPartitionCubeSize) {
+      numPartitions: Int,
+      numElements: Long,
+      maxWorkingSetSize: Long): Double = {
+    val numGroups = Math.max(numPartitions, numElements / maxWorkingSetSize)
+    if (numGroups > 0) {
+      val groupCubeSize =
+        Math.ceil(desiredCubeSize.toDouble / numGroups + 1)
```
osopardo1, line 133: Here the `groupCubeSize` is set as `desiredCubeSize / numGroups + 1`

In case we did not round this number, the next situation happens:

1. We index a dataframe of `10000` elements and we set the cubeSize to `10000`.
2. The groupCubeSize is `10000`
3. We end up having one cube with `9998` elements and another one with `2`. Because the CubeWeights only outputs a single cube with max weight 0.999.. and when storing the elements, the ones in the range [0.999, 1.0] end up in the children.

My question is: What is the behavior expected? To have 2 cubes or to store all the data in a single one if the cube size matches the size of the dataset?

@cugni @alexeiakimov any comment on this would be helpful!

src/main/scala/io/qbeast/spark/index/OTreeDataAnalyzer.scala
osopardo1: Solved in b9cff7b30c0f53bcf067c451d9455a98fe1ab68e

src/main/scala/io/qbeast/spark/index/OTreeDataAnalyzer.scala
cugni: It is a target cube size, it is a "soft limit", or better a recommendation, we do not have to comply strictly with it.

build.sbt
```diff
@@ -37,6 +37,9 @@ ThisBuild / libraryDependencies ++= Seq(
   scalaTest % Test,
   mockito % Test)
 
+Test / javaOptions += "-Xmx2G"
```
osopardo1, line 40: Why are we testing with 2GB? Is it for the Documentation? In this way we may not detect memory problems, right?

cugni: The point is to set a value that will be used in any platform, so the tests are run in a similar environment. If you want to lower it, it's ok, but check if it works. The problem usually is the documentation tests anyway.

osopardo1: It seems ok to me!

osopardo1 (approved):

PR #55, opened 2021-12-14
Using Spark Config to manage Qbeast configuration

cugni:
At the moment we are using Typesafe Conf to set some configuration parameters, but this causes a weird situation for which we use one mechanism to set up the configuration for spark and other libraries, and a different one for Qbeast. There is no reason for this, and we should use a standard approach. 
Also, we would like to be able to pass arguments in spark-submit or shell like
```
--conf spark.qbeast.something=10
```
instead of
```
--spark.driver.extraJavaOptions="-Dio.qbeast.something=10"
```

To do
- [x] create a package that contains our settings and exports them to our code
- [x] migrate all configuration 
- [x] Add test


osopardo1:
Just one comment before closing the PR. We didn't migrate the keeper configuration anywhere because it wasn't needed for a LocalKeeper. We just make sure it starts with the "spark" prefix!

osopardo1 (approved):

src/main/scala/io/qbeast/context/QbeastContext.scala
```diff
@@ -65,7 +65,7 @@ object QbeastContext
 
   // Override methods from QbeastContext
 
-  override def config: Config = current.config
```
eavilaes, line 68: Shouldn't it be `current.config`, as it was previously?

eavilaes (commented):
I had some troubles testing with your changes. To put you in context, I've tried to merge this with #51 to keep testing new functionalities, but the SparkSession for the tests was not being created.
I think the reason for the failing tests was the following change, could you please check it?

src/main/scala/io/qbeast/context/QbeastContext.scala
osopardo1: It is a possibility! Have you tried with this modification?

osopardo1: Nein, I just did it and the same error appears :cry:

osopardo1: Change it to current.config in 888da87

osopardo1 (approved):

eavilaes (approved):

PR #60, opened 2021-12-22
Add check to publish only when repository is "Qbeast-io/qbeast-spark"

### LINKED ISSUE #52: CI failing on repository Forks, due to GitHub Package Registry key missing on them
**What went wrong?**
When a fork of this project is synchronized, CI workflows always fail. The reason is that the CI tries to publish the artifacts, while the GitHub Package Registry (GHPR) key is only available in this repository.
You can see an example of failing workflow in my fork here: https://github.com/eavilaes/qbeast-spark/runs/4468737606?check_suite_focus=true
This can be annoying for all users who fork the project.

I suggest fixing this step from the CI, but I thought about different alternatives:
1. Add a check in the publishing step, which verifies that the current repo is `Qbeast-io/qbeast-spark`. I think this can be done with `${{ github.repository }}`
2. Enable this step as `workflow_dispatch`-triggered; which will let the workflow to be executed manually.
3. Publish the artifact manually, apart from the CI.

**How to reproduce?**
1. Code that triggered the bug, or steps to reproduce:
N/A

2. Branch and commit id:
main, d9bd04a

3. Spark version:
N/A

4. Hadoop version:
N/A

5. Are you running Spark inside a container? Are you launching the app on a remote K8s cluster? Or are you just running the tests in a local computer?
N/A

6. Stack trace:
N/A

eavilaes:
Adding a simple check for the repository name to avoid running the `sbt publish` on the repository forks' CI. The syntax used is the one from [GitHub Actions Quickstart](https://docs.github.com/en/actions/quickstart).

Fixes #52

### DIFF
--- a/.github/workflows/test-and-publish-artifact.yml
+++ b/.github/workflows/test-and-publish-artifact.yml
@@ -25,7 +25,7 @@ jobs:
       run: sbt scalafmtSbtCheck doc
 
     # Publish only if it is on the main branch
-    - if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/main' }}
+    - if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/main' && github.repository == 'Qbeast-io/qbeast-spark' }}
       name: Publish artifact
       run: |
         QBEAST_SPARK_VERSION="nightly-$(git rev-parse --short ${GITHUB_SHA})"


osopardo1:
Seems good to me!

osopardo1 (approved):

eavilaes:
I can confirm it has been fixed. Publish step is not being executed on forks:
https://github.com/eavilaes/qbeast-spark/runs/4607461599?check_suite_focus=true

Issue #53, opened 2021-12-09
Unexpected exception when reading non-qbeast-formatted data


eavilaes:
**What went wrong?**
The following exception should be thrown when you load data that is not in qbeast format or when the path does not exist. It works well when the path does not exist; however, a different exception is thrown when the path exists and it is non-qbeast-formatted data:
https://github.com/Qbeast-io/qbeast-spark/blob/d9bd04aa17f60b2ca3e2f7143193e931ab40d389/src/main/scala/io/qbeast/spark/internal/sources/QbeastDataSource.scala#L89-L94

My conclusion is that the format of the table is not checked, as this happens as well when trying to load a table indexed with an old version of the qbeast-spark format.

**How to reproduce?**
1. Code that triggered the bug, or steps to reproduce:
- Load an empty path:
```scala
val df = spark.read.format("qbeast").load("nonExistingPath")
org.apache.spark.sql.AnalysisException: 'nonExistingPath' is not a Qbeast formatted data directory.
```
- But try to load a `delta`-formatted table or a `qbeast` table written with the old version of the format, and the exception will refer to the revision:
```scala
val df = spark.read.format("qbeast").load("deltaTablePath")
org.apache.spark.sql.AnalysisException: No space revision available with -1
```

2. Branch and commit id:
main, d9bd04a

3. Spark version:
3.1.1

4. Hadoop version:
3.2.0

5. Are you running Spark inside a container? Are you launching the app on a remote K8s cluster? Or are you just running the tests in a local computer?
N/A

6. Stack trace:
```scala
val df = spark.read.format("qbeast").load("deltaTablePath")
org.apache.spark.sql.AnalysisException: No space revision available with -1
  at org.apache.spark.sql.AnalysisExceptionFactory$.create(AnalysisExceptionFactory.scala:36)
  at io.qbeast.spark.delta.DeltaQbeastSnapshot.$anonfun$getRevision$1(DeltaQbeastSnapshot.scala:81)
  at scala.collection.immutable.Map$EmptyMap$.getOrElse(Map.scala:104)
  at io.qbeast.spark.delta.DeltaQbeastSnapshot.getRevision(DeltaQbeastSnapshot.scala:81)
  at io.qbeast.spark.delta.DeltaQbeastSnapshot.loadLatestRevision(DeltaQbeastSnapshot.scala:140)
[...]

PR #62, opened 2022-01-10
Implement data format versioning

### LINKED ISSUE #44: Implement data format versioning
**What went wrong?**
During the development and cleaning of the software, we missed having a version control on each of the implementations. 
Without this track, in the main branch, we have two different ways of writing the metadata with no possible hint to differentiate between the releases. 

For example, one dataset written before 7eb77dd2b192bef9fad796831e382393959f72b0 would contain this information:

```json
{
  "add" : {
    "..." : {},
    "tags" : {
      "cube" : "A",
      "indexedColumns" : "ss_sales_price,ss_ticket_number",
      "maxWeight" : "462168771",
      "minWeight" : "-2147483648",
      "rowCount" : "508765",
      "space" : "{\"timestamp\":1631692406506,\"transformations\":[{\"min\":-99.76,\"max\":299.28000000000003,\"scale\":0.0025060144346431435},{\"min\":-119998.5,\"max\":359999.5,\"scale\":2.083342013925058E-6}]}",
      "state" : "FLOODED"
    }
  }
}
```

While the new ones have a different structures:

```json
{
  "add": {
    "path": "4b36340e-0bf7-44ec-97da-f7a03bf06ea3.parquet",
    ...
    "tags": {
      "state": "FLOODED",
      "rowCount": "3",
      "cube": "gA",
      "revision": "1634196697656",
      "minWeight": "-2147483648",
      "maxWeight": "-857060062"
    }
  }
}
```

We should: [EDITED]

- [ ] Add the File Format Version in the Delta Log. Version should be an increasing number starting from 1
- [ ] Check if we are reading the right version when loading a new DataFrame
- [ ] Update the docs, listing the versions and tagging the latest version of the code that can read it.

eavilaes:
As noted in #44, we need to implement a data format versioning system. We can do this by specifying a Protocol Version, which will be an increasing number starting from 1, which will uniquely identify the file's metadata version.

To achieve this, I approached two different alternatives:
-  **Use DeltaLog's `protocol` tag** to store information about our protocol versioning. (**it did not work**, but I left the commits in this PR so you can get an idea of what I tried, in case you want to keep trying🙂).
> You can find information at [Action Reconciliation](https://github.com/delta-io/delta/blob/a5cbeb16e7cc1bef5d7371c88e391a681b313bfd/PROTOCOL.md#action-reconciliation) and [Protocol Evolution](https://github.com/delta-io/delta/blob/a5cbeb16e7cc1bef5d7371c88e391a681b313bfd/PROTOCOL.md#protocol-evolution) from Delta can help to understand the solution.
As mentioned in [this comment](https://github.com/Qbeast-io/qbeast-spark/issues/44#issuecomment-1005898584) I tried to modify their `protocol` tag, with no luck due to "Action Reconciliation" reading only the latest protocol tag.

- **Use AddFile metadata** to include a `protocolVersion` tag which specifies the protocol version for each file. This is the result for this PR, which includes the following tasks to be completed:
  - [x] Add the File Format Version in the Delta Log. Version should be an increasing number starting from 1, which will be the old protocol, being 2 the current protocol version.
  - [ ] Check if we are reading the right version when loading a new DataFrame. Throw an exception otherwise.
  - [ ] Update the docs, listing the versions and tagging the latest version of the code that can read it.

Related issues: 
- Fixes #44 
- Solves #53

### DIFF
--- a/src/main/scala/io/qbeast/spark/delta/DeltaMetadataWriter.scala
+++ b/src/main/scala/io/qbeast/spark/delta/DeltaMetadataWriter.scala
@@ -49,17 +49,14 @@ private[delta] case class DeltaMetadataWriter(
    * TODO: create QbeastProtocol, which stores information about the versions for QbeastReaderVersion and
    * QbeastWriterVersion. This information will help to control data versioning.
    */
-  class QbeastProtocol(
-      minQbeastReaderVersion: Int = Action.readerVersion,
-      minQbeastWriterVersion: Int = Action.writerVersion)
+  private[delta] class QbeastProtocol(minQbReaderVersion: Int, minQbWriterVersion: Int)
       extends Protocol {
+    val minQbeastReaderVersion: Int = minQbReaderVersion
+    val minQbeastWriterVersion: Int = minQbWriterVersion
 
-    // Mock method
-    def show(): Unit = {
-      printf(
-        s"minQbeastReaderVersion: ${minQbeastReaderVersion}, " +
-          s"minQbeastWriterVersion: ${minQbeastWriterVersion}\n")
-    }
+    override def simpleString: String =
+      s"($minReaderVersion, $minWriterVersion, " +
+        s"$minQbeastReaderVersion, $minQbeastWriterVersion)"
 
   }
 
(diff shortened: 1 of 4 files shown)


osopardo1:
I will close this PR since it's out of the scope. We will reopen it later!

Thank you for the work done, @eavilaes , don't throw it away! Keep it for the next updates!

PR #65, opened 2022-01-17
Avoid repeated reading of the DeltaLog

### LINKED ISSUE #61: Why do we read twice the delta log when doing a query?
In the current implementation, when we do a query, we read at least twice the delta log: 
1. At first, in the OTreeIndex:33 class, when we call 
```scala
val tahoeMatchingFiles = index.matchingFiles(partitionFilters, dataFilters)
```
on the TahoeLogFileIndex
2. Then in QueryExecutor:28 when we execute
```scala
val indexStatus = qbeastSnapshot.loadIndexStatus(revision.revisionID)
```
 we read the full deltaLog once for each revision. 

Also, we do an intersection of the files in **tahoeMatchingFiles** and the files found by the QueryExecutor twice, one explicitly in OTreeIndex:41
```scala
 tahoeMatchingFiles.filter { a => qbeastMatchingFiles.exists(_.path == a.path) }
```
and also when we do the flatMap in the QueryExecutor:63
```scala
outputFiles ++= files.flatMap(fileMap.get)
```

The questions are:
1. Do we need to filter also for the tahoeIndexFiles? Which are the data predicates that it filters that we don't? 
2. If we need to use the file matched by the TahoeLogFileIndex, why don't we use these files to build the IndexStatus, so we don't have to scan the deltaLog multiple times, and we don't have to intersect the list of files.

osopardo1:
This PR fixes #61 

The changes made are:
- `OTreeIndex` now extends `FileIndex` instead of `TahoeLogFileIndex`
- QueryExecutor doesn't need `previouslyMatchedFiles` anymore
- `QbeastFile` now it's named `QbeastBlock`
- Add more metadata to the `QbeastBlock`: **size** and **modificationTime**
- `CubeStatus` now contains `QbeastBlock` objects instead of the path

### DIFF
--- a/core/src/main/scala/io/qbeast/core/model/QbeastFile.scala
+++ b/core/src/main/scala/io/qbeast/core/model/QbeastFile.scala
@@ -9,6 +9,8 @@ package io.qbeast.core.model
  * @param maxWeight
  * @param state
  * @param elementCount
+ * @param size
+ * @param modificationTime
  */
 case class QbeastFile(
     path: String,
@@ -17,7 +19,9 @@ case class QbeastFile(
     minWeight: Weight,
     maxWeight: Weight,
     state: String,
-    elementCount: Long)
+    elementCount: Long,
+    size: Long,
+    modificationTime: Long)
 
 /**
  * Companion object for QbeastFile
@@ -38,9 +42,15 @@ object QbeastFile {
    * Creates a QbeastFile from a file path and metadata map
    * @param path
    * @param fileMetadata
+   * @param size
+   * @param modificationTime
    * @return
    */
-  def apply(path: String, fileMetadata: Map[String, String]): QbeastFile = {
+  def apply(
+      path: String,
+      fileMetadata: Map[String, String],
+      size: Long,
+      modificationTime: Long): QbeastFile = {
     checkFileMetadata(fileMetadata)
 
     QbeastFile(
@@ -50,7 +60,9 @@ object QbeastFile {
       Weight(fileMetadata("minWeight").toInt),
       Weight(fileMetadata("maxWeight").toInt),
       fileMetadata("state"),
-      fileMetadata("elementCount").toLong)
+      fileMetadata("elementCount").toLong,
+      size,
+      modificationTime)
   }
 
 }
--- a/src/main/scala/io/qbeast/spark/delta/IndexStatusBuilder.scala
+++ b/src/main/scala/io/qbeast/spark/delta/IndexStatusBuilder.scala
@@ -4,7 +4,7 @@
 package io.qbeast.spark.delta
 
 import io.qbeast.core.model._
-import org.apache.spark.sql.Dataset
+import org.apache.spark.sql.{Dataset, SparkSession}
 
 import scala.collection.immutable.SortedMap
 
@@ -44,30 +44,37 @@ private[delta] class IndexStatusBuilder(
    */
   def buildCubesStatuses: SortedMap[CubeId, CubeStatus] = {
 
+    val spark = SparkSession.active
+    import spark.implicits._
     val rev = revision
     val builder = SortedMap.newBuilder[CubeId, CubeStatus]
     revisionFiles
-      .collect()
-      .toVector
-      .groupBy(_.cube)
-      .map { case (cube, files) =>
+      .groupByKey(_.cube)
+      .mapGroups((cube, f) => {
         var minMaxWeight = Int.MaxValue
         var elementCount = 0L
-        for (file <- files) {
+        val files = Vector.newBuilder[QbeastBlock]
+        for (file <- f) {
           elementCount += file.elementCount
           val maxWeight = file.maxWeight.value
           if (maxWeight < minMaxWeight) {
             minMaxWeight = maxWeight
           }
+          files += file
         }
         val cubeStatus = if (minMaxWeight == Int.MaxValue) {
-          CubeStatus(Weight.MaxValue, NormalizedWeight(rev.desiredCubeSize, elementCount), files)
+          CubeStatus(
+            Weight.MaxValue,
+            NormalizedWeight(rev.desiredCubeSize, elementCount),
+            files.result())
         } else {
           val w = Weight(minMaxWeight)
-          CubeStatus(w, NormalizedWeight(w), files)
+          CubeStatus(w, NormalizedWeight(w), files.result())
         }
-        builder += ((rev.createCubeId(cube), cubeStatus))
-      }
+        (rev.createCubeId(cube), cubeStatus)
+      })
+      .collect()
+      .foreach(builder += _)
     builder.result()
   }
 
--- a/src/main/scala/io/qbeast/spark/delta/IndexStatusBuilder.scala
+++ b/src/main/scala/io/qbeast/spark/delta/IndexStatusBuilder.scala
@@ -26,9 +26,17 @@ private[delta] class IndexStatusBuilder(
    * Dataset of files belonging to the specific revision
    * @return the dataset of AddFile actions
    */
-  def revisionFiles: Dataset[QbeastBlock] =
+  def revisionFiles: Dataset[(String, QbeastBlock)] = {
     // this must be external to the lambda, to avoid SerializationErrors
-    qbeastSnapshot.loadRevisionBlocks(revision.revisionID)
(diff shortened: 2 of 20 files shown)


core/src/main/scala/io/qbeast/core/model/QbeastFile.scala
```diff
@@ -9,6 +9,8 @@ package io.qbeast.core.model
  * @param maxWeight
  * @param state
  * @param elementCount
+ * @param size
+ * @param modificationTime
  */
 case class QbeastFile(
```
cugni, line 15: Maybe call it QbeastBlock, or just Block, as we agreed earlier to do not call the single objects/files files.

src/main/scala/io/qbeast/spark/delta/IndexStatusBuilder.scala
```diff
@@ -28,10 +27,16 @@ private[delta] class IndexStatusBuilder(
...
-  def revisionFiles: Dataset[AddFile] = {
+  def revisionFiles: Dataset[QbeastFile] = {
+    val spark = SparkSession.active
+    import spark.implicits._
+
     // this must be external to the lambda, to avoid SerializationErrors
     val revID = revision.revisionID.toString
-    qbeastSnapshot.snapshot.allFiles.filter(_.tags(TagUtils.revision) == revID)
+    qbeastSnapshot.snapshot.allFiles
+      .filter(_.tags(TagUtils.revision) == revID)
```
cugni, line 37: filtering files by revID, should be part of the functionality of QbeastSnapshot. Or in other words, wouldn't be better if the field snapshot is private in QbeastSnapshot? So we don't rely on Delta's API too much.

src/main/scala/io/qbeast/spark/delta/IndexStatusBuilder.scala
```diff
@@ -48,37 +53,30 @@ private[delta] class IndexStatusBuilder(
...
   def buildCubesStatuses: SortedMap[CubeId, CubeStatus] = {
 
-    val spark = SparkSession.active
-    import spark.implicits._
     val rev = revision
     val builder = SortedMap.newBuilder[CubeId, CubeStatus]
     revisionFiles
-      .groupByKey(_.tags(TagUtils.cube))
-      .mapGroups((cube, f) => {
+      .collect()
```
cugni, line 59: why did you change this? Why are we aggregating in-memory in the master instead of using Spark? What was the need for this?

src/main/scala/io/qbeast/spark/delta/OTreeIndex.scala
```diff
@@ -26,19 +29,40 @@ case class OTreeIndex(index: TahoeLogFileIndex)
...
+      p
+    } else {
+      new Path(index.path, p)
+    }
+  }
 
-    val tahoeMatchingFiles = index.matchingFiles(partitionFilters, dataFilters)
-    val previouslyMatchedFiles =
-      tahoeMatchingFiles.map(addFile => QbeastFile(addFile.path, addFile.tags))
+  def matchingFiles(
```
cugni, line 41: this method should be private, and probably better change its name, so it can't be confused with the TahoeIndex's method.

cugni (changes requested):
I don't understand why we do buildCubesStatuses in memory now, what was the need?
Also, we could change some class/method names to improve readability and be more consistent with the documentation.

osopardo1:
https://github.com/Qbeast-io/qbeast-spark/pull/65#discussion_r792652606

In relation to this comment, I did the following comparisons using different approaches (collect before, collect after and the code in the main branch) in the `store_sales` table of the 10gb TPC-DS dataset. 

```scala
df.write
      .format("qbeast")
      .option("columnsToIndex", "ss_item_sk,ss_customer_sk")
      .option("cubeSize", "300000")
      .save(qbeastDir)
```

The times compared are from two different methods:
- **Query**. I compared running a count on the whole dataset both with Delta and Qbeast implementations
```scala 
val qbeast = spark.read.format("qbeast").load(qbeastDir)
val delta = spark.read.format("delta").load(qbeastDir)

qbeast.count()
delta.count()
```
- **ListFiles**. This is the method used by the TahoeLogFileIndex and OTreeIndex for filtering the files needed to query. It is useful to know if we improved something by reading only once the log
```scala
val deltaLog = DeltaLog.forTable(spark, qbeastDir)
val tahoeLogFileIndex =
  TahoeLogFileIndex(spark, deltaLog, deltaLog.dataPath, deltaLog.snapshot)
val oTreeIndex = OTreeIndex(tahoeLogFileIndex)

tahoeLogFileIndex.listFiles(partitionFilters = Seq.empty, dataFilters = Seq.empty)
oTreeIndex.listFiles(partitionFilters = Seq.empty, dataFilters = Seq.empty)
```
These are the implementations tested:

- Delta → qbeast indexed but read as delta
- Qbeast main branch → qbeast indexed but read with the main version
- Qbeast PR65 commit f2445ec1e58e61b23b5eea25d715b2a0c8831670 → Collect after `groupBy` in `buildCubeStatuses`
- Qbeast PR65 commit 36eea23a00be8f772df60ceb2dd86dd9cc8fec95 → Collect before `groupBy` in `buildCubeStatuses`

|Operation |Delta     |Qbeast Main       |QBeast PR65 Collect After  |Qbeast PR65 Collect Before         |
|------------|:----------:|:----------:|:----------:|:----------:|
|Query     |1.38    |2.94   |2.36   |1.73 |
|ListFiles | 0.18 | 1.4 | 0.9 |0.3 |

The tests were run locally. Feel free to discuss this approach here, and to suggest reversion of the changes! Thank you!

src/main/scala/io/qbeast/spark/delta/IndexStatusBuilder.scala
osopardo1: Yes, we can exclude the IndexStatusBuilder of using delta objects.

src/main/scala/io/qbeast/spark/delta/IndexStatusBuilder.scala
osopardo1: Discussion on https://github.com/Qbeast-io/qbeast-spark/pull/65#issuecomment-1022232484 :smile:

osopardo1:
Made more test on a distributed cluster, just to have more numbers. Here are the results:

`mode`: DISTRIBUTED, `cluster`: 4 node cluster, `data`:10gb store sales, `columnsToIndex`: ss_customer_sk, ss_item_sk, `cubeSize` default , `num files`: 5000+ files

|Operation |Delta     |Qbeast Main       |QBeast PR65 Collect After  |Qbeast PR65 Collect Before         |
|------------|:----------:|:----------:|:----------:|:----------:|
|Query     |8.2s    |10.74s   |10.05s   |10.2s |
|ListFiles | 0.43s | 2.1s | 0.91s |0.53s |

`mode`: DISTRIBUTED, `cluster`: 4 node cluster, `data`:10gb store sales, `columnsToIndex`: ss_customer_sk, ss_item_sk, `cubeSize` 500000 , `num files`: 800+ files

|Operation |Delta     |Qbeast Main       |QBeast PR65 Collect After  |Qbeast PR65 Collect Before         |
|------------|:----------:|:----------:|:----------:|:----------:|
|Query     |1.64s    |2.19s   |3.14s   |2.55s |
|ListFiles | 0.32s | 0.65s | 0.96s |0.42s |


`mode`: DISTRIBUTED, `cluster`: 4 node cluster, `data`:100gb store sales, `columnsToIndex`: ss_customer_sk, ss_item_sk, `cubeSize` 50000000 , `num files`: 86 files

|Operation |Delta     |Qbeast Main       |QBeast PR65 Collect After  |Qbeast PR65 Collect Before         |
|------------|:----------:|:----------:|:----------:|:----------:|
|Query     |0.96s    |1.47s   |1.46s   |1.83s |
|ListFiles | 0.24s | 0.59s | 0.59s |0.42s |

We can see two things:
- Penalization of reading multiple files.
- For the largest DeltaLogs (for example the first one, with 5000+ files), this PR outputs better results on listFiles. In the other cases, the results are similar. 

I will revert to a `collect()` after the `groupBy` because due to https://github.com/delta-io/delta/issues/779 it an aggressive action with large DeltaLog can give problems. We could think about optimization later.

cugni (approved):

Issue #10, opened 2021-06-11
NULL values on indexed columns support


polsm91:
**What went wrong?**

We tried to index a dataset from the TPC-DS, and two of the chosen columns for indexing had null values. We should add support to null values.

```
21/09/28 09:17:18 ERROR Executor: Exception in task 10.0 in stage 4.0 (TID 28)
org.apache.spark.sql.AnalysisException: Column to index contains null values. Please initialize them before indexing
```

**How to reproduce?**
1. Code that triggered the bug, or steps to reproduce:
Indexing notebook s3 public dataset by "ss_customer_sk", "ss_item_sk", "ss_sold_date_sk"

```scala
parquet_df.write.mode("overwrite").format("qbeast").option("columnsToIndex", "ss_customer_sk,ss_item_sk,ss_sold_date_sk").save(qbeast_table_path)
```
2. Branch and commit id:
Main on 15667c27bb2cc6d76cecd680d61e22fa8f571d49

3. Spark version:
3.1.1

4. Hadoop version:
3.2.2

5. Are you running Spark inside a container? Are you launching the app on a remote K8s cluster? Or are you just running the tests on a local computer?
I am running Spark shell on a local computer.

6. Stack trace:

```bash
Driver stacktrace:
  at org.apache.spark.scheduler.DAGScheduler.failJobAndIndependentStages(DAGScheduler.scala:2253)
  at org.apache.spark.scheduler.DAGScheduler.$anonfun$abortStage$2(DAGScheduler.scala:2202)
  at org.apache.spark.scheduler.DAGScheduler.$anonfun$abortStage$2$adapted(DAGScheduler.scala:2201)
  at scala.collection.mutable.ResizableArray.foreach(ResizableArray.scala:62)
  at scala.collection.mutable.ResizableArray.foreach$(ResizableArray.scala:55)
  at scala.collection.mutable.ArrayBuffer.foreach(ArrayBuffer.scala:49)
  at org.apache.spark.scheduler.DAGScheduler.abortStage(DAGScheduler.scala:2201)
  at org.apache.spark.scheduler.DAGScheduler.$anonfun$handleTaskSetFailed$1(DAGScheduler.scala:1078)
  at org.apache.spark.scheduler.DAGScheduler.$anonfun$handleTaskSetFailed$1$adapted(DAGScheduler.scala:1078)
  at scala.Option.foreach(Option.scala:407)
[...]

PR #67, opened 2022-01-25
Add support for null values

osopardo1:
This PR shows the current status on the solution for issue #10 

The theory is to substitute/map null values in the indexed columns transformation with a Random number in the [min, max] range. 

- [x] New API for `LinearTransformation` and `HashTransformation` case classes to include `nullValue` as part of the contained information
- [x] New API for the `Transformer` which includes an `optionalNullValue`. If none is given, then a random is calculated internally with the min/max values
- [x] New API for the user (optional): 
```scala 
df.write.format("qbeast").option("columnsToIndex", "column:type(nullValue),column2:type(nullValue2)...")
```
- [x] Adaptation of the tests to this new functionality


core/src/main/scala/io/qbeast/core/transform/LinearTransformer.scala
```diff
@@ -26,18 +37,60 @@ case class LinearTransformer(columnName: String, dataType: QDataType) extends Tr
...
       statsSqlPredicates = Seq(s"max($columnName) AS $colMax", s"min($columnName) AS $colMin"))
 
   override def makeTransformation(row: String => Any): Transformation = {
-    val min = getValue(row(colMin))
-    val max = getValue(row(colMax))
-    assert(min != null && max != null)
+    val minAux = row(colMin)
+    val maxAux = row(colMax)
+    val (min, max) =
+      // TODO If all values are null...
```
osopardo1, line 80: Hey! Opening a discussion here: what should be the measure when **all values** are null? 
Right now,  min and max are initialized with the minimum and maximum values of the dataType.

cugni: The ideal is that we remove this dimension, and we add it later when we find no null data we issue a new revision with the new dimension. In the end, we should distinguish between the columns that the user thinks we should take into consideration and the ones that are really necessary/useful. I think is best if we disconnect the two, so we can simply drop columns where all data is null or have the same value

cugni (changes requested):
We are still missing the filtering push down for the null values, meaning that a query with x=null, should be converted to x>=randomNullValue and x<randomNullValue and executed using the index.

core/src/main/scala/io/qbeast/core/transform/LinearTransformation.scala
```diff
@@ -57,12 +63,18 @@ case class LinearTransformation(minNumber: Any, maxNumber: Any, orderedDataType:
...
   override def merge(other: Transformation): Transformation = {
     other match {
-      case LinearTransformation(otherMin, otherMax, otherOrdering)
+      case LinearTransformation(otherMin, otherMax, otherNullValue, otherOrdering)
           if orderedDataType == otherOrdering =>
-        LinearTransformation(min(minNumber, otherMin), max(maxNumber, otherMax), orderedDataType)
+        LinearTransformation(
+          min(minNumber, otherMin),
+          max(maxNumber, otherMax),
+          nullValue, // we should keep the original null value?
```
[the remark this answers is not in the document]
osopardo1: We agree to choose otherNullValue since:

- If it's specified by the user, we should maintain the choice
- If it's a new null value, it would be adapted to the new space

core/src/main/scala/io/qbeast/core/transform/HashTransformation.scala
```diff
@@ -1,14 +1,16 @@
 package io.qbeast.core.transform
 
+import scala.util.Random
 import scala.util.hashing.MurmurHash3
 
 /**
  * A hash transformation of a coordinate
  */
-case class HashTransformation() extends Transformation {
+case class HashTransformation(nullValue: Any) extends Transformation {
```
eavilaes, line 9: Add `nullValue` to scaladoc

core/src/main/scala/io/qbeast/core/transform/LinearTransformation.scala
```diff
@@ -29,7 +29,11 @@ import java.math.BigDecimal
  */
 @JsonSerialize(using = classOf[LinearTransformationSerializer])
 @JsonDeserialize(using = classOf[LinearTransformationDeserializer])
-case class LinearTransformation(minNumber: Any, maxNumber: Any, orderedDataType: OrderedDataType)
+case class LinearTransformation(
```
eavilaes, line 32: Add `nullValue` to scaladoc

core/src/test/scala/io/qbeast/core/transform/LinearTransformationTest.scala
```diff
@@ -27,36 +27,43 @@ class LinearTransformationTest extends AnyFlatSpec with Matchers {
...
-      90000,
-      IntegerDataType)
+      LinearTransformation(
+        -100,
+        90000,
+        nullValue,
+        IntegerDataType)) shouldBe LinearTransformation(-100, 90000, nullValue, IntegerDataType)
 
-    linearT.merge(LinearTransformation(6, 9, IntegerDataType)) shouldBe linearT
+    linearT.merge(LinearTransformation(6, 9, 7, IntegerDataType)) shouldBe linearT
```
eavilaes, line 66: Probably check the values instead of the whole Transformation?

core/src/main/scala/io/qbeast/core/transform/LinearTransformer.scala
```diff
@@ -15,7 +25,11 @@ object LinearTransformer extends TransformerType {
  * @param columnName the column name
  * @param dataType the data type of the column
  */
-case class LinearTransformer(columnName: String, dataType: QDataType) extends Transformer {
+case class LinearTransformer(
```
eavilaes, line 28: Add `optionalNullValue` to scaladoc

eavilaes (changes requested):
Just a few comments, but generally looks good! :)

core/src/test/scala/io/qbeast/core/transform/LinearTransformationTest.scala
osopardo1: Yes, thank you!

core/src/main/scala/io/qbeast/core/transform/LinearTransformer.scala
osopardo1: Solved on e8cdf06

core/src/test/scala/io/qbeast/core/transform/LinearTransformationTest.scala
osopardo1: Solved on e8cdf06

core/src/main/scala/io/qbeast/core/transform/LinearTransformation.scala
osopardo1: Solved on e8cdf06

core/src/main/scala/io/qbeast/core/transform/LinearTransformer.scala
osopardo1: I opened a PR on my fork to check if it makes sense to drop columns that have the same value (null or not)
https://github.com/osopardo1/qbeast-spark/pull/1

Please check it when you can @eavilaes @cugni ! :pray: Thank you!

osopardo1: Update: If all values are null, we pick null/zero value. PR osopardo1#1 is left for later

core/src/main/scala/io/qbeast/core/transform/LinearTransformation.scala
```diff
@@ -38,17 +43,19 @@ case class LinearTransformation(minNumber: Any, maxNumber: Any, orderedDataType:
 
   private val scale: Double = {
     val mx = maxNumber.toDouble
-    require(mx > mn, "Range cannot be not null, and max must be > min")
-    1.0 / (mx - mn)
+    require(mx >= mn, "Range cannot be not null, and max must be >= min")
+    if (mx == mn) 0.0 else 1.0 / (mx - mn)
```
osopardo1, line 47: Please just check this and let me know if it's correct :)

In case **min and max are equal** (nulls or not), `scale` should be `0.0`

eavilaes: I think that's correct now. Sorry for the late response!

eavilaes (approved):

eavilaes:
I've seen some typos on the new `AdvancedConfiguration.md` file, you could maybe check it with Grammarly! :)

osopardo1:
> I've seen some typos on the new `AdvancedConfiguration.md` file, you could maybe check it with Grammarly! :)

Thank you, fixed!

osopardo1:
Hey! I think is ready to merge, changes are addressed and it's compatible with the current version in the `main` :+1:

eavilaes (approved):
Looks good to me! Great job :)

core/src/main/scala/io/qbeast/core/transform/LinearTransformer.scala
```diff
@@ -32,14 +32,21 @@ case class LinearTransformer(columnName: String, dataType: QDataType) extends Tr
...
-    assert(min != null && max != null)
-    dataType match {
-      case ordered: OrderedDataType =>
-        LinearTransformation(min, max, ordered)
-
+    val minAux = row(colMin)
+    val maxAux = row(colMax)
+    if (minAux == maxAux) {
+      // If all values are null or equal we return an IdentityTransformation
+      IdentityTransformation
```
cugni, line 39: Maybe is best just to put a new NullToZeroTransformation that does only that, instead of adding the case null => 0.0 in the identity. It is more explicitly. And also, it is not identity

core/src/main/scala/io/qbeast/core/transform/Transformation.scala
```diff
@@ -51,6 +51,7 @@ object IdentityTransformation extends Transformation {
 
   @inline
   override def transform(value: Any): Double = value match {
+    case null => 0.0
```
cugni, line 54: then it is not "identity"

cugni, line 54: then it is not "identity"

core/src/main/scala/io/qbeast/core/transform/LinearTransformer.scala
osopardo1: Cool!

cugni (approved):
Perfect 👌🏻

Issue #66, opened 2022-01-24
NullPointerException when indexing Strings in non-null columns.


eavilaes:
**What went wrong?**
Indexing the table `call_center` from TPC-DS using non-null columns throws a NullPointerException. `call_center` is a really small table with **24 rows** (in the 10Gb dataset), with no nulls in the indexed columns.
The column causing problems is `cc_call_center_id`, which is a **String**. In the following example, I tried to index using two columns, but I also did the test using only this column and it also fails (Indexing the other column, `cc_call_center_sk`, works well).
```scala
df.printSchema
root
 |-- cc_call_center_sk: integer (nullable = true)
 |-- cc_call_center_id: string (nullable = true)
  ...
```

There are only 24 rows with no nulls in them::
```scala
scala> df.select("cc_call_center_sk", "cc_call_center_id").where("cc_call_center_id is null or cc_call_center_sk is null").show()
+-----------------+-----------------+
|cc_call_center_sk|cc_call_center_id|
+-----------------+-----------------+
+-----------------+-----------------+

scala> df.select("cc_call_center_sk", "cc_call_center_id").show(30)
+-----------------+-----------------+
|cc_call_center_sk|cc_call_center_id|
+-----------------+-----------------+
|                1| AAAAAAAABAAAAAAA|
|                2| AAAAAAAACAAAAAAA|
|                3| AAAAAAAACAAAAAAA|
|                4| AAAAAAAAEAAAAAAA|
|                5| AAAAAAAAEAAAAAAA|
|                6| AAAAAAAAEAAAAAAA|
|                7| AAAAAAAAHAAAAAAA|
|                8| AAAAAAAAIAAAAAAA|
|                9| AAAAAAAAIAAAAAAA|
|               10| AAAAAAAAKAAAAAAA|
|               11| AAAAAAAAKAAAAAAA|
|               12| AAAAAAAAKAAAAAAA|
|               13| AAAAAAAANAAAAAAA|
|               14| AAAAAAAAOAAAAAAA|
|               15| AAAAAAAAOAAAAAAA|
|               16| AAAAAAAAABAAAAAA|
|               17| AAAAAAAAABAAAAAA|
|               18| AAAAAAAAABAAAAAA|
|               19| AAAAAAAADBAAAAAA|
|               20| AAAAAAAAEBAAAAAA|
|               21| AAAAAAAAEBAAAAAA|
|               22| AAAAAAAAGBAAAAAA|
[...]

PR #69, opened 2022-01-27
Solved NullPointerException when indexing strings

osopardo1:
This PR solves issue #66 where a NullPointerException was thrown when indexing non-null strings.

This was due to `ColumnStats` initialization. We were trying to cast `stddev` and `mean` values to `Double` directly when for this parameter is `null` for `Strings`.


eavilaes (approved):
Looks good to me, thank you!

core/src/main/scala/io/qbeast/core/transform/Transformer.scala
```diff
@@ -147,8 +147,8 @@ object ColumnStats {
     val min = stats("min")
     val max = stats("max")
     val count = stats("count").toLong
-    val stddev = stats("stddev").toDouble
-    val mean = stats("mean").toDouble
+    val stddev = stats("stddev")
```
cugni, line 150: If we are not using stddev, we should not calculate it. we should avoid using all metrics that are not useful for us.

cugni (changes requested):

core/src/main/scala/io/qbeast/core/transform/Transformer.scala
```diff
@@ -127,40 +118,19 @@ trait Transformer extends Serializable {
...
-      case LongDataType => (min.toLong, max.toLong)
-      case FloatDataType => (min.toFloat, max.toFloat)
-      case DecimalDataType => (min.toDouble, max.toDouble)
-      case StringDataType => (Nil, Nil)
-      case _ => throw new RuntimeException(s"Data type $dataType not supported")
-    }
-
-    ColumnStats(minVal, maxVal, count, stddev, mean)
-  }
+case class ColumnStats(names: Seq[String], columns: Seq[String]) extends Serializable {
```
cugni, line 128: that is a column? I guess is the type of statistic that we need? Like if a column is named age, we will get max_age, min_age, no? I think it would be best to call names columns, and columns stats ( or aggregation, SQL, metrics, whatever you like).

src/main/scala/io/qbeast/spark/index/OTreeDataAnalyzer.scala
```diff
@@ -142,41 +167,24 @@ object DoublePassOTreeDataAnalyzer extends OTreeDataAnalyzer with Serializable {
...
       isReplication: Boolean): (DataFrame, TableChanges) = {
 
     // Compute the statistics for the indexedColumns
-    val columnStats = getColumnStats(dataFrame, indexStatus.revision.columnTransformers)
+    val columnTransformers = indexStatus.revision.columnTransformers
+    val dataFrameStats = getDataFrameStats(dataFrame, columnTransformers)
 
-    // Check if the DataFrame is empty
-    if (columnStats.head.count == 0) {
+    if (dataFrameStats.isEmpty) {
```
cugni, line 179: If you are adding the count in the SQL statement, it should never return 0 rows, no? Is this if necessary?

osopardo1: True! But I believe if it's empty we should not continue, or it will fail on:
 
`assert(min != null && max != null)`
https://github.com/osopardo1/qbeast-spark/blob/a041269b09f0da4cbf95b92b0771183318663be5/core/src/main/scala/io/qbeast/core/transform/LinearTransformer.scala#L37

cugni (approved):

Issue #43, opened 2021-11-16
Easier information API


osopardo1:
In the current implementation, configurations like: `columnsToIndex`, `desiredCubeSize` and other index-related characteristics such as `revisions` available are not visible (easily) to the user/developer. 

We should provide an API through the `QbeastTable` interface to allow anyone to consult those values. 

Example of the functionality:

```scala
import io.qbeast.spark.table._

val dir = "path/to/mytable"
val qbeastTable = QbeastTable.forPath(spark, dir)

qbeastTable.getRevisions()
qbeastTable.getIndexedColumns()
qbeastTable.getDesiredCubeSize()
...

```

osopardo1:
Merged in #72

PR #72, opened 2022-02-04
Easier information API

osopardo1:
Related to #43, this PR adds new methods to `QbeastTable` that allows both developers and users to understand the metadata of qbeast, such as:

```scala
val qbeastTable = QbeastTable.forPath(spark, path)
qbeastTable.indexedColumns() // Outputs the actual indexed columns
qbeastTable.cubeSize() // Outputs the actual cube size
qbeastTable.revisionsID() // Outputs the list of revision identifiers available for a table
qbeastTable.latestRevisionID() // Outputs the latest revision identifier available for a table
```

Any suggestion is welcome! Feel free to ask for metadata information that would be relevant to you :smile:

### DIFF
--- a/src/main/scala/io/qbeast/spark/QbeastTable.scala
+++ b/src/main/scala/io/qbeast/spark/QbeastTable.scala
@@ -6,11 +6,10 @@ package io.qbeast.spark
 import io.qbeast.context.QbeastContext
 import io.qbeast.core.model.{QTableID, RevisionID}
 import io.qbeast.spark.delta.DeltaQbeastSnapshot
+import io.qbeast.spark.internal.commands.{AnalyzeTableCommand, OptimizeTableCommand}
 import io.qbeast.spark.table._
 import org.apache.spark.sql.SparkSession
 import org.apache.spark.sql.delta.DeltaLog
-import io.qbeast.spark.internal.commands.AnalyzeTableCommand
-import io.qbeast.spark.internal.commands.OptimizeTableCommand
 
 /**
  * Class for interacting with QbeastTable at a user level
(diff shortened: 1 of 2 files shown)


Jiaweihu08 (approved):

src/main/scala/io/qbeast/spark/QbeastTable.scala
```diff
@@ -59,12 +58,52 @@ class QbeastTable private (
...
+    qbeastSnapshot.loadAllRevisions.map(_.revisionID)
+  }
+
+  /**
+   * Outputs the identifier of the latest revision available
+   * @return
+   */
+
+  def latestRevisionID(): RevisionID = {
+    qbeastSnapshot.loadLatestRevision.revisionID
```
eavilaes, line 104: Just `latestRevisionAvailable`

src/main/scala/io/qbeast/spark/QbeastTable.scala
```diff
@@ -59,12 +58,52 @@ class QbeastTable private (
...
+
+  def revisionsID(): Seq[RevisionID] = {
+    qbeastSnapshot.loadAllRevisions.map(_.revisionID)
+  }
+
+  /**
+   * Outputs the identifier of the latest revision available
+   * @return
+   */
+
```
eavilaes, line 102: Remove empty line.

src/main/scala/io/qbeast/spark/QbeastTable.scala
```diff
@@ -59,12 +58,52 @@ class QbeastTable private (
...
+  def cubeSize(revisionID: RevisionID = -1L): Int = {
+    val revID = getAvailableRevision(revisionID)
+    qbeastSnapshot.loadRevision(revID).desiredCubeSize
+  }
+
+  /**
+   * Outputs all the revision identifiers available for the table
+   * @return
+   */
+
```
eavilaes, line 93: Remove empty line, so the doc is stick to the method.

eavilaes (changes requested):
Great work and nice tests! Just a couple of comments, really small things:

src/main/scala/io/qbeast/spark/QbeastTable.scala
osopardo1: Cool!

eavilaes (approved):

src/main/scala/io/qbeast/spark/QbeastTable.scala
```diff
@@ -46,7 +45,7 @@ class QbeastTable private (
    * @param revisionID the identifier of the revision to optimize.
    *                          If doesn't exist or none is specified, would be the last available
    */
-  def optimize(revisionID: Option[RevisionID] = None): Unit = {
+  def optimize(revisionID: RevisionID = -1L): Unit = {
```
cugni, line 48: Why the -1L, it is a bit ugly to use a specific number with a specific value. It was better before with the Option internally. 
I understand that having to do optimize(Some(23))  is ugly, but that can be solved by having two public methods optimize() and optimize(123) that map to the same private method. In the end, is what the scala does under the hood with default arguments.

src/main/scala/io/qbeast/spark/QbeastTable.scala
```diff
@@ -59,12 +58,50 @@ class QbeastTable private (
...
+  def cubeSize(revisionID: RevisionID = -1L): Int = {
+    val revID = getAvailableRevision(revisionID)
+    qbeastSnapshot.loadRevision(revID).desiredCubeSize
+  }
+
+  /**
+   * Outputs all the revision identifiers available for the table
+   * @return
+   */
+  def revisionsID(): Seq[RevisionID] = {
```
cugni, line 93: Shouldn't it be revisionsIDs?

cugni (changes requested):

src/main/scala/io/qbeast/spark/QbeastTable.scala
osopardo1: Yes, the reason was to not put Some(x), but two different public methods is a good catch

cugni (approved):
TOP! 🔝

PR #73, opened 2022-02-07
Ensuring that the TableChange is always broadcasting

cugni:
To avoid issues when we write a large dataset, we must ensure that we always broadcast the TableChanges ( and we try to keep it small, but that is for another PR): 

I've drafted the first change, but there are still a few points to resolve:
- [x] how to manage the state? It does not make sense to manage the state for each row independently, but we should merge it in the file meta. 
- [x] Ensure everything works with the last changes. 
- [x]  Ensure the changes make sense and that the structure of the new metadata makes sense


src/main/scala/io/qbeast/spark/index/SparkPointWeightIndexer.scala
```diff
@@ -70,22 +34,26 @@ private class SparkPointWeightIndexer(tableChanges: TableChanges, isReplication:
...
+              col(cubeToReplicateColumnName))))
+    } else {
+      weightedDataFrame
+        .withColumn(
+          cubeColumnName,
+          explode(
+            findTargetCubeIdsUDF(
+              struct(columnsToIndex.map(col): _*),
+              col(weightColumnName),
+              lit(null))))
```
osopardo1, line 55: I don't know which is the best workaround for separating replication and indexing at this stage. The options are:

1. Create different findTargetCubeIdsUDF with different arguments so we don't have to create a column when it's not necessary.
2. Create different SparkPointWeightIndexer implementations
3. Leave it as it is: a column with a null value is created
4. Other 

Any suggestion is welcome @eavilaes  @Jiaweihu08 @cugni !

cugni: I think it is the case it is best to create 2 UDF. We haven't analyzed well enough in the optimize phase to understand the cost of adding columns and adding more logic, of if it is best to change the algorithm as a whole. Also, an udf without replication can simplify a bit the function logic, which is not bad.

osopardo1: Yes, less overhead on creating a column and checking if it's null :+1:

osopardo1 (approved):

core/src/main/scala/io/qbeast/core/model/PointWeightIndexer.scala
```diff
@@ -2,13 +2,9 @@ package io.qbeast.core.model
 
 /**
  * Indexes a point by its weight and cube transformation.
- * @param cubeWeights the map of cube and weight
- * @param announcedOrReplicatedSet the set of cubes in announced or replicated state
+ * @param tableChanges the map of cube and weight
```
eavilaes, line 5: Wrong scaladoc

src/test/scala/io/qbeast/spark/index/SparkPointWeightIndexerTest.scala
```diff
@@ -17,34 +17,46 @@ class SparkPointWeightIndexerTest extends QbeastIntegrationTestSpec {
...
+//    indexed.find(_._1.isRoot) shouldBe Some((CubeId.root(3), "REPLICATED"))
+//    val sec = CubeId.root(3).firstChild
+//    indexed.find(_._1 == sec) shouldBe Some((sec, "REPLICATED"))
+//
+//    val gen3 = CubeId.root(3).firstChild.firstChild
+//    indexed.find(_._1 == gen3) shouldBe Some((gen3, "ANNOUNCED"))
+//
+//    indexed.count(_._2 == "FLOODED") shouldBe 7
+//
+//  })
```
eavilaes, line 59: Are we leaving this commented?

eavilaes (approved):
Just a couple of comments, everything looks good to me!

src/test/scala/io/qbeast/spark/index/SparkPointWeightIndexerTest.scala
osopardo1: Ugh thank you!

PR #74, opened 2022-02-09
Add API to get Index Metrics

Jiaweihu08:
Add an API to `QbeastTable` to retrieve OTree index metrics more easily!

This is how would you use it:
```bash
val qbeastTable = QbeastTable.forPath(spark, tmpDir)
val metrics = qbeastTable.getIndexMetrics()
```

The metrics included so far are:
General index metadata:
- `desiredCubeSize`
- number of cubes
- depth
- average fan out of the cubes
- dimension count
- number of rows

Some more specific details such as `depthOverLogNumNodes = depth / log(cubeCounts)`, `depthOnBalance = depth / log(rowCount/desiredCubeSize)`, both logs use base = `dimensionCount`.

We also take a closer look at the non leaf cube sizes. `NonLeafCubeSizeDetails` contains their `min`, `max`, `quantiles`, and how far each of the cube sizes are from the `desiredCubeSize`.

`Map[CubeId, CubeStatus]` of the index is also returned, since some of the information stored in `CubeStatus` are interesting to analyze, such as the distribution of `cube weights` for different indexes.

You can access this information through `metrics.cubeStatuses`.

Example output:
```
OTree Index Metrics:
dimensionCount: 2
elementCount: 1001
depth: 2
cubeCounts: 7
desiredCubeSize: 100
avgFanOut: 2.0
depthOverLogNumNodes: 0.7124143742160444
depthOnBalance: 0.6020599913279624
Non-lead Cube Size Stats:
(All values are 0 if there's no non-leaf cubes):
- min: 3729
- firstQuartile: 3729
- secondQuartile: 4832
- thirdQuartile: 5084
- max: 5084
- dev: 2.0133907E7
```

### DIFF
--- a/docs/IndexMetrics.md
+++ b/docs/IndexMetrics.md
@@ -1,0 +1,61 @@
+# Peeking into the OTree Index Metrics
+
+The API is implemented to retrieve information about your table's **OTree index**.
+
+You can use it to compare the index build with different indexing parameters such as the `desiredCubeSize` and `columnsToIndex`.
+
+If you're experimenting with new ways of implementing the OTree algorithm, you can also use this API to analyze the resulting index!
+
+## Metrics
+1. General index metadata:
+- **Desired cube size**
+- **Number of cubes**
+- **Tree depth**
+- **Average fan out of the cubes**
+- **Dimension count**
+- **Number of rows**
+
+2. Some more specific details such as:
+- **depthOverLogNumNodes = depth / log(cubeCounts)**
+- **depthOnBalance = depth / log(rowCount/desiredCubeSize)**
+
+  both logs use **base = dimensionCount**
+
+3. Cube sizes for non-leaf cubes:
+- `NonLeafCubeSizeDetails` contains their **min**, **max**, **quantiles**, and how far each of the cube sizes are from the `desiredCubeSize`(**dev**).
+
+4. `Map[CubeId, CubeStatus]`
+- Some information from the map can be interesting to analyze - for example, the distribution of cube weights. You can access this information through `metrics.cubeStatuses`.
+
+
+## How to use it
+```scala
+import io.qbeast.spark.QbeastTable
+
+val qbeastTable = QbeastTable.forPath(spark, tablePath)
+val metrics = qbeastTable.getIndexMetrics()
+
+println(metrics)
+```
+
+## Example output:
+
+```
+OTree Index Metrics:
+dimensionCount: 2
+elementCount: 1001
+depth: 2
+cubeCounts: 7
+desiredCubeSize: 100
+avgFanOut: 2.0
+depthOverLogNumNodes: 0.7124143742160444
+depthOnBalance: 0.6020599913279624
+Non-lead Cube Size Stats:
+(All values are 0 if there's no non-leaf cubes):
+- min: 3729
+- firstQuartile: 3729
+- secondQuartile: 4832
+- thirdQuartile: 5084
+- max: 5084
+- dev: 2.0133907E7
+```


src/main/scala/io/qbeast/spark/QbeastTable.scala
```diff
@@ -65,6 +65,52 @@ class QbeastTable private (
...
+      allCubeStatuses.filter(_._1.children.exists(allCubeStatuses.contains)).values
+    val nonLeafCubeSizes = nonLeafStatuses.flatMap(_.files.map(_.size)).toSeq.sorted
+    val nonLeafCubeSizeDeviation =
+      nonLeafCubeSizes
+        .map(cubeSize => math.pow(cubeSize - desiredCubeSize, 2) / nonLeafCubeSizes.size)
+        .sum
+
+    val avgFanOut = nonLeafStatuses
+      .map(_.cubeId.children.count(allCubeStatuses.contains))
+      .sum / nonLeafStatuses.size
```
osopardo1, line 87: If nonLeafStatuses happens to be 0 (because it only exists the cube root), a `java.lang.ArithmeticException: / by zero` is thrown

src/main/scala/io/qbeast/spark/QbeastTable.scala
```diff
@@ -65,6 +65,52 @@ class QbeastTable private (
...
+      avgFanOut,
+      depthOverLogNumNodes,
+      depthOnBalance,
+      NonLeafCubeSizeDetails(
+        nonLeafCubeSizes.min,
+        nonLeafCubeSizes((nonLeafCubeSizes.size * 0.25).toInt),
+        nonLeafCubeSizes((nonLeafCubeSizes.size * 0.50).toInt),
+        nonLeafCubeSizes((nonLeafCubeSizes.size * 0.75).toInt),
+        nonLeafCubeSizes.max,
+        nonLeafCubeSizeDeviation))
```
osopardo1, line 107: We should return IndexMetrics or directly the String with the information? So we keep private the case class

src/main/scala/io/qbeast/spark/QbeastTable.scala
```diff
@@ -65,6 +65,52 @@ class QbeastTable private (
...
+
+    val cubeCounts = allCubeStatuses.size
+    val depth = allCubeStatuses.map(_._1.depth).max
+    val row_count = allCubeStatuses.flatMap(_._2.files.map(_.elementCount)).sum
+    val dimensionCount = allCubeStatuses.toList.head._1.dimensionCount
+    val desiredCubeSize = qbeastSnapshot.loadLatestRevision.desiredCubeSize
+
+    val nonLeafStatuses =
+      allCubeStatuses.filter(_._1.children.exists(allCubeStatuses.contains)).values
+    val nonLeafCubeSizes = nonLeafStatuses.flatMap(_.files.map(_.size)).toSeq.sorted
```
osopardo1, line 79: Bug found!
Here you are mapping to the file `size` in bytes instead of the `elementCount`

src/main/scala/io/qbeast/spark/QbeastTable.scala
Jiaweihu08: Fixed!

src/main/scala/io/qbeast/spark/QbeastTable.scala
Jiaweihu08: Fixed!

Jiaweihu08:
Shouldn't we also add IndexStatuses as well? @osopardo1 
The distribution of weights can also be interesting to analize. We can include it in `IndexMetrics` and leave `toString` untouched.

Jiaweihu08:
> > Shouldn't we also add `IndexStatuses` as well? @osopardo1 The distribution of weights can also be interesting to analize. We can include it in `IndexMetrics` and leave `toString` untouched.
> 
> Do you mean the Cube and Weight map? Because I don't know if including the files makes sense. But yeah, sure!
> 
> Also, let's include a document in `docs` where we explain the API (sorry, I missed doing this part when I added new methods). Let me know if you take the task or I will!

Yes, I meant Cube Statuses. I'll do the documentation. Thanks!

osopardo1 (approved):

Issue #41, opened 2021-11-11
Set transaction protocol for optimizing the index


osopardo1:
On write data, we use the Delta Commit Protocol to ensure the atomicity of the operation. 
On optimizing the index, we should encapsulate the write of new files (the ones belonging to the replicated cubes) on a Keeper transaction. 

The pseudocode for resolving conflicting files should look like this: 

```scala
  beginKeeperOperation()
  var tries = 2
    try {
      // Here new data is written
      val filesUpdates = writeData()
      while (tries > 0) {
        deltaLog.withNewTransaction(tnx => {
          val startingTnx = //number of transaction at the beginning of the operation
          try {
            tnx.commit(filesUpdates)
            tries = 0
            succeeded = Some(true)
          } catch {
            case cme: ConcurrentModificationException
                if isConflicted || tries == 0 => succeeded = Some(False); throw cme
            case _: ConcurrentModificationException => tries -= 1
          }
        })
      }

    } finally {
      endKeeperOperation()
    }
```

PR #77, opened 2022-02-26
Set transaction protocol for optimizing

osopardo1:
This PR is solving #41 and adding the logic on Qbeast-Spark to interact with the Keeper. 

There's two main changes on Indexed Table:

- In the **writing** part, it would try to commit and check if some conflict error has been thrown. In the affirmative case, it would need to understand if the conflict is unsolvable or not (e.g: some cubes that weren't announced at the moment of writing are replicated...). If it is unsolvable, it would fail. On the contrary, it would retry the operation one more time.
- In the **optimizing** part, the keeper is called properly. 

And also a new function is added to the Metadata Manager:
- `hasConflicts` checks if there has been any conflict on the commit log.


osopardo1:
I am closing this for the moment. It will be reopened when working on productization.

osopardo1:
I reopen again this PR since we need to integrate with the Keeper/Curator properly.

src/main/scala/io/qbeast/spark/table/IndexedTable.scala
```diff
@@ -193,16 +195,30 @@ private[table] class IndexedTableImpl(
 
   private def write(data: DataFrame, indexStatus: IndexStatus, append: Boolean): BaseRelation = {
     val revision = indexStatus.revision
-
-    if (exists) {
-      keeper.withWrite(tableID, revision.revisionID) { write =>
-        val announcedSet = write.announcedCubes.map(revision.createCubeId)
+    keeper.withWrite(tableID, revision.revisionID) { write =>
+      var tries = 2
```
polsm91, line 199: Should it be a config property? It may make sense to tune it on different scenarios

polsm91 (approved):
In general looks good to me. Regarding testing, could we add a unit test which conflicts and another that does not, and call the `isConflicted` method?

src/main/scala/io/qbeast/spark/table/IndexedTable.scala
osopardo1: Mm, makes sense. Perhaps something like `spark.qbeast.keeper.numOfTries`?

core/src/main/scala/io/qbeast/core/model/MetadataManager.scala
```diff
@@ -47,4 +47,20 @@ trait MetadataManager[DataSchema, FileDescriptor] {
...
+   * This function checks if there's a conflict. A conflict happens if there
+   * are new cubes that have been optimized but they were not announced.
+   *
+   * @param tableID the table ID
+   * @param revisionID the revision ID
+   * @param knownAnnounced the cubes we know they were announced when the write operation started.
+   * @param oldReplicatedSet the old replicated set
+   * @return true if there a conflict, false otherwise
+   */
+  def isConflicted(
```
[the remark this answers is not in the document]
osopardo1: `hasConflicts` better :+1:

PR #78, opened 2022-03-02
Single pass OTreeDataAnalyzer

Jiaweihu08:
Current implementation of `OTreeDataAnalyzer` requires first gathering the statistics on the indexed columns to define the dimension limits before proceeding to calculate the cube weights.

The aim of this implementation is to remove the need to compute the stats in a separate scan, and let the individual partitions to index the their data within their own dimensional limits. The generated partition cube weights are then mapped to the global vector space to output the global estimated cube weights.

Here is a quick comparison between the indexes from the two implementations:
|Metrics                       |Single pass(cs=5k)|Main(cs=5k)|Single pass(cs=5k)|Main(cs=50k)|
|:--------------------:|:------------------:|:-----------:|:------------------:|:-------------:|
|cubeCounts                |349                          |55                |28                            |7                       |
|depth                          |4                               |3                  |2                               |1                       |
|avgFanOut                  |6                               |7                  |2                               |6                      |
|dimCount                    |3                               |3                  |3                               |3                      |
|nonLeafCubeSizeDev|1.8E7                        |34347         |7921                         |7921                |
|size(MB)                     |6.8                            |5.1                |3.6                            |3.1                   |

- Main from main commit: `c9910cc`.
- Dataset: **ecommerce100K_2019_Oct.csv**
- Row counts: **100k**
- File size in parquet: **2.3MB**
- Indexed columns: `event_time`, `user_id`, and `price`

- [ ] Check if the partition stats can be calculated using Spark SQL API instead of iterating through the partition rows.
- [ ] Experiment with Spark Accumulators to find the global ColStats.
- [ ] Find a better way to find overlapping cubes during `toGlobalCubeWeights`
- [ ] Is the current way of assigning weights to overlapping cubes (ol * weight) correct?
- [ ] Is there any risk of losing precision when mapping local coordinates to global coordinates?
- [ ] Find a more efficient way to store column min/max than the case class `ColStats`.
- [ ] Test index quality by comparing index metrics.
- [ ] Test speed up.


osopardo1:
I think we can close this PR. @Jiaweihu08 do you agree?

PR #82, opened 2022-03-08
Releases on Sonatype and Maven Central

eavilaes:
Artifacts should be released on Maven Central to ease the use of qbeast-spark. This way, users will be able to open a spark-shell or pyspark adding our package like:
```
spark-shell --packages io.qbeast:qbeast-spark_2.12:0.2.0
```

To make this easy and not mess up with the current _nightly_ releases on GitHub package registry, I have created a wrapper for the main `qbeast-spark` project, which contains special properties in the `build.sbt`. This new "project" is called `qbeastSparkMaven` (you can suggest a new name if you feel inspired).

This project can be used to release the artifacts to the sonatype repositories.
The goal is to have Sonatype credentials as secrets in the repository to automate it within the CI/CD workflows. Meanwhile, Admins of this repository will be able to manually publish releases using sbt-sonatype and sbt-pgp plugins.


osopardo1:
The only question I have is about the **`delta`** library. If it's not provided, then it's packed in the jar, right? Should we do it like this to not put an extra line on` --packages`?

Second opinions? @polsm91 @cugni

osopardo1:
The only question I have is about the **`delta`** library. If it's not provided, then it's packed in the jar, right? Should we do it like this to not put an extra line on` --packages`?

Second opinions? @polsm91 @cugni

cugni:
As far as I know, when you add a package, maven/ivy should download all
dependencies. If Qbeast depends on delta, it should not necessarily to
specify delta, unless you want to force a specific version.
Il giorno ven 11 mar 2022 alle 15:34 Paola Pardo ***@***.***>
ha scritto:

eavilaes:
> I will leave delta as `% Provided ` and merge in the meantime

>As far as I know, when you add a package, maven/ivy should download all
dependencies. If Qbeast depends on delta, it should not necessarily to
specify delta, unless you want to force a specific version.

Reverted 07665fa in bc4d5af

osopardo1 (approved):

osopardo1:
Feel free to merge :+1:  @eavilaes

Issue #84, opened 2022-03-15
CubeWeights not properly used


Jiaweihu08:
**What went wrong?**
The `update` method from `CubeWeights` duplicates its content each time the `queue` exceeds the `bufferSize`

Basically the behavior is:
```
def update(...){
    ...
    if (queue.size >= bufferSize) {
        // resultBuffer ++= resultInternal() ++ resultBuffer
        resultBuffer ++= result()
        ...
    }
}

def result(): Seq[CubeNormalizedWeight] = {
    resultInternal() ++ resultBuffer
}
```
So aside from adding the elements from `resultInternal()`, each time queue is full, the `resultBuffer` also duplicates its content.

This is obviously going to affect the performance of the OTreeDataAnalyzer, requiring a much large memory space to index than it should. For example, indexing a `store_sales` table of 1.8GB, the current implementation won't be able to pass the `OTreeDataAnalyzer` phase.

**How to reproduce?**
1. Code that triggered the bug, or steps to reproduce:

2. Branch and commit id:
`main`, [16735d6](https://github.com/Qbeast-io/qbeast-spark/commit/16735d63cb0d96c7ba30cf4fbd219d332d956e12) 
3. Spark version:
`3.1.1`

4. Hadoop version:
`3.2`

5. Are you running Spark inside a container? Are you launching the app on a remote K8s cluster? Or are you just running the tests in a local computer?
Local mode

6. Stack trace:

PR #85, opened 2022-03-15
Replace `result` with `resultInternal` for `CubeWeights`

Jiaweihu08:
The `CubeWeights.update` method flushes the `queue` into the `resultBuffer` when its size exceeds a predetermined `bufferSize`.

However, aside from adding the current elements to the `resultBuffer`, it also duplicates the current elements of the `buffer`. Which greatly increases the amount of memory required to do the operations.

```scala
resultBuffer ++= results()
// results() = resultInternal ++ resultBuffer
```

Instead of `resultBuffer ++= results()`, it should be `resultBuffer ++= resultInternal()`.

Solve Issue #84


osopardo1:
Good job! Once tests pass, feel free to merge :+1:

osopardo1 (approved):

PR #90, opened 2022-03-24
Headers and footers for scaladoc generation, including version in HTML

eavilaes:
The current version of scaladoc (`sbt unidoc`) generates HTML documents like the one in the screenshot below. My idea is to include in these some information about the **qbeast-spark version**, both in the **header** and the **footer** of the files:

Screenshot 2022-03-24 at 19 19 24

Any suggestions or modifications are more than welcome!☺️

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -30,8 +30,7 @@ qbeastSpark / Compile / doc / scalacOptions ++= Seq(
   "-doc-version",
   qbeast_spark_version,
   "-doc-footer",
-  "Copyright 2022 Qbeast - Docs for version " + qbeast_spark_version +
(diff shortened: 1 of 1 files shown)


osopardo1:
Header is :top: 

I don't know if for the footer is necessary to include the GitHub link. Did you inspire by other projects? In any case, seems good!

eavilaes:
> Header is 🔝
> 
> I don't know if for the footer is necessary to include the GitHub link. Did you inspire by other projects? In any case, seems good!

As there's no reference or link to the project in the docs, I thought it was great to have it. But I agree it looks kinda weird🙃

osopardo1 (approved):

PR #96, opened 2022-03-31
Minor fixes in documentation and Pull Request Template

eavilaes:
## Description

I realized there were minor things to fix regarding the documentation, so I made some changes.

## Type of change

Documentation update.

Any suggestions or proposed changes are more than welcome!

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,38 +1,38 @@
 ## Description
 
 
-- Fixes # (issue) -> A reference to a related issue in your repository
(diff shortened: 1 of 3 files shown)


osopardo1:
Another thing that we are missing is to change the `--package io.qbeast:qbeast-spark_2.12:0.2.0` :see_no_evil: 

Sorry for not having noticed before, but now that it's up and running, we should switch to that in the documentation. 
For the rest, all good! You are right, the Pull Request Template looks better without the tasks!

osopardo1 (approved):

Issue #102, opened 2022-05-09
Add Convert To Qbeast 


osopardo1:
The only way of writing in Qbeast Format is to load your data and write it again with Spark Dataframes API. 

It could be good to have some more easy ways to convert data in other formats to Qbeast, and that can be compatible with reading when no Metadata is found.

For that, we can think of two approaches:
1. Write the data in the same place but organized with the Qbeast index. If more data is added while the conversion is taking place, we are targeting this data as non-indexed and reading all of them in case we need it. 
2. Write the data in the same place and mark it as replicated cubes. So we will only duplicate the data we need for optimizing. 

Doubts/things we need to figure out:
- How to specify the columns to index in the API
- How to handle partitioning? Should be useful to index the columns that are in partition values?
- Study the feasibility of the second approach
- Other design problems that could arise

osopardo1:
UPDATE

The `Convert To Qbeast` command would be a naïve implementation and would only mark the table with Qbeast Metadata. It will not index any of the existing files, and not even add extra metadata to each entry.
The objective is to **slowly convert the table into the Qbeast Format**, to avoid rewriting the whole dataset in one single process.  

The usage would be something like:

`QbeastTable.convertToQbeast(columnsToIndex="col1,col2", cubeSize=500)`

The operation will trigger a **Metadata Update** that will change the ` Delta Log` with an entry like:

```JSON
{
  "metaData": {
    "id": "aa43874a-9688-4d14-8168-e16088641fdb",
    ...
    "configuration": {
      "qbeast.lastRevisionID": "1",
      "qbeast.revision.1": "{\"revisionID\":1,\"timestamp\":1637851757680,\"tableID\":\"/tmp/qb-testing1584592925006274975\",\"desiredCubeSize\":500,\"columnTransformers\":..}"
    },
    "createdTime": 1637851765848
  }
}
```

osopardo1:
Other aspects/scope of the command:
- **A table entirely written in Delta would not be readable from Qbeast** unless we trigger Convert To Qbeast command.
- **The files not converted to Qbeast would not be optimized, analyzed or compacted**. Manage the compaction of files without Qbeast metadata can be complex. The goal of the conversion is that those files would be gradually written in Qbeast Format. 
- How to manage updates? -> This is a complex topic, not sure how should be handled. We did not tested it yet. Need to explore and understand more about it.

Issue #98, opened 2022-04-05
Add support for compacting files


osopardo1:
## What went wrong?
Recently, Delta contributors added the functionality to Optimize tables through SQL on the Open Source version. :raised_hands: 

You can read everything in the issue related: https://github.com/delta-io/delta/commit/e366ccd6179c70dd603c2093a912aacfe719ed00 but summarize:

Having many small files can turn into a performance issue. To address this, Optimize operation **compacts** them into a **single bigger file**.
```
OPTIMIZE ('/path/to/dir' | delta.table) [WHERE part = 25];
```

This new feature would be part of **next release** `1.2.0 `of **Delta Lake**. Right now, we only have compatibility with` 1.0.0`, and Qbeast does not handle this type of compaction.

In order to be compatible with future Delta versions, we should:
1. Upgrade to the newest version of Delta 
2. Solve compatibility problems that could arise (since it also changes Spark version to 3.2.0)
3. Think about how to address compaction of small files at cube level
4. Implement new functionality
5. Add tests

PR #103, opened 2022-05-10
Convert To Qbeast

osopardo1:
## Description

Adds feature #102 

The idea is to have a way of converting an already existing table (either written in `Parquet` or `Delta`) into the `Qbeast` Format. 

To do so, I added two commands called `ConvertToQbeast` and `PartiallyConvertToQbeast` for experimenting the two approaches proposed in the issue.

## Type of change

It is a **New Feature**  (non-breaking change, which adds new functionality).

## Checklist:
- [x] It's a WIP
- [x] Add tests.

- [x] It works with the parquet format, partitioned or not
- [x] It works with the delta format, partitioned or not
- [x] It extracts record count from the log if it exists; otherwise, it's retrieved from the file metadata
- [x] The operation is idempotent
- [x] It's able to infer the table format; an error is thrown if the format is not a supported one
- [x] It should not break Analyze and Optimize
- [x] It should not break Compaction
- [ ] Intelligent min max retrieval for indexing columns from AddFile stats if exists

I've created `ConvertToQbeastTest` under `utils` to make sure the solution works properly.


osopardo1:
I close it for now, I would reopen it once we finish #98

osopardo1:
Related issues to this Operation are documented on #103 and #102.

There are two ways in which we process Files without Qbeast Metadata:
1. Read. When we read the `DeltaLog` and build an `IndexStatus`, we can process those files without **qbeast tags** separately and include a fake `weight`, `cube`, `elementCount`...
2. Write. While ConvertingToQbeast, we can rewrite the Log directly with the extra metadata.

src/main/scala/io/qbeast/spark/delta/IndexStatusBuilder.scala
```diff
@@ -55,7 +55,9 @@ private[delta] class IndexStatusBuilder(
 
     import spark.implicits._
     val ndims: Int = rev.transformations.size
-    revisionFiles
+    // TODO some files may not include metadata
+    val filesWithMetadata = revisionFiles.where("tags is not null")
```
Jiaweihu08, line 59: With the changes made, is this still necessary?

osopardo1: I don't know if we should support the reading of files without Qbeast Metadata. This case would only happen if the user does compaction with Delta (or other underlying format). 
We can solve this in another PR.

Jiaweihu08:
> Hey @Jiaweihu08 ! Since we are prioritizing other features (such as #141 or #133 ), I think it's best to close this PR for now and get back at any time. Is it ok for you?

Agree

PR #107, opened 2022-05-27
Add lower bound for groupCubeSize

Jiaweihu08:
When the number of elements in a table is large, the `groupCubeSize` used by `CubeWeightsBuilder` tends to 1. This can leads to a large number of estimated Cube Weight pairs that can cause OOM errors.

For instance, using default values for `bufferCapacity` and `desiredCubeSize`, a table with 2600M records split between 1200 partitions, the `groupCubeSize` is 190. The algorithm estimates 37K cubes, among which only around 1K are used. The workaround is to set a minimum value for `groupCubeSize`, for instance 1000, so the number of estimated cubes won't be as large. Using this setting, the number of estimated cubes is reduced to about 5K.

Test added:
Using default `bufferCapacity` and `DesiredCubeSize`, make sure that the minimum value for `groupCubeSize` is respected for different `numElements`.

### DIFF
--- a/core/src/main/scala/io/qbeast/core/model/CubeWeights.scala
+++ b/core/src/main/scala/io/qbeast/core/model/CubeWeights.scala
@@ -6,6 +6,7 @@ import io.qbeast.core.model.Weight.MaxValue
 import scala.collection.mutable
 
 object CubeWeightsBuilder {
+  val minGroupCubeSize = 1000
 
   /**
    * Estimates the groupCubeSize depending on the input parameters.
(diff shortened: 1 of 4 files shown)


Jiaweihu08:
The sampling test failed because now the groupCubeSize is 1000 and there are only a handful of cubes. Use a sampling fraction of 0.01 can solve the problem, should be modify the test? @osopardo1

osopardo1:
This test in particular only checks if the filtering of the files is done correctly. Yes, is best to do it with a sample of 0.01.

But then we need to test better the number we are using to solve the issue. Or at least, test it in corner cases.

osopardo1 (approved):
Looks good to me!

PR #110, opened 2022-06-20
Compaction

osopardo1:
## Description

Adds new feature #98 . :raised_hands: Compaction of small files is coming!

## Type of change

In this PR, we present a new feature: compaction of small files. 

These changes include a new command to produce the compaction:

```scala
val qbeastTable = QbeastTable.forPath(spark, "/tmp/qbeast_table")
qbeastTable.compact()
```

Under the hood, Qbeast will find the files larger than `spark.qbeast.compact.minFileSize` and small than `spark.qbeast.compact.maxFileSize` and compact them into single files. 

You can change this configuration through the `SparkConf`:

```shell
--conf spark.qbeast.compact.minFileSize=1\
--conf spark.qbeast.compact.maxFileSize=10000
```

The default configuration is set to 1024 * 1024 * 1024 bytes as in Delta `DELTA_OPTIMIZE_MIN_FILE_SIZE`.

## Checklist:

- [x] WIP
- [x] Add tests.

Still needs to be tested! I will keep updates on this Pull Request. 

**Test Configuration**:
* Spark Version:
* Hadoop Version:
* Cluster or local?


osopardo1:
I put you both @Jiaweihu08 and @polsm91 as reviewers to give me feedback if you can :) If you need an explanation on something, please let me know.

osopardo1:
Changes explained:
- `IndexedTable` would have a `.compact() `method that would follow the same pipeline as the write: Compact data (group info into "bins" for each cube), write data, and commit.
- `DataWriter` would also have a `.compact()` method that would be in charge of **writing the data into files and outputting the necessary file actions** (Add the compacted files and remove the small ones)
- A case class `Compactor` is created. Since the process is more parallelizable and simpler than the `BlockWriter`, I added a new process to be in charge of writing the parquet file and adding the proper tags.
- `MetadataManager` would **commit the information to the Log**.

src/main/scala/io/qbeast/spark/delta/IndexStatusBuilder.scala
```diff
@@ -59,7 +58,7 @@ private[delta] class IndexStatusBuilder(
     revisionFiles
       .groupBy(TagColumns.cube)
       .agg(
-        weight(min(TagColumns.maxWeight)).as("maxWeight"),
+        min(weight(TagColumns.maxWeight)).as("maxWeight"),
```
polsm91, line 61: I read that this change is because we read a dataset which was partially written in Qbeast and Delta format?
https://github.com/osopardo1/qbeast-spark/commit/ca6c4e07c7f89c72c9956dd08373900103a5887a

osopardo1: Nope, this was a bug found while trying to do so. 

In the end I decided to link that part to another issue (#121 ). Because some parameters can cause problems (`numElements` it's not easy to retrieve) or questions like what revision to use for those files.

src/main/scala/io/qbeast/spark/delta/writer/SparkDeltaDataWriter.scala
```diff
@@ -0,0 +1,182 @@
...
+            // we output a group of files for that cube
+            groups += g.result()
+            // Clear the current group
+            g.clear()
+            // Add the block to the current group
+            g += b
+            count = b.size
+          } else {
+            g += b
+            count += b.size
```
polsm91, line 113: Maybe this part can be generalized outside the If? It seems to be done regardless of the condition :)

src/main/scala/io/qbeast/spark/delta/writer/Compactor.scala
```diff
@@ -0,0 +1,82 @@
+/*
+ * Copyright 2021 Qbeast Analytics, S.L.
+ */
```
polsm91, line 3: These copyright notices should be removed but I see that they are also present in files outside the scope of this PR. Maybe we should open a new issue & PR to clean the headers.

osopardo1: They are in all files :laughing: . Scalafmt is set to enforce them.

Why they should be removed? Is it because of the Copyright?

src/main/scala/io/qbeast/spark/delta/writer/SparkDeltaDataWriter.scala
osopardo1: Yes, you are right, just need to set the count to 0 in the condition

src/main/scala/io/qbeast/spark/delta/IndexStatusBuilder.scala
polsm91: Understood, thanks! :)

src/main/scala/io/qbeast/spark/delta/writer/SparkDeltaDataWriter.scala
osopardo1: Done!

src/main/scala/io/qbeast/spark/delta/writer/Compactor.scala
polsm91: They are only in the files under `src/main` but files under `core/..` do not have them. Neither on the auxiliary files used to build the projects.

My concern is, why do we have a Copyright saying that open-source code belongs to a company?

polsm91: I started to think about this because I saw that the file in `core` on the PR did not have the copyright notice. There are different approaches with different implications, it is something to think for the future. 

ASF recommends including the License header on each file (I guess that for people to understand how the code is licensed in an easier way) but it is not required. 

I'd suggest to keep the Copyright notice as it is and add it on the `core/` files.

src/main/scala/io/qbeast/spark/delta/writer/SparkDeltaDataWriter.scala
```diff
@@ -0,0 +1,194 @@
...
+
+  override def write(
+      tableID: QTableID,
+      schema: StructType,
+      qbeastData: DataFrame,
+      tableChanges: TableChanges): IISeq[FileAction] = {
+
+    val sparkSession = qbeastData.sparkSession
+
+    val (factory: OutputWriterFactory, serConf: SerializableConfiguration) = {
```
alexeiakimov, line 36: Interesting fragment. Of course it is a matter of taste, however I wonder what are the reasons to compute factory ans serConf as a pair? Could the following code have a better readability:
```scala
val job = Job.getInstance()
val factory = new ParquetFileFormat().prepareWrite(sparkSession, job, Map.empty, schema)
val serConf = new SerializableConfiguration(job.getConfiguration)
```

src/main/scala/io/qbeast/spark/delta/writer/SparkDeltaDataWriter.scala
```diff
@@ -0,0 +1,194 @@
...
+   * Based on groupFilesIntoBins from Delta Lake OptimizeTableCommand
+   * @param cubeStatuses the cubes statuses present in the index
+   * @return
+   */
+  def groupFilesToCompact(
+      cubeStatuses: IISeq[(CubeId, IISeq[QbeastBlock])]): IISeq[(CubeId, IISeq[QbeastBlock])] = {
+
+    // Check what cubes are suitable for compaction
+    val cubesToCompact = cubeStatuses
+      .map(g => (g._1, g._2.filter(_.size >= MIN_FILE_SIZE_COMPACTION)))
```
alexeiakimov, line 78: This one is a bit cryptic to my taste, possibly because the letter `g` does not appear in `cubeStatuses`.

src/main/scala/io/qbeast/spark/delta/writer/SparkDeltaDataWriter.scala
```diff
@@ -0,0 +1,194 @@
...
+      .filter(_._2.nonEmpty)
+
+    cubesToCompact.flatMap { case (cube, blocks) =>
+      val groups = Seq.newBuilder[Seq[QbeastBlock]]
+      val g = Seq.newBuilder[QbeastBlock]
```
alexeiakimov, line 83: Possibly `group` is easier to read.

src/main/scala/io/qbeast/spark/delta/writer/SparkDeltaDataWriter.scala
```diff
@@ -0,0 +1,194 @@
...
+
+    val cubesToCompact = indexStatus.cubesStatuses.mapValues(_.files).toIndexedSeq
+    val cubesToCompactGrouped = groupFilesToCompact(cubesToCompact)
+
+    val parallelJobCollection = new ParVector(cubesToCompactGrouped.toVector)
+
+    // Create a thread pool to be able to parallelize the compaction task
+    // Each thread will be in charge of compacting the set of files for one single cube
+    val threadPool =
+      QbeastThreadUtils.threadUtils.newForkJoinPool("Compaction", maxThreadNumber = 15)
```
alexeiakimov, line 134: Normally functions should not create thread pools for every call. Thread in JVM is an expensive abstraction, so the less threads we have the better. The application should control the maximum number of the threads. Here in theory there can be multiplicative effect when the method is called in parallel by several clients. I would prefer to have one dedicated thread pool with some configurable number of threads, the rule of thumb is 2 threads per core.

This part should be improved in the future, possibly as a separate issue in order not to block this PR.

src/main/scala/io/qbeast/spark/delta/writer/SparkDeltaDataWriter.scala
```diff
@@ -0,0 +1,194 @@
...
+
+        }.seq
+      } finally {
+        threadPool.shutdownNow()
+      }
+
+    // Return the sequence of updates
+    // which contains both Added and Removed files
+    // This code is maintained in case we need some pre-processing
+    val addedFiles = updates.collect { case a: AddFile => a }
```
alexeiakimov, line 187: Is it possible to collect both added and removed files with a single scan?

src/main/scala/io/qbeast/spark/delta/IndexStatusBuilder.scala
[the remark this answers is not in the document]
alexeiakimov: Maybe weight function is not monotonic?

src/main/scala/io/qbeast/spark/delta/writer/SparkDeltaDataWriter.scala
osopardo1: Actually, would be even better to return `updates` without processing anything

src/main/scala/io/qbeast/spark/delta/writer/SparkDeltaDataWriter.scala
osopardo1: Sure :laughing: 
It was something copied-and-pasted from based methods. But yes, nothing seems to indicate we need that programming style.

src/main/scala/org/apache/spark/util/QbeastThreadUtils.scala
```diff
@@ -0,0 +1,8 @@
+/*
+ * Copyright 2021 Qbeast Analytics, S.L.
+ */
+package org.apache.spark.util
+
+object QbeastThreadUtils {
```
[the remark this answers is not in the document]
osopardo1: I agree on treating this optimizations in another PR. I removed the use of Thread on 8e64b2ad9ef7f63253cf04b2aa583f75fd568aa6

src/main/scala/io/qbeast/spark/delta/IndexStatusBuilder.scala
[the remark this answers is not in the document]
osopardo1: The problem is that the value is readen as a String, so the numeric comparison isn't possible without casting it first.

src/main/scala/io/qbeast/spark/delta/writer/Compactor.scala
```diff
@@ -0,0 +1,82 @@
...
+    val writtenPath = new Path(tableID.id, s"${UUID.randomUUID()}.parquet")
+    val writer: OutputWriter = factory.newInstance(
+      writtenPath.toString,
+      schema,
+      new TaskAttemptContextImpl(
+        new JobConf(serConf.value),
+        new TaskAttemptID("", 0, TaskType.REDUCE, 0, 0)))
+
+    // Write data
+
```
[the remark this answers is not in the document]
osopardo1: It should be used in the process, yes. Solved

polsm91 (approved):

src/main/scala/io/qbeast/spark/QbeastTable.scala
```diff
@@ -65,17 +73,32 @@ class QbeastTable private (
...
+   */
+  def compact(revisionID: RevisionID): Unit = {
+    checkRevisionAvailable(revisionID)
+    CompactTableCommand(revisionID, indexedTable)
       .run(sparkSession)
       .map(_.getString(0))
   }
 
+  def compact(): Unit = {
+    compact(latestRevisionAvailableID)
```
alexeiakimov, line 99: I like this approach, it would be nice to apply it to the analyze methods as well. I mean that operation without revision identifier calls the operation with revision passing the latest one.

src/main/scala/io/qbeast/spark/delta/QbeastMetadataSQL.scala
```diff
@@ -0,0 +1,37 @@
...
+import io.qbeast.core.model.{CubeId, NormalizedWeight, Weight}
+import io.qbeast.spark.utils.TagColumns
+import org.apache.spark.sql.Column
+import org.apache.spark.sql.expressions.UserDefinedFunction
+import org.apache.spark.sql.functions.{col, struct, udf}
+
+object QbeastMetadataSQL {
+  val weight: UserDefinedFunction = udf((weight: Int) => Weight(weight))
+
+  val norm: UserDefinedFunction = udf((mw: Weight, elementCount: Long, desiredSize: Int) =>
```
alexeiakimov, line 15: Could normWeight or normalizedWeight be a better name?

src/main/scala/io/qbeast/spark/delta/writer/Compactor.scala
```diff
@@ -0,0 +1,95 @@
...
+    schema: StructType,
+    cubeId: CubeId,
+    cubeBlocks: IISeq[QbeastBlock],
+    tableChanges: TableChanges)
+    extends Serializable {
+
+  def writeBlock(it: Iterator[InternalRow]): Iterator[FileAction] = {
+
+    val minWeight = cubeBlocks.map(_.minWeight).min
+    val maxWeight =
```
alexeiakimov, line 44: It is not critical, but I wonder if both min and max can be computed traversing the sequence just once.

src/main/scala/io/qbeast/spark/delta/writer/Compactor.scala
```diff
@@ -0,0 +1,95 @@
...
+      cubeBlocks.map(_.maxWeight).min // we pick the lesser updated value
+    val state = tableChanges.cubeState(cubeId).getOrElse(State.FLOODED)
+    val rowCount = cubeBlocks.map(_.elementCount).sum
```
alexeiakimov, line 47: Sum can also be computed with min and max traversing the blocks just once.

src/main/scala/io/qbeast/spark/delta/writer/Compactor.scala
```diff
@@ -0,0 +1,95 @@
...
+      partitionValues = Map(),
+      size = fileStatus.getLen,
+      modificationTime = fileStatus.getModificationTime,
+      dataChange = true,
+      stats = "",
+      tags = tags)
+
+    // Add the file to the commit log and remove the old ones
+    // Better to check all the metadata associated
+    (Seq(addFile) ++ cubeBlocks.map(b =>
```
alexeiakimov, line 91: The removed files can also be collected together with min, max, rowCount traversing the blocks just once.

src/main/scala/io/qbeast/spark/delta/writer/SparkDeltaDataWriter.scala
```diff
@@ -0,0 +1,172 @@
...
+   * Based on groupFilesIntoBins from Delta Lake OptimizeTableCommand
+   * @param cubeStatuses the cubes statuses present in the index
+   * @return
+   */
+  def groupFilesToCompact(
+      cubeStatuses: IISeq[(CubeId, IISeq[QbeastBlock])]): IISeq[(CubeId, IISeq[QbeastBlock])] = {
+
+    // Check what cubes are suitable for compaction
+    val cubesToCompact = cubeStatuses
+      .map(cubeStatus => {
```
alexeiakimov, line 73: I wonder if it is possible to apply pattern matching here like
```scala
.map(case (cubeId, cubeBlocks) =>
...
)
```

src/main/scala/org/apache/spark/qbeast/config.scala
```diff
@@ -26,6 +26,18 @@ package object config {
       .intConf
       .createWithDefault(2)
 
+  private[config] val min_file_size_compaction: ConfigEntry[Int] =
```
alexeiakimov, line 29: Please use camel style for member names.

src/main/scala/org/apache/spark/qbeast/config.scala
```diff
@@ -26,6 +26,18 @@ package object config {
...
+    ConfigBuilder("spark.qbeast.compact.minFileSize")
+      .version("0.2.0")
+      .intConf
+      .createWithDefault(1024 * 1024 * 1024)
+
+  private[config] val max_file_size_compaction: ConfigEntry[Int] =
```
alexeiakimov, line 35: Please use camel style for member names.

src/main/scala/io/qbeast/spark/QbeastTable.scala
osopardo1: I did the same for analyze() and optimize(). Doesn't it appear on the changes?

src/main/scala/io/qbeast/spark/delta/QbeastMetadataSQL.scala
osopardo1: Yep or normalizeWeight, since it's a method

src/main/scala/io/qbeast/spark/delta/writer/Compactor.scala
osopardo1: It could be possible. Solved on 3d7ad28

src/main/scala/io/qbeast/spark/delta/writer/Compactor.scala
osopardo1: Same, all in 3d7ad28

src/main/scala/org/apache/spark/qbeast/config.scala
osopardo1: Solved on 93af690. Changed all the other member names as well.

alexeiakimov (approved):

PR #113, opened 2022-06-23
Space limit

### LINKED ISSUE #111: Query fails to retrieve elements that are in the 1.0 space
## What went wrong?
After indexing Github dataset by `repo_main_language` and `year`, we started doing queries for testing tolerance in each group.

In some of those queries, elements were missing. For example:

```scala
df_qbeast.where("""repo_main_language == "SAS" and year == 2022""").count()
```

was not equal to 

```scala
df_parquet.where("""repo_main_language == "SAS" and year == 2022""").count()
```
In a more deep search, we found out that `year == 2022` query was translated on a space` (from = 1.0, to = 1.0)`. The condition to select or not a cube to retrieve, is:

```scala
  private def intersects(f: Double, t: Double, cube: CubeId, coordinate: Int): Boolean = {
    val cf = cube.from.coordinates(coordinate)
    val ct = cube.to.coordinates(coordinate)
    (f <= cf && cf < t) || (cf <= f && f < ct)
  }
```
So, in the case of f == t == 1.0, the answer is always false. 

## How to reproduce?

1. Read and write the Github Archive Dataset
```scala

df.write.format("qbeast").option("columnsToIndex", "repo_main_language,year").save("path")

```
2. Read the data frame and run:
```scala
val df_qbeast = spark.read.format("qbeast").load("path")
val df_parquet = spark.read.format("parquet").load("path")

df_qbeast.where("""repo_main_language == "SAS" and year == 2022""").count()
df_parquet.where("""repo_main_language == "SAS" and year == 2022""").count()

```

### 2. Branch and commit id:

main bb080839e03d5cebd75dceefa2e0f5c0e824e06c
### 3. Spark version:

3.1.2

### 4. Hadoop version:

3.2.0

### 6. Stack trace:

Jiaweihu08:
When filtering with the max value from an indexed column, the QuerySpace filtering doesn't address the case because the right limit of a cube is always excluded.

The issue is addressed now by adding an additional condition where if the range left limit and the cube right limit are both 1.0, then we take the cube.

Fixes #111 

Test added to `QueryExecutorTest.scala`

* Spark Version: 3.1.1.
* Hadoop Version: 3.2.0
* Local test

### DIFF
--- a/core/src/main/scala/io/qbeast/core/model/QuerySpace.scala
+++ b/core/src/main/scala/io/qbeast/core/model/QuerySpace.scala
@@ -41,7 +41,7 @@ class QuerySpaceFromTo(private val from: Seq[Option[Double]], private val to: Se
   private def intersects(f: Double, t: Double, cube: CubeId, coordinate: Int): Boolean = {
     val cf = cube.from.coordinates(coordinate)
     val ct = cube.to.coordinates(coordinate)
-    (f <= cf && cf < t) || (cf <= f && f < ct)
+    (f <= cf && cf < t) || (cf <= f && f < ct) || (f == 1.0 && ct == 1.0)
   }
 
   override def intersectsWith(cube: CubeId): Boolean = {
--- a/core/src/test/scala/io/qbeast/core/model/QuerySpaceFromToTest.scala
+++ b/core/src/test/scala/io/qbeast/core/model/QuerySpaceFromToTest.scala
@@ -20,9 +20,59 @@ class QuerySpaceFromToTest extends AnyFlatSpec with Matchers {
     querySpaceFromTo.intersectsWith(cube) shouldBe true
   }
 
-  it should "detect a wrong intersection" in {
-    val from = Seq(Some(2))
+  it should "not intersect with query space larger than the max value" in {
+    val from = Seq(Some(3))
     val to = Seq(Some(4))
+    val transformation = Seq(LinearTransformation(1, 2, IntegerDataType))
+    val querySpaceFromTo = QuerySpaceFromTo(from, to, transformation)
+
+    val cube = CubeId.root(3)
+    querySpaceFromTo.intersectsWith(cube) shouldBe false
+  }
+
+  it should "not intersect with query space smaller than the min value" in {
+    val from = Seq(Some(-1))
+    val to = Seq(Some(0))
+    val transformation = Seq(LinearTransformation(1, 2, IntegerDataType))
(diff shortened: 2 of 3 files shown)


Jiaweihu08:
This is the failed test. The QuerySpace boundary built from `from,to` is 1.0 and 3.0.
With the changes I made, the test would fail.
Is it possible for a `QuerySpace` to extend beyond 1.0? If the answer is NO then we would rewrite the test.
```
val from = Seq(Some(2))
val to = Seq(Some(4))
val transformation = Seq(LinearTransformation(1, 2, IntegerDataType))
val querySpaceFromTo = QuerySpaceFromTo(from, to, transformation)

val cube = CubeId.root(3)
querySpaceFromTo.intersectsWith(cube) shouldBe false
```

osopardo1:
> This is the failed test. The QuerySpace boundary built from `from,to` is 1.0 and 3.0. With the changes I made, the test would fail. Is it possible for a `QuerySpace` to extend beyond 1.0? If the answer is NO then we would rewrite the test.
> 
> ```
> val from = Seq(Some(2))
> val to = Seq(Some(4))
> val transformation = Seq(LinearTransformation(1, 2, IntegerDataType))
> val querySpaceFromTo = QuerySpaceFromTo(from, to, transformation)
> 
> val cube = CubeId.root(3)
> querySpaceFromTo.intersectsWith(cube) shouldBe false
> ```

Mmh, good question. Having a requested query space > 1.0 means that the user asked for data outside the indexed space. I believe that case should be possible.

Jiaweihu08:
> > This is the failed test. The QuerySpace boundary built from `from,to` is 1.0 and 3.0. With the changes I made, the test would fail. Is it possible for a `QuerySpace` to extend beyond 1.0? If the answer is NO then we would rewrite the test.
> > ```
> > val from = Seq(Some(2))
> > val to = Seq(Some(4))
> > val transformation = Seq(LinearTransformation(1, 2, IntegerDataType))
> > val querySpaceFromTo = QuerySpaceFromTo(from, to, transformation)
> > 
> > val cube = CubeId.root(3)
> > querySpaceFromTo.intersectsWith(cube) shouldBe false
> > ```
> 
> Mmh, good question. Having a requested query space > 1.0 means that the user asked for data outside the indexed space. I believe that case should be possible.

Then the result `shouldBe true`?

osopardo1 (approved):
Looks good! Thanks for adding new tests :)

Issue #5, opened 2021-08-16
Include INSERT INTO in Qbeast save contract


osopardo1:
**What went wrong?**
Right now on the master, we cannot add new elements to a qbeast table through the INSERT INTO sql command

**How to reproduce?**
1. Code that triggered the bug, or steps to reproduce:

On a previously indexed dataframe named df:

```scala 
df.createOrReplaceTempView("t")
spark.sql("insert into table t (value) values (4)")
```

And the output is: 

```
org.apache.spark.sql.AnalysisException: QbeastBaseRelation(parquet) does not allow insertion.;
'InsertIntoStatement Relation[value#1297] QbeastBaseRelation(parquet), false, false
+- LocalRelation [col1#1543]

  at org.apache.spark.sql.execution.datasources.PreWriteCheck$.failAnalysis(rules.scala:512)

```

2. Branch and commit id:

master ef6b9f45ac54ab156e5b3474e3014b639b2ac827

3. Spark version:
3.x

4. Hadoop version:

2.4.7

5. Are you running Spark inside a container? Are you launching the app on a remote K8s cluster? Or are you just running the tests in a local computer?
Local

6. Stack trace:

eavilaes:
Solved by #116

PR #116, opened 2022-07-15
Support SQL statement INSERT INTO with Qbeast format

Adricu8:
## Description

- Fixing this [issue](https://github.com/Qbeast-io/qbeast-spark/issues/5)

- We have removed the class QbeastBaseRelation. Now we only make use of its object to instantiate a BaseRelation (or HadoopFsRelation)
In this way, the function createRelation is implemented in object QbeastBaseRelation and the insertable relation is not lost.
Moreover, we do not need the ReplaceFileIndex extension and has also been removed. LogicalRelation now makes use of a HadoopFsRelation instead of QbeastBaseRelation.

Now we are able to use SQL statement INSERT INTO like:

`spark.sql("insert into table t select * from t2")`

We had to modify more code to make the following statement work aswell:

`spark.sql("insert into table t (value) values (4)")`

- We have to support IdentityTransformation to be merged with a qbeast table when we insert a single value. Right now we only check the case where we merge a new LinearTransformation. We added the logic to LinearTransformation and LinearTransformer files, modifying the necessary objects to make it work.

It would be nice to create a test with all the possible insert statements that spark can handle
https://spark.apache.org/docs/3.1.2/sql-ref-syntax-dml-insert-into.html
Furthermore, we might want to test that the implemented logic is correct.

When we use this syntax we get an error that the transformation is unknown. Instead of a Linear transformation the transformation type is: io.qbeast.core.transform.IdentityTransformation.

## Type of change

Describe the change you're making: how it affects the API, user experience...

- We are fixing a bug, qbeast format should be able to process INSERT INTO sql statements.

## Checklist:

- [x] Add tests.

- Test QbeastInsertIntoTest: We run the different INSERT INTO SQL statements and assert that the final table containes the expected DataFrame.
- Test LinearTransformationTest: I updated this test to assert that the merge function works as intended when merging an IdentityTransformation

**Test Configuration**:
* Spark Version: 3.1.1
* Hadoop Version: 3.2
* Local environment (M1)


osopardo1:
Hey @Adricu8 , it seems like the description of the PR is not updated, could you explain more about the issue? Thanks!

Adricu8:
Yes of course, I updated the description. Let me know if there is anything unclear.

src/main/scala/io/qbeast/spark/internal/sources/QbeastDataSource.scala
```diff
@@ -26,6 +26,18 @@ import org.apache.spark.sql.{
   SparkSession
 }
 
+// import org.apache.hadoop.fs.{Path}
```
osopardo1, line 29: Comments here as well

src/main/scala/io/qbeast/spark/internal/sources/QbeastBaseRelation.scala
```diff
@@ -38,9 +47,51 @@ object QbeastBaseRelation {
...
+
+    val spark = SparkSession.active
+    val tableID = table.tableID
+    val snapshot = QbeastContext.metadataManager.loadSnapshot(tableID)
+    val schema = QbeastContext.metadataManager.loadCurrentSchema(tableID)
+    val revision = snapshot.loadLatestRevision
+    val columnsToIndex = revision.columnTransformers.map(row => row.columnName).mkString(",")
+    val cubeSize = revision.desiredCubeSize
+    val parameters = Map[String, String](
+      "path" -> table.tableID.toString(),
```
osopardo1, line 66: I'm not sure now if the path is an option that should be added as well, can we verify it somehow?

src/main/scala/io/qbeast/spark/internal/sources/QbeastBaseRelation.scala
```diff
@@ -3,29 +3,38 @@
...
+import io.qbeast.spark.delta.OTreeIndex
+import org.apache.spark.sql.execution.datasources.HadoopFsRelation
+import io.qbeast.spark.table.IndexedTable
+import io.qbeast.context.QbeastContext
+// import io.qbeast.spark.utils.SparkToQTypesUtils
+// import org.apache.spark.sql.execution.datasources.FileFormat
+
+import org.apache.hadoop.fs.{Path}
+import org.apache.spark.sql.catalyst.catalog.{BucketSpec}
+// import org.apache.spark.sql.{
```
osopardo1, line 30: More comments hehe

core/src/main/scala/io/qbeast/core/transform/Transformation.scala
```diff
@@ -47,7 +47,7 @@ trait OrdinalTransformation extends Transformation {
 /**
  * Identity transformation.
  */
-object IdentityTransformation extends Transformation {
+case class IdentityTransformation(newVal: Any) extends Transformation {
```
osopardo1, line 50: Instead of `newVal,` would it be best if we call it `identityVal`?

src/main/scala/io/qbeast/spark/internal/sources/QbeastBaseRelation.scala
Adricu8: As an option in createRelation function?

core/src/main/scala/io/qbeast/core/transform/Transformation.scala
Adricu8: Since the class name relates to the IdentityTransformation and contains the "identity" string I think it is a bit redundant, if you think otherwise we can change it.

core/src/main/scala/io/qbeast/core/transform/LinearTransformation.scala
```diff
@@ -139,7 +167,17 @@ class LinearTransformationSerializer
...
+   * Creates a LinearTransformation that has random value for the nulls
+   * within the [minNumber, maxNumber] range
+   * @param minNumber
+   * @param maxNumber
+   * @param orderedDataType
+   * @param seed
+   * @return
+   */
+
+  def generateRandomNumber(min: Any, max: Any, seed: Option[Long]): Any = {
```
osopardo1, line 180: probably one of these two functions needs to get deleted, right?

core/src/main/scala/io/qbeast/core/transform/Transformation.scala
osopardo1: It may be redundant, but newVal does not define what is this variable. We can name it `value` maybe, but it's not new.

core/src/main/scala/io/qbeast/core/transform/LinearTransformation.scala
osopardo1: Or maybe it's best to put this method on an object `TransformationUtils`, so we can use it in different parts of the code and does not affect codecov

src/main/scala/io/qbeast/spark/internal/sources/QbeastBaseRelation.scala
osopardo1: Yep. Try to delete it from the parameters if you have time, and see if you encounter any problem.

core/src/main/scala/io/qbeast/core/transform/Transformation.scala
Adricu8: Okay, then let's be a bit redundant since it wont harm us anyway. I will rename it to identityVal

src/main/scala/io/qbeast/spark/internal/rules/SampleRule.scala
```diff
@@ -42,13 +45,12 @@ class SampleRule(spark: SparkSession) extends Rule[LogicalPlan] with Logging {
   private def transformSampleToFilter(
       sample: Sample,
       logicalRelation: LogicalRelation,
-      qbeastBaseRelation: QbeastBaseRelation): Filter = {
+      columnsToIndex: IndexedColumns): Filter = {
```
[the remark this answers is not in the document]
Adricu8: Not sure about this, following the other named parameters in [package.scala](https://github.com/Qbeast-io/qbeast-spark/blob/a34feaea97911c52a253b77ae06a66ca9c1a3609/core/src/main/scala/io/qbeast/package.scala) they start with caps, should we change the others aswell?

core/src/main/scala/io/qbeast/core/transform/LinearTransformation.scala
Adricu8: Good idea, I have created an object `LinearTransformationUtils` with the said function

src/main/scala/io/qbeast/spark/internal/sources/QbeastBaseRelation.scala
Adricu8: I deleted it and there is no error so far

core/src/test/scala/io/qbeast/core/transform/LinearTransformationTest.scala
```diff
@@ -76,4 +76,29 @@ class LinearTransformationTest extends AnyFlatSpec with Matchers {
...
+
+    otherNullValue = LinearTransformationUtils.generateRandomNumber(-100, 10000, Option(42.toLong))
+    linearT.merge(IdentityTransformation(-100)) shouldBe LinearTransformation(
+      -100,
+      10000,
+      otherNullValue,
+      IntegerDataType)
+
+    otherNullValue = LinearTransformationUtils.generateRandomNumber(0, 10000, Option(42.toLong))
+    linearT.merge(IdentityTransformation(10)) shouldBe LinearTransformation(
```
osopardo1, line 98: I think this test is not correct.

Because if an `IdentityTransformation` falls in the `LinearTransformation` min-max, then the existing `LinearTransformation` is used. The `merge` method is called only if `isSuperseededBy` returns true. 

Probably, in this case, we should test how `isSuperseededBy` behaves.

src/main/scala/io/qbeast/spark/internal/QbeastSparkSessionExtension.scala
```diff
@@ -20,9 +20,9 @@ class QbeastSparkSessionExtension extends DeltaSparkSessionExtension {
       new SampleRule(session)
     }
 
-    extensions.injectOptimizerRule { session =>
-      new ReplaceFileIndex(session)
-    }
+    // extensions.injectOptimizerRule { session =>
```
osopardo1, line 23: Let's remove this :)

src/main/scala/io/qbeast/spark/internal/rules/SampleRule.scala
```diff
@@ -5,14 +5,17 @@ package io.qbeast.spark.internal.rules
 
 import io.qbeast.core.model.Weight
 import io.qbeast.spark.internal.expressions.QbeastMurmur3Hash
-import io.qbeast.spark.internal.sources.QbeastBaseRelation
+// import io.qbeast.spark.internal.sources.QbeastBaseRelation
```
osopardo1, line 8: And remove this one as well

src/main/scala/io/qbeast/spark/internal/sources/QbeastDataSource.scala
```diff
@@ -16,7 +16,7 @@ import org.apache.spark.sql.sources.{
   DataSourceRegister,
   RelationProvider
 }
-import org.apache.spark.sql.types.StructType
+import org.apache.spark.sql.types.{StructType}
```
osopardo1, line 19: The braces are not needed if we only import one class, better to remove them

osopardo1 (changes requested):

core/src/test/scala/io/qbeast/core/transform/LinearTransformationTest.scala
Adricu8: You are right, the logic in isSupersededBy is not tested by this. I will add a test in `TransformerTest` for this.

osopardo1:
Looks good to me! Just add simple documentation on the README and that's it!

osopardo1 (approved):

PR #118, opened 2022-07-18
Modify index metrics

Jiaweihu08:
Modify index metrics computation. Some others statistics are added since during many experiments they are found useful.

Changes made:
- Add `L1`, `L2` errors to measure how inner cube sizes deviate from the `desiredCubeSize`, values are between 0 and 1.
- For inner cubes, add a method to display average normalizedWeight and cube size per level
- Compute `depthOnBalance` using geometrics sum since many indexes don't have a high number of levels
- Add test for single cube tree

How to use:
```scala
import io.qbeast.spark.QbeastTable

val path = "yourTablePath"
val metrics = QbeastTable.forPath(spark, path).getIndexMetrics()

println(metrics)
```

* Spark Version: 3.1.3
* Hadoop Version: 3.2
* Cluster or local? Local

### DIFF
--- a/docs/QbeastTable.md
+++ b/docs/QbeastTable.md
@@ -39,7 +39,7 @@ qbeastTable.optimize() // optimizes the cubes
 
 You can use it to **compare the index** build **with different indexing parameters** such as the `desiredCubeSize` and `columnsToIndex`.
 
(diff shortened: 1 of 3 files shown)


osopardo1:
Looks good! The only missing part is to document what are those metrics in the` docs/QbeastTable.md` file

osopardo1 (approved):

PR #125, opened 2022-08-18
Infer parameters when appending

### LINKED ISSUE #81: ColumnsToIndex and/or CubeSize should be optional on Append
When writing data with `qbeast` format, the user needs to specify every time the `columnsToIndex` or `cubeSize`. This is ok if you want to change them, but it shouldn't be always explicit.

For example, if the user wants to append data to an existing table and maintain the same configuration, it should be able to write:

```scala
df.write.format("qbeast").save("existing-path")
```

instead of
```scala
df.write.format("qbeast").mode("append").option("columnsToIndex", "x,y,z").option("cubeSize", "10000").save("existing-path")
```

Jiaweihu08:
## Description
Fixes #81.

As described in #81, we want the `Append` operations to existing qbeast tables to be able to omit providing values for `desiredCubeSize` and `columnsToIndex`, if there's no desire to create a new revision.

Instead of
```
df.write
  .mode("append")
  .format("qbeast")
  .option("cubeSize", cs)
  .option("columnsToIndex", cti)
  .save("path-to-existing-qtable")
```
we want to be able to just do the following instead:
```
df.write
  .mode("append")
  .format("qbeast")
  .save("path-to-existing-qtable")
```

If the user was to provide any value for `cubeSize` and/or `columnsToIndex` when appending, then they will be respected. If the values provided are different than that from the last revision, a new revision should be created for the Append operation.

My assumption for this PR is that , at the moment, we are only aware of revision changes when `desiredCubeSize` changes, not when `columnsToIndex` is changed. The changes made has considerations when the 'desiredCubeSize' is modified.

## Type of change

This change is an `enhancement`.

- Fixes #81 

- Adds a new feature - Appending to existing table without providing indexing params when no revision change is wanted.

- No **Breaking change** introduced

- No documentation change required

## Checklist:

- [x] Add tests.

## Tests

When appending, tests for the following scenarios are added:
- User provide no value for neither `cubeSize` or `columnsToIndex`:
```
df.write
  .mode("append")
  .format("qbeast")
  .save("path-to-existing-qtable")
```
- Only `cubeSize` is provided, regardless of the last `cubeSize` value:
```
df.write
  .mode("append")
  .format("qbeast")
  .option("cubeSize", cs)
  .save("path-to-existing-qtable")
```
- Only `columnsToIndex` is provided. It has **NOT** been tested when the values are different than that from the last revision:
```
df.write
  .mode("append")
  .format("qbeast")
  .option("columnsToIndex", cti)
  .save("path-to-existing-qtable")
```

**Test Configuration**:
* Spark Version: `3.1.3`
* Hadoop Version: `3.2`
* Cluster or local? `Local`

### DIFF
--- a/src/main/scala/io/qbeast/spark/internal/sources/QbeastDataSource.scala
+++ b/src/main/scala/io/qbeast/spark/internal/sources/QbeastDataSource.scala
@@ -65,9 +65,11 @@ class QbeastDataSource private[sources] (private val tableFactory: IndexedTableF
       mode: SaveMode,
       parameters: Map[String, String],
       data: DataFrame): BaseRelation = {
-    require(
-      parameters.contains("columnsToIndex"),
-      throw AnalysisExceptionFactory.create("'columnsToIndex is not specified"))
+    if (mode != SaveMode.Append) {
+      require(
+        parameters.contains("columnsToIndex"),
+        throw AnalysisExceptionFactory.create("'columnsToIndex' is not specified"))
+    }
     val tableId = SparkToQTypesUtils.loadFromParameters(parameters)
     val table = tableFactory.getIndexedTable(tableId)
     mode match {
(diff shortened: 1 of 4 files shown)


osopardo1:
Looks good!

osopardo1 (approved):

Issue #127, opened 2022-08-29
Add support for indexing type Date


osopardo1:
Currently, we only support indexing Numeric types (`Long`, `Int`, `Double`, `BigDecimal`) and `String` types. 

A typical use case consists of dividing the datasets per date to retrieve records from a period more easily. Qbeast should support indexing by this type of column information.

 A possible solution in the meantime is to index the columns by hashing the value:

```scala
df.write.format("qbeast").option("columnsToIndex", "date:hashing").save("path")
```

Or casting first the column to a Long:

```scala
df.withColumn("date", unix_timestamp(col("date"))).format("qbeast").option("columnsToIndex", "date:hashing").save("path")
```

Issue #128, opened 2022-08-30
Exception on cubeId on DF with single row


polsm91:
## What went wrong?

Indexing a DataFrame on a column of type Int and with a single row throws an exception. It seems related to how we estimate the space with Integers, since I can't reproduce the bug with Strings.

### 1. Code that triggered the bug, or steps to reproduce:

```scala
// This fails
List(95).toDF.write.format("qbeast").option("columnsToIndex","value").save("/tmp/debug-assert")
// While this works
List("95").toDF.write.format("qbeast").option("columnsToIndex","value").save("/tmp/debug-assert2")
```

### 2. Branch and commit id:
Pull request: https://github.com/Qbeast-io/qbeast-spark/pull/124
Branch: `42-table-writes-catalog`
Commit id: 8894e91a94e6e595cd702d9d5194f094713d91ee

### 3. Spark version:
3.1.3

### 4. Hadoop version:
3.2.0

### 5. How are you running Spark?
Local

### 6. Stack trace:

```java
java.lang.IllegalArgumentException: requirement failed
        at scala.Predef$.require(Predef.scala:268)
        at io.qbeast.core.model.CubeId$.containers(CubeId.scala:66)
        at io.qbeast.core.model.CubeWeightsBuilder.resultInternal(CubeWeights.scala:101)
        at io.qbeast.core.model.CubeWeightsBuilder.result(CubeWeights.scala:92)
        at io.qbeast.spark.index.DoublePassOTreeDataAnalyzer$.$anonfun$estimatePartitionCubeWeights$4(OTreeDataAnalyzer.scala:161)
        at org.apache.spark.sql.execution.MapPartitionsExec.$anonfun$doExecute$3(objects.scala:195)

```

Adricu8:
The same error happens when you index a DF with multiple rows of the same value.

osopardo1:
As @Adricu8 pointed out, there's a bug in `IdentityTransformation` (when a column indexed has a unique value). This value is mapped as `value.byteValue()`, that can give you a Double up to 2^8. An easy solution is to map all values for that transformation to 0.0 (as we do for nulls).

PR #129, opened 2022-08-30
Adding support for indexing Date data types

Adricu8:
## Description

Adding support for indexing *Datetime* data type in qbeast-spark.
In reference to [issue](https://github.com/Qbeast-io/qbeast-spark/issues/127).

## Type of change

It is an enhancement that would allow for Datetime data type to be indexed without the need of a workaround like casting the value to a Long type or similar.
Spark has [two](https://spark.apache.org/docs/latest/sql-ref-datatypes.html) date data types, we could add support for both.

More elaborated explanation and guidelines regarding the addition of new types [here](https://www.notion.so/qbeastanalytics/How-to-add-new-Types-to-Qbeast-Spark-dd88cb80759040b5a0ae4019581ee29a)

## Checklist:

- [x] Add tests.

## How Has This Been Tested?

- TransformerTest: I have added a test to check the correct behavior of LinearTransformation with Timestamps
- TransformerIndexingTest: I have added a test to ensure timestamps read and write correctly with spark

**Test Configuration**:
* Spark Version: 3.1.1
* Hadoop Version: 3.2
* Running Locally on M1


core/src/main/scala/io/qbeast/core/transform/LinearTransformation.scala
```diff
@@ -185,6 +180,7 @@ class LinearTransformationDeserializer
       case (LongDataType, long: NumericNode) => long.asLong
       case (FloatDataType, float: DoubleNode) => float.floatValue
       case (DecimalDataType, decimal: DoubleNode) => decimal.asDouble
+      case (TimestampDataType, long: NumericNode) => long.asLong
```
osopardo1, line 183: I will name the variable `timestamp` instead of `long` . I think it will make it more readable.

core/src/test/scala/io/qbeast/core/transform/TransformerTest.scala
```diff
@@ -40,6 +42,28 @@ class TransformerTest extends AnyFlatSpec with Matchers {
...
+    val transformer = Transformer(columnName, dataType)
+
+    val minTimestamp = Timestamp.valueOf("2017-01-01 12:02:00")
+    val maxTimestamp = Timestamp.valueOf("2017-01-02 12:02:00")
+
+    val minTimestampLong = minTimestamp.getTime
+    val maxTimestampLong = maxTimestamp.getTime
+
+    // If I do not print the values it complains that variables are not used
+    Console.print(minTimestampLong)
```
osopardo1, line 57: This should not be the case, since you are using them below. Maybe was something related to sbt configuration?

core/src/main/scala/io/qbeast/core/transform/Transformation.scala
```diff
@@ -53,6 +54,9 @@ case class IdentityTransformation(newVal: Any) extends Transformation {
   override def transform(value: Any): Double = value match {
     case v: Number =>
       v.byteValue()
+    // Not sure that we need this
+    case v: Timestamp =>
+      v.getTime().byteValue()
```
osopardo1, line 59: Try to test a case with a single timestamp. It will fail now, but the resolution of #128 could give us a hint on this.

core/src/main/scala/io/qbeast/core/model/QDataType.scala
```diff
@@ -52,29 +53,43 @@ trait OrderedDataType extends QDataType {
...
   override def name: String = "DecimalDataType"
+
   override val ordering: Numeric[Any] = implicitly[Numeric[Double]].asInstanceOf[Numeric[Any]]
 }
 
 object StringDataType extends QDataType {
   override def name: String = "StringDataType"
 }
+
+object TimestampDataType extends OrderedDataType {
```
osopardo1, line 88: Since we are adding this in the core, we must also add a test in that package. You can give a look at `JSONSerializationTest`, `LinearTransformationTest` and `TransformerTest` to cover this.

osopardo1:
Well done! Just a few more comments/questions:
1. We could put something more user-friendly in the title of the PR. For example: `Support for indexing Date columns`.  Then everyone can understand what this is about. Because Transformer and QDataTypes are something more dependent on implementation rather than functionality. :man_technologist: 
2. I guess for the implementation we are only adding Timestamp. But you have intentions to add a Date as well, right?

core/src/test/scala/io/qbeast/core/transform/TransformerTest.scala
Adricu8: I don't know why this happens, I think it is related to the variables being used inside the matchPattern. If I remove the print I get this: 
`local val minTimestampLong in value <local TransformerTest> is never used
 val minTimestampLong = minTimestamp.getTime`

core/src/main/scala/io/qbeast/core/transform/Transformation.scala
Adricu8: I get the same error as in Issue [#128](https://github.com/Qbeast-io/qbeast-spark/issues/128). Should we create a new PR to solve it or can we try to solve it in this one?

core/src/test/scala/io/qbeast/core/transform/TransformerTest.scala
osopardo1: Instead of asserting with matchPattern, try with the clause` shouldBe`. I don't know much about the matchPattern behaviour.

core/src/main/scala/io/qbeast/core/transform/Transformation.scala
osopardo1: Yes, I will create a new PR. Hope we can merge it today and you can synch quickly.

core/src/main/scala/io/qbeast/core/model/QDataType.scala
Adricu8: I added more tests for `JSONSerializationTest` and `TransformerTest`.

core/src/main/scala/io/qbeast/core/model/QDataType.scala
```diff
@@ -52,29 +59,50 @@ trait OrderedDataType extends QDataType {
...
 
 object FloatDataType extends OrderedDataType {
   override def name: String = "FloatDataType"
+
   override val ordering: Numeric[Any] = implicitly[Numeric[Float]].asInstanceOf[Numeric[Any]]
 }
 
 object DecimalDataType extends OrderedDataType {
   override def name: String = "DecimalDataType"
+
```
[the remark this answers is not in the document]
Adricu8: yes, it must be a scalafmt change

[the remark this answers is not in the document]
osopardo1: I think we should delete them if we can, not to make more changes than the necessary ones.

core/src/test/scala/io/qbeast/core/transform/TransformerTest.scala
Adricu8: I changed the test implementation, It feels way more complicated than it should because the object does not have methods implemented to get the parameters of the class. It works but I would change it if we figure out a better way to do it.

core/src/test/scala/io/qbeast/core/transform/TransformerTest.scala
```diff
@@ -40,6 +42,50 @@ class TransformerTest extends AnyFlatSpec with Matchers {
...
+    val resTransformation = transformer.makeTransformation(transformation)
+
+    // Find all integers in the string separated by commas and remove characters
+    val regex: Regex = "\\d+".r
+    val resMinMax = regex.findAllIn(resTransformation.toString).toList.map(_.toLong)
+    resMinMax(0) shouldBe minTimestamp.getTime
+    resMinMax(1) shouldBe maxTimestamp.getTime
+    // Get the the datatype of the transformation
+    val resDataType = resTransformation.toString.split(",")(3).dropRight(1)
+    resDataType shouldBe dataType.toString
```
Adricu8, line 63: I changed the test implementation, It feels way more complicated than it should because the object does not have methods implemented to get the parameters of the class. It works but I would change it if we figure out a better way to do it.

osopardo1: It's okay, but maybe you don't have to parse everything to string and use regex. You can instantiate the result transformation as a LinearTransformation, and access the corresponding methods from there.

Like:
`val resTransformation = transformer.makeTransformation(transformation).asInstanceOf[LinearTransformation]`

Adricu8: Yeah, that is a much better solution

osopardo1:
Looks good!

osopardo1 (approved):

PR #130, opened 2022-08-31
Fixed indexing columns with unique values

### LINKED ISSUE #128: Exception on cubeId on DF with single row

osopardo1:
## Description

Fixes #128. 

## Type of change

When indexing a column with the same value, an IdentityTransformation is instantiated. A Transformation is in charge of mapping the original values of the column to a [0.0, 1.0] space. In the case of IdentityTransformation, the method was returning the `value.byteValue()`, which can give you a `Double` up to 2^8. 

We solved this by changing `IdentityTransformation` to `IdentityToZeroTransformation` and returning 0.0 if the number corresponds to the transformation value.

## Checklist:

- [x] Add tests.

- I changed one line of an existing test because this was badly designed. It checked if qbeast could index a column with the same value, but that value was 1. So, when transforming the value, it was in the [0.0, 1.0] range and no requirement failed. I change this value to 10.

### DIFF
--- a/core/src/main/scala/io/qbeast/core/transform/LinearTransformation.scala
+++ b/core/src/main/scala/io/qbeast/core/transform/LinearTransformation.scala
@@ -77,7 +77,7 @@ case class LinearTransformation(
           otherNullValue,
           orderedDataType)
           .asInstanceOf[Transformation]
-      case IdentityTransformation(newVal) =>
+      case IdentityToZeroTransformation(newVal) =>
         val otherNullValue =
           LinearTransformationUtils.generateRandomNumber(
             min(minNumber, newVal),
@@ -104,7 +104,7 @@ case class LinearTransformation(
       case LinearTransformation(newMin, newMax, _, otherOrdering)
           if orderedDataType == otherOrdering =>
         gt(minNumber, newMin) || lt(maxNumber, newMax)
-      case IdentityTransformation(newVal) =>
+      case IdentityToZeroTransformation(newVal) =>
         gt(minNumber, newVal) || lt(maxNumber, newVal)
     }
 
--- a/core/src/main/scala/io/qbeast/core/transform/Transformation.scala
+++ b/core/src/main/scala/io/qbeast/core/transform/Transformation.scala
@@ -47,12 +47,11 @@ trait OrdinalTransformation extends Transformation {
 /**
  * Identity transformation.
  */
-case class IdentityTransformation(newVal: Any) extends Transformation {
+case class IdentityToZeroTransformation(identityValue: Any) extends Transformation {
 
   @inline
   override def transform(value: Any): Double = value match {
-    case v: Number =>
-      v.byteValue()
+    case v: Number if v == identityValue => 0.0
   }
 
   override def isSupersededBy(newTransformation: Transformation): Boolean = false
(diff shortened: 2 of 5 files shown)


core/src/main/scala/io/qbeast/core/transform/Transformation.scala
```diff
@@ -47,12 +47,11 @@ trait OrdinalTransformation extends Transformation {
...
  */
-case class IdentityTransformation(newVal: Any) extends Transformation {
+case class IdentityToZeroTransformation(identityValue: Any) extends Transformation {
 
   @inline
   override def transform(value: Any): Double = value match {
-    case v: Number =>
-      v.byteValue()
+    case v: Number if v == identityValue => 0.0
   }
```
Adricu8, line 55: Is the `if v == identityValue` really needed?

core/src/main/scala/io/qbeast/core/transform/LinearTransformation.scala
```diff
@@ -77,7 +77,7 @@ case class LinearTransformation(
           otherNullValue,
           orderedDataType)
           .asInstanceOf[Transformation]
-      case IdentityTransformation(newVal) =>
+      case IdentityToZeroTransformation(newVal) =>
         val otherNullValue =
           LinearTransformationUtils.generateRandomNumber(
             min(minNumber, newVal),
```
Adricu8, line 83: I understand the name change comes from the fact that the IdentityTransformation will always map the values to zero for the index by default, right?

Adricu8 (commented):
Good job, the changes seem pretty straightforward :)

core/src/main/scala/io/qbeast/core/transform/LinearTransformation.scala
osopardo1: Yes, it's purely a naming convention :smile:

core/src/main/scala/io/qbeast/core/transform/Transformation.scala
osopardo1: I'm not entirely sure, but if this transformation is trying to transform a value that is not the original one, it should, at least, be noticed (because that can lead to the bad behavior of the index).

Adricu8 (approved):

Issue #68, opened 2022-01-27
Overhead of qbeast_hash filtering when doing a Sample


osopardo1:
**What went wrong?**
When a sample() is performed against a qbeast dataset, the qbeast sql extension changes the Sample operation into a Filter to:
- Pushdown the filter to the data source
- Filter the cubes that satisfy the [lowerBound, upperBound]
- In the end, read less data

But in this process, we are producing a hash in memory of the columns indexed to make sure that the records matched with the Sample range. 
This hash has an extra cost when we don't need the information of those columns to satisfy the query. For example, if we have a dataset indexed by `(user_id,product_id)` and we want to know the avg `price`, we are **avoiding the pruning of `user_id, product_id` columns**. 

We need to study how to overcome this problem. 
One solution is to avoid the in-memory filtering for those files that have a `maxWeight < upperBound`, because we know for sure that all the records in that file match the predicate.

Adricu8:
We can just use Random.nextInt() to implement the qbeast_hash function and forget about the indexed columns (only for writing). The complex part comes when we read the data. Two options could be to either write a random weight per cube (user invasive) or create the random weight in memory to use it to filter the data per cube.

osopardo1:
Due to the complexity of adding these changes to the code, we will postpone the solution of this issue.

Keep in mind that the only case affected on performance is when reading a SAMPLE of files.

PR #134, opened 2022-09-21
Use random weights instead of hash

Adricu8:
## Description

WIP on issue #68 

We want to change the current implementation of how the weights are written. Using a hash on the indexed columns both creates an overhead and does not provide the desired randomness in some cases. E.g. the cardinality of the indexed columns will affect the quality of samples, especially when queries do not relate to the indexed columns.
We want to use scala.Random.nextInt() as the weight to ensure randomness. However, this change introduces challenges on how we can filter the data. We need to ensure the randomness is spread across cubes in a correct way.

## Type of change

Bug fix to ensure the correct randomness of a sample operator.
This change can potentially improve the performance of the algorithm significantly, but we are still unsure that it will work as we expect. 

As explained in #68, the idea is to use scala.random.nextInt() instead of the qbeast_hash() function per row weight. The issue arises when we read the data. We need to select the rows for each cube, which could be done by knowing the min/max of each cube and selecting the amount of data according to the desired sample and specific min/max per cube.

**Note**: This is a WIP and there are some parts that I am unsure how to implement. I am not currently working on this PR, any suggestions are welcome.

## Checklist:

- [ ] Add tests.

## How Has This Been Tested?
TODO

**Test Configuration**:
* Spark Version: 3.1.1
* Local environment: M1


osopardo1:
I would like close this PR for now, we will open it again when it's necessary. Is it okay for you @Adricu8?

Adricu8:
> I would like close this PR for now, we will open it again when it's necessary. Is it okay for you @Adricu8?

Sure! Let's close it

osopardo1:
I reopen this PR since, after some experiments with real data, we find out that it's a more critical aspect to improve on the open source.

Issue #135, opened 2022-10-04
The same CubeId with different hashCodes


Jiaweihu08:
## What went wrong?

The same cube created using two different methods can have different `bitMask`'s. This leads to different hashCodes, making `CubeId` an unreliable candidate as `Map` keys.

## How to reproduce?

```scala
val level65 =
      CubeId(2, "wQwwwQwwQwwQwwwwwQwQwwwQQwwwwQQwQwwwQwwQwwQwwwwwQwwQQQQQQQQQQQQQQ")
      
// cubes from level 64
val cubeThroughParentMethod = level65.parent.get
val cubeThroughConstructor =
      CubeId(2, "wQwwwQwwQwwQwwwwwQwQwwwQQwwwwQQwQwwwQwwQwwQwwwwwQwwQQQQQQQQQQQQQ")

println(cubeThroughParentMethod.bitMask) // Array(-1441504016541484037, -6148914603906580738, 0, 0)
println(cubeThroughConstructor.bitMask) // Array(-1441504016541484037, -6148914603906580738)
```

### 2. Branch and commit id:
`main`, [880203a](https://github.com/Qbeast-io/qbeast-spark/commit/880203addde1cff4a74f0b6b0aece324d900d560)

### 5. How are you running Spark?
Local

alexeiakimov:
Good catch, please do not forget to fix `equals` as well. Thank you.

PR #136, opened 2022-10-07
135 the same cubeid with different hashcodes

alexeiakimov:
## Description

Fixes [The same CubeId with different hashCodes](https://github.com/Qbeast-io/qbeast-spark/issues/135).
The PR proposes to change implementations of equals and `hashCode` of the `CubeId` class so the `bitMask` arrays are compared without possible trailing zeros. The same approach is implemented for `CubeId.isAncestor` method.

A note on performance: removing trailing zeros for every call to `equals`, `hashCode`, and `isAncestor` seems to be cheap enough, because for the majority of CubeId instances do not have trailing zeros in their `bitMask`, in such cases the new `trimBitMask` function returns the same array on discovering that the last element of the array is not 0.

@osopardo1 @Jiaweihu08 could you please review, thank you.

## Type of change

This is a bug fix. The existing API was not modified, no new API was added.

## How Has This Been Tested?

The examples to reproduce the bug were added to existing tests of the affected CubeId methods.

### DIFF
--- a/core/src/test/scala/io/qbeast/core/model/CubeIdTest.scala
+++ b/core/src/test/scala/io/qbeast/core/model/CubeIdTest.scala
@@ -23,6 +23,19 @@ class CubeIdTest extends AnyFlatSpec with Matchers {
     id1 == id2 shouldBe false
     id3 == id4 shouldBe false
     id1 == id5 shouldBe false
+    val id6 =
+      CubeId(2, "wQwwwQwwQwwQwwwwwQwQwwwQQwwwwQQwQwwwQwwQwwQwwwwwQwwQQQQQQQQQQQQQQ").parent.get
+    val id7 =
+      CubeId(2, "wQwwwQwwQwwQwwwwwQwQwwwQQwwwwQQwQwwwQwwQwwQwwwwwQwwQQQQQQQQQQQQQ")
+    id6 == id7 shouldBe true
+  }
+
+  it should "implement hashCode correctly" in {
+    val id1 =
+      CubeId(2, "wQwwwQwwQwwQwwwwwQwQwwwQQwwwwQQwQwwwQwwQwwQwwwwwQwwQQQQQQQQQQQQQQ").parent.get
+    val id2 =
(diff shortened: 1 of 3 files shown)


Jiaweihu08 (approved):

osopardo1 (approved):

core/src/test/scala/io/qbeast/core/model/CubeIdTest.scala
```diff
@@ -23,6 +23,19 @@ class CubeIdTest extends AnyFlatSpec with Matchers {
     id1 == id2 shouldBe false
     id3 == id4 shouldBe false
     id1 == id5 shouldBe false
+    val id6 =
```
cugni, line 26: Probably a bit overkilling, but maybe a few more tests would be good to be sure we are covering all the cases: when the trim aligns with a long, when it is just a bit shorter or just a bit longer.

cugni (commented):
Wouldn't it be faster, simpler and safer to use Scala's Vectors instead of Array? I understand Vectors have the [slice](https://www.scala-lang.org/api/2.12.8/scala/collection/immutable/Vector.html#slice(from:Int,until:Int):scala.collection.immutable.Vector[A]). 
They also implement their hashcode safely.

cugni (approved):
We discussed whether it is best to use a CubeId composed of an array of long or a BitMask. The trade-off is that using a BitMask internally and serializing it to binary in Spark is cleaner from our code point of view but makes the CubeId an utterly opaque object for Spark, making it impossible to use SQL operations on top of that. We are not sure yet which is the best approach, but there is no strong reason now to change the actual implementation.

alexeiakimov:
Could you please help me to understand how a patch in CubeId can decrease the code coverage in BlockWriter in the code which does not use CubeId directly. What should I do to fix the coverage problem? Thank you.

osopardo1:
> Could you please help me to understand how a patch in CubeId can decrease the code coverage in BlockWriter in the code which does not use CubeId directly. What should I do to fix the coverage problem? Thank you.

As I suggested before, I think you can ignore `codecov` since they seem they are not related issues (I don't know why is behaving like that). Nevertheless, we can open an issue to add a test for this specific case.

Issue #137, opened 2022-10-19
[Delta] Register operationMetrics for table history


osopardo1:
## What went wrong?
The underlying format Delta Lake is capable of registering metrics of the different operations (WRITE, UPDATE, DELETE)... When writing with Qbeast Format, since we change the pipeline, we missed adding this information.

These metrics are useful to monitor the performance and analyze the behavior of the operations allowed. Here you can find the documentation: 
- Schema https://docs.databricks.com/delta/delta-utility.html#history-schema 
- Metrics https://docs.databricks.com/delta/delta-utility.html#operation-metrics-keys

## How to reproduce?

Different steps about how to reproduce the problem.

### 1. Code that triggered the bug, or steps to reproduce:

You can create or load test data into a dataFrame and write it using both Delta and Qbeast.

```scala
    val tmpDir = "/tmp/dir"
    val data = spark.read
    .format("csv")
    .option("header", "true")
    .option("inferSchema", "true")
    .load("src/test/resources/ecommerce100K_2019_Oct.csv")

    data.write.format("qbeast").option("columnsToIndex", "user_id,price").save(tmpDir + "/qbeast")
    data.write.format("delta").save(tmpDir + "/delta")

```

When retrieving the history of a Qbeast Table, the output shows some empty fields:
```scala
spark.sql(s"DESCRIBE HISTORY '$tmpDir/qbeast'").select("operationMetrics").show(false)
+----------------+
|operationMetrics|
+----------------+
|{}              |
+----------------+

```

But when analyzing a Delta Table:

```scala
spark.sql(s"DESCRIBE HISTORY '$tmpDir/delta'").select("operationMetrics").show(false)
+------------------------------------------------------------------+
|operationMetrics                                                  |
+------------------------------------------------------------------+
|{numFiles -> 8, numOutputRows -> 99986, numOutputBytes -> 4303479}|
+------------------------------------------------------------------+

```
### 2. Branch and commit id:
`main` 

### 3. Spark version:
On the spark shell run `spark.version`.

3.2.2

### 4. Hadoop version:
On the spark shell run `org.apache.hadoop.util.VersionInfo.getVersion()`.

3.2.1

### 5. How are you running Spark?
Are you running Spark inside a container? Are you launching the app on a remote K8s cluster? Or are you just running the tests in a local computer?

Local spark
### 6. Stack trace:
Trace of the log/error messages.

NA

osopardo1:
Here are my current thoughts:

[...]

PR #139, opened 2022-11-03
Add Operation and File Metrics [Delta]

osopardo1:
## Description

Fixes issues #137 and #4 

## Type of change

The goal was to add **new information to the Commit Log entries**. On Delta Lake, **statistics at the operation level and file level** are getting collected. Since Qbeast has a modified implementation of the writing process, we were skipping this part. 

In this PR, we use `JobStatsTrackers` (spark processes that keep track of specific metrics when managing data) to fill in stats information. As a result, the `CommitInfo` and the `AddFile` present in the commit log, should contain the following fields:

```json
{
  "commitInfo": {
    ...
    "operationMetrics": {
      "numFiles": "1",
      "numOutputRows": "3",
      "numOutputBytes": "660"
    },
   ...
}

```

```json
 "add": {
   ...
    "stats": "{\"numRecords\":3,\"minValues\":{\"a\":\"A\",\"b\":1},\"maxValues\":{\"a\":\"C\",\"b\":3},\"nullCount\":{\"a\":0,\"b\":0}}",
   ...
  }
}

```

UPDATE.
After doing an initial benchmark, we will be leaving the data-skipping strategy for another Pull Request. We saw that reading and filtering twice the Delta Log (as a naïve solution) is creating an overhead on the queries. We should think of a better way os processing this information, but in the meantime, I think it's more important to have this metadata in the commit log. 

## Checklist:

- [x] Add tests.

Created a test under `QbeastDataSourceIntegrationTest` class in which we make sure the information is commited correctly. 
Since these additional measurements are going to impact performance, we should also benchmark this againts version 0.3.1.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,12 @@ project/plugins/project/
 .worksheet
 .idea
 
+# VSCode specific
+.bloop/
+.metals/
+.vscode/
+metals.sbt
+
 # publication project
 nodep
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -125,7 +125,7 @@ For example:
 sbt assembly
 
 $SPARK_HOME/bin/spark-shell \
---jars ./target/scala-2.12/qbeast-spark-assembly-0.3.0-alpha.jar \
+--jars ./target/scala-2.12/qbeast-spark-assembly-0.3.0.jar \
 --conf spark.sql.extensions=io.qbeast.spark.internal.QbeastSparkSessionExtension \
 --packages io.delta:delta-core_2.12:1.2.0
 ```
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -125,7 +125,7 @@ For example:
 sbt assembly
 
 $SPARK_HOME/bin/spark-shell \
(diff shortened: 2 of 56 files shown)


osopardo1:
On the new commit [8e7f304](https://github.com/Qbeast-io/qbeast-spark/pull/139/commits/8e7f304124f3b15cb7320f71a745a0a16c652ab6), I added a workaround to filter files using the Stats. 

Since Delta Lake has already implemented a `DataSkippingReader` strategy on the `Snapshot` class, we would use their method to filter according to the gathered stats. Once we had their filtering and ours, we can join the outputs with the file paths. 

This is a first naive implementation. We should think about something more efficient, since we are duplicating the number of collects on the Delta Log, and we can simplify the data skipping with one single process.

osopardo1:
After doing an initial benchmark, I think it's best to leave the data-skipping strategy for another Pull Request. We saw that reading and filtering twice the Delta Log (as a naïve solution) is creating an overhead on the queries. We should think of a better way of processing this information, but in the meantime, probably it's more important to have this metadata in the commit log.

Jiaweihu08 (approved):

PR #141, opened 2022-11-29
Improved double pass

Jiaweihu08:
## Description

One of the main issues of the current OTree implementation is its failure to enforce desiredCubeSize on the tree nodes. As shown in the image below, the deeper we go down the tree, the larger the cube sizes increase.

This leads to an undesired outcome, namely, files that contain an unpredictable number of records with very large sizes. The aim of this PR is to address this issue.

A comparison with the existing DoublePass implementation is shown in the following image. As we can see, the average cube sizes are just as good as the sequential implementation. The dataset used is derived from the GitHub Archive project dataset; The index is built on two columns but all columns are used for weight computation:

Average cube size per level  All columns used for weight computation

## How is this achieved?
Instead of merging partition-level `NormalizedWeight` via Harmonic Means to obtain cube weights, the improvements are achieved by estimating weights so that the expected value of the node size is that of the `desiredCubeSize`.

### Existing cubes
Given a cube `c` with domain `Nc`, knowing the weight of its parent node `Wpc`, find the weight of c, `Wc`, so the amount of `Nc` we leave in c has an expected value of `desiredCubeSize`, in other words `desiredCubeSize = (Wc - Wpc) • |Nc|`. Weights can now be computed using `Wc = Wpc + desiredCubeSize / |Nc|`.

The domain of a cube, `Nc`, is the dataset records that fall within the limits of c. Generally, `Nc` is distributed among the line of ancestors of c, and the subtree of c (including c). 

The subtree of c, `Dc`, is where the remaining of Nc is found, so `|Dc| = (1 - Wpc) • |Nc|`; The local domain of a cube can be found via `|Nc,i|  = |Dc,i| / (1 - Wpc,i)`. And the global domain is to be found via a simple sum: **|Nc| = $\sum_{} |Nc,i|$**.

This requires two changes:
1. The computation of `Dc` for the local trees, with is the sum of c's payload size and the tree size of its child cubes.
2. The computation of global weights `Wc` has to be done in a top-down fashion, for `Wpc` is needed in each step. Having, Wpc, find |Nc| for c in all local trees via `|Nc,i|  = |Dc,i| / (1 - Wpc,i)`.

####Missing cubes
What if c is not present in a given local tree? Since partition data are different, we can expect local trees to have different topologies. The domain of c has to be computed differently in this case because |Dci| = 0.

The idea is to find a good upper-bound of the missing cube domain in the local tree. Since c is not present, we know that all its payload is left in its chain of parent nodes. **The approach we've taken is to compute the fraction of each ancestor payload that belongs to the missing cube region**.

Given a cube A with child cubes A1, A2, and A3. The fraction of A's payload that belongs to the A3 domain can be estimated by `f = |A3| / (|A1| + |A2| + |A3|)`. In doing so, we are assuming the distribution of A's payload is the same as that of its child cube tree sizes.

Now, we can compute `f` for each existing ancestor and extract the correct fraction from their payload and add it to the domain. `f_acc` at each ancestor going upwards is updated by `f`: `f_acc  = f_acc * f`. The fraction `f` from a given c is the fraction for its immediate child cube, which then has to be multiplied by the existing `f_acc` to find its contribution to the target domain. 

For example, given the following ancestor chain for an missing cube: **Root -> A -> AB -> ABC -> MissingCube**. We initiate the cube domain as zero and add the corresponding payload fraction from each ancestor cube, updating `f_acc` at each step:

1. f_acc = f1 = f(ABC)
2. f2 = f(AB), f_acc *= f2, domain += f_acc * |AB|
3. f3 = f(A), f_acc *= f3, domain += f_acc * |A|
4. f4 = f(Root), f_acc *= f4, domain += f_acc * |Root|

### Summary
- Cube weights are now computed through the computation of cube domains
- Since partition data are different, different local trees have different topologies
- If the cube is present in a local tree, this can be computed using a cube's tree size
- Otherwise, an estimation of the cube domain upper-bound is done, and we do so by estimating the domain of our missing cube has left in each ancestor cube

### Potential improvements
- [ ] Do we need to improve the efficiency of the `populateTreeSize` method, used to compute local cube tree sizes?
- [ ] Merge local trees. Currently we are creating a SortedSet from all LocalTree keys, and use this to traverse branches. For each distinct cube we have to go through all localtrees to extract |Nc,i|.
- [ ] Unlike the existing implementation where a Seq[CubeNormalizedWeight] is created for each partition, Map[CubeId, (NormalizedWeight, TreeSize)] is used as LocalTree  for each partition. This increases the size of our collect from each partition, do we need to make this less memory consuming?

### DIFF
--- a/core/src/main/scala/io/qbeast/core/model/CubeWeights.scala
+++ b/core/src/main/scala/io/qbeast/core/model/CubeWeights.scala
@@ -63,7 +63,7 @@ class CubeWeightsBuilder protected (
 
   private val byWeight = Ordering.by[PointWeightAndParent, Weight](_.weight).reverse
   protected val queue = new mutable.PriorityQueue[PointWeightAndParent]()(byWeight)
-  private var resultBuffer = Seq.empty[CubeNormalizedWeight]
+  private var resultBuffer = Seq.empty[LocalTree]
 
   /**
    * Updates the builder with given point with weight.
@@ -77,7 +77,8 @@ class CubeWeightsBuilder protected (
   def update(point: Point, weight: Weight, parent: Option[CubeId] = None): CubeWeightsBuilder = {
     queue.enqueue(PointWeightAndParent(point, weight, parent))
     if (queue.size >= bufferCapacity) {
-      resultBuffer ++= resultInternal()
+      val unpopulatedTree = resultInternal()
+      resultBuffer = resultBuffer :+ populateTreeSize(unpopulatedTree).toMap
       queue.clear()
     }
     this
@@ -88,11 +89,16 @@ class CubeWeightsBuilder protected (
    *
    * @return the resulting cube weights map
    */
-  def result(): Seq[CubeNormalizedWeight] = {
-    resultInternal() ++ resultBuffer
+  def result(): Seq[LocalTree] = {
+    val unpopulatedTree = resultInternal()
+    if (unpopulatedTree.nonEmpty) {
+      resultBuffer :+ populateTreeSize(unpopulatedTree).toMap
+    } else {
+      resultBuffer
+    }
   }
 
-  def resultInternal(): Seq[CubeNormalizedWeight] = {
+  def resultInternal(): mutable.Map[CubeId, CubeInfo] = {
     val weights = mutable.Map.empty[CubeId, WeightAndCount]
     while (queue.nonEmpty) {
       val PointWeightAndParent(point, weight, parent) = queue.dequeue()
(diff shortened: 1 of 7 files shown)


src/main/scala/io/qbeast/spark/index/OTreeDataAnalyzer.scala
```diff
@@ -160,8 +142,171 @@ object DoublePassOTreeDataAnalyzer extends OTreeDataAnalyzer with Serializable {
...
+
+    // Unique cubes among all local trees
+    val globalCubes = spark.sparkContext.broadcast(
+      localTrees
+        .select(explode(map_keys(col("value"))))
+        .select(col("col.*"))
+        .as[CubeId]
+        .distinct()
+        .rdd
+        .sortBy(identity)
```
cugni, line 221: why are you sorting?

src/main/scala/io/qbeast/spark/index/OTreeDataAnalyzer.scala
```diff
@@ -160,8 +142,171 @@ object DoublePassOTreeDataAnalyzer extends OTreeDataAnalyzer with Serializable {
...
+  private[index] def computeGlobalDomain(
+      localTrees: Dataset[LocalTree],
+      isReplication: Boolean,
+      dimensionCount: Int,
+      desiredCubeSize: Int): Seq[(CubeId, Double)] = {
+    val spark = SparkSession.active
+    import spark.implicits._
+
+    // Unique cubes among all local trees
+    val globalCubes = spark.sparkContext.broadcast(
```
cugni, line 214: A collect, to do a broadcast and then use it in another dataframe doesn't seem very efficient. Can we do the same with a JOIN (e.g. an outer join)?

cugni (commented):
Did you test how much slower is the new way algorithm for creating the CubeWeightMap? I have the impression it might be much slower, and that we can optimize it in many ways.

src/main/scala/io/qbeast/spark/index/OTreeDataAnalyzer.scala
Jiaweihu08: The domain of a missing cube is computed using that of its parent, so the computation has to be done from top-down for each branch

core/src/main/scala/io/qbeast/core/model/CubeWeights.scala
```diff
@@ -113,15 +119,50 @@ class CubeWeightsBuilder protected (
...
+   */
+  def populateTreeSizeAndParentWeight(
+      cubeMap: mutable.Map[CubeId, CubeInfo]): mutable.Map[CubeId, CubeInfo] = {
+    // Compute tree sizes and parent weights from bottom up.
+    val levelCubes = cubeMap.keys.groupBy(_.depth)
+    val minLevel = levelCubes.keys.min
+    val maxLevel = levelCubes.keys.max
+    (maxLevel until minLevel by -1) foreach { level =>
+      levelCubes(level)
+        .groupBy(c => c.parent.get)
```
alexeiakimov, line 149: Do you really need to group the keys by parent? It seems that grouping by level is enough to ensure that the cubes are processed in the right order. Below every sibling only contributes to it's parent weight, so mixing the cubes from the same level should not be a problem. It would make the algorithm simpler to understand and will reduce a bit the memory consumption. The price is more CubeInfo lookups.

cugni:
Hi @Jiaweihu08, what's the current status of this PR? I  understand that you tested the writing time, and it is good,  but the writing was somehow slower because now we actually respect the desired cubeSize. Maybe @victorcuevasv can help you with the benchmarking? I would really love to see this merged :)

cugni:
Hi @Jiaweihu08, what's the current status of this PR? Does the simplified version in  https://github.com/Qbeast-io/qbeast-spark/pull/156 supersede it, or will we still add the last optimizations? What is missing?

Jiaweihu08:
> Hi @Jiaweihu08, what's the current status of this PR? Does the simplified version in #156 supersede it, or will we still add the last optimizations? What is missing?

I think this one should come after PR #156, both are pending to be tested.
#156 works fine for most cases and this one can be useful as an additional analyzer strategy when #156 isn't sufficient.

Jiaweihu08:
> > Hi @Jiaweihu08, what's the current status of this PR? Does the simplified version in #156 supersede it, or will we still add the last optimizations? What is missing?
> 
> I think this one should come after PR #156, both are pending to be tested. #156 works fine for most cases and this one can be useful as an additional analyzer strategy when #156 isn't sufficient.

The development for this PR is paused until #156 is finished, and will be added as an additional configurable analyzer strategy.

PR #143, opened 2022-12-12
Unified Catalog

osopardo1:
## Description

In this PR, we are modifying the `QbeastCatalog` to process also writes in Delta (or other formats such Iceberg or Hudi once they're integrated). 

## Type of change

It is an improvement of an existing Catalog feature. Instead of managing two catalogs at the same time, the user would only need to set up one by adding the following configuration to any Spark Application:

`--conf spark.sql.catalog.spark_catalog=io.qbeast.spark.internal.sources.catalog.QbeastCatalog`

Then, the user can create `qbeast` tables and `delta` tables with either DataFrame API or SQL:

```scala
data.write.format("delta").saveAsTable("delta_table")

data.write
 .format("qbeast")
 .option("columnsToIndex", "id")
 .saveAsTable("qbeast_table")
```

```scala
spark.sql(
  s"CREATE TABLE student (id INT, name STRING, age INT) USING qbeast " +
    "OPTIONS ('columnsToIndex'='id')")

spark.sql(
  s"CREATE TABLE student (id INT, name STRING, age INT) USING delta")

```

**The user can also choose to configure them as separate catalogs**. But for some integrations (such as `dbt`) it's not recommended.

## Checklist:

- [x] Add tests.

We add a test in `QbeastCatalogIntegrationTest` in which we try to write as `delta` and `qbeast` with the same Catalog.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -125,7 +125,7 @@ For example:
 sbt assembly
 
 $SPARK_HOME/bin/spark-shell \
---jars ./target/scala-2.12/qbeast-spark-assembly-0.3.0-alpha.jar \
+--jars ./target/scala-2.12/qbeast-spark-assembly-0.3.0.jar \
 --conf spark.sql.extensions=io.qbeast.spark.internal.QbeastSparkSessionExtension \
 --packages io.delta:delta-core_2.12:1.2.0
 ```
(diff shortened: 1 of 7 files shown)


osopardo1:
Hello! I would suggest @Adricu8 and @eavilaes try the solution in your DBT project. 
You can do an `sbt assembly` on the code and test it. This assembly would contain both `qbeast-spark` and `qbeast-core` packages. I will also put you as reviewers to verify everything works.
Thank you!

Jiaweihu08 (approved):

PR #144, opened 2022-12-20
Filter revisions through the creation of empty query spaces

Jiaweihu08:
## Description
The construction of IndexStatus when the entire Revision is out of query space is a waste.
The changes from this PR filter the Revisions out of the query space before their corresponding index construction from the log and cube filtering.

## Type of change
**Improvement for filtering**

## Checklist:
- [x] Add tests.

- Test 1: Modify QuerySpaceFromToTest to skip Revisions that create EmptySpace instances for a given query. Revisions that create AllSpace instances don't need to traverse the index; they can return all FLOODED blocks directly.
- Test 2: QuerySpecBuilderTest: Filter QuerySpaces that are not QuerySpaceFromTo

**Test Configuration**:
* Spark Version: 3.1.2
* Hadoop Version: 3.2
* Cluster or local: Local

### DIFF
--- a/core/src/test/scala/io/qbeast/core/model/QuerySpaceFromToTest.scala
+++ b/core/src/test/scala/io/qbeast/core/model/QuerySpaceFromToTest.scala
@@ -6,7 +6,7 @@ import org.scalatest.matchers.should.Matchers
 
 class QuerySpaceFromToTest extends AnyFlatSpec with Matchers {
 
-  val toSpaceCoordinates
+  private val toSpaceCoordinates
       : (Seq[Option[Any]], Seq[Transformation]) => Seq[Option[NormalizedWeight]] =
     (originalValues: Seq[Option[Any]], transformations: Seq[Transformation]) => {
       originalValues.zip(transformations).map {
@@ -30,7 +30,7 @@ class QuerySpaceFromToTest extends AnyFlatSpec with Matchers {
   }
 
   it should "exclude beyond left limit" in {
-    val cube = CubeId.root(3)
+    val cube = CubeId.root(1)
     val transformation = Seq(LinearTransformation(1, 2, IntegerDataType))
     val from = toSpaceCoordinates(Seq(Some(-1)), transformation)
     val to = toSpaceCoordinates(Seq(Some(0)), transformation)
@@ -40,7 +40,7 @@ class QuerySpaceFromToTest extends AnyFlatSpec with Matchers {
   }
 
   it should "exclude beyond right limit" in {
-    val cube = CubeId.root(3)
+    val cube = CubeId.root(1)
     val transformation = Seq(LinearTransformation(1, 2, IntegerDataType))
     val from = toSpaceCoordinates(Seq(Some(3)), transformation)
     val to = toSpaceCoordinates(Seq(Some(4)), transformation)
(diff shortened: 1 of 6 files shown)


core/src/test/scala/io/qbeast/core/model/QuerySpaceFromToTest.scala
```diff
@@ -1,100 +1,116 @@
...
 
-import io.qbeast.core.transform.LinearTransformation
+import io.qbeast.core.transform.{LinearTransformation, Transformation}
 import org.scalatest.flatspec.AnyFlatSpec
 import org.scalatest.matchers.should.Matchers
 
 class QuerySpaceFromToTest extends AnyFlatSpec with Matchers {
 
-  "QuerySpace from to" should "intersect correctly" in {
+  val toSpaceCoordinates
```
osopardo1, line 9: Better to declare this variable as private, since you are not using it outside the class.

Jiaweihu08: I should change the variable names to columnsFrom and columnsTo, since the usage of _ as a prefix is a Java convention of bean getters/setters, and its usage as suffixes is discouraged:

Quoting from [Scala style guide](https://docs.scala-lang.org/style/naming-conventions.html):
> The underscore is in the prefix position rather than the suffix to avoid any danger of mistakenly typing name _ instead of name_.

src/main/scala/io/qbeast/spark/index/query/QuerySpecBuilder.scala
```diff
@@ -92,37 +93,35 @@ private[spark] class QuerySpecBuilder(sparkFilters: Seq[Expression]) extends Ser
...
+
+    val (from, to) =
       indexedColumns.map { columnName =>
         // Get the filters related to the column
         val columnFilters = filters.filter(hasColumnReference(_, columnName))
 
         // Get the coordinates of the column in the filters,
         // if not found, use the overall coordinates
-        val from = columnFilters
+        val from_ = columnFilters
```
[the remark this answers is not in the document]
Jiaweihu08: I should change the variable names to columnsFrom and columnsTo, since the usage of _ as a prefix is a Java convention of bean getters/setters, and its usage as suffixes is discouraged:

Quoting from [Scala style guide](https://docs.scala-lang.org/style/naming-conventions.html):
> The underscore is in the prefix position rather than the suffix to avoid any danger of mistakenly typing name _ instead of name_.

osopardo1:
One small thing before approving this PR. Do you think it's possible to test also the query times to see if there's an improvement in skipping revisions?
Maybe we could use the append TPC-DS workload (which should output more Revisions) and see the difference on reading times. 

If you don't think it's necessary, I'm ready to approve it anyways.

Jiaweihu08:
> One small thing before approving this PR. Do you think it's possible to test also the query times to see if there's an improvement in skipping revisions? Maybe we could use the append TPC-DS workload (which should output more Revisions) and see the difference on reading times.
> 
> If you don't think it's necessary, I'm ready to approve it anyways.

We can but I don't think we need to; it WILL improve the performance because we are avoiding unnecessary index constructions and tree traversals.

osopardo1:
👍🏻 perfect then

osopardo1 (approved):

PR #146, opened 2022-12-23
Compressing leaves - vanilla DP

Jiaweihu08:
## Description

Having too many small leaf cubes can hinder the read performance, which is the case when the resulting otree index has many levels.

These changes aim to test whether reducing the number of small files by collapsing small leaf cubes to their parent cube and creating a separate block can improve read performance.

## Type of change
- Enhancement

## Checklist:
- [ ] Add tests.

No collapsing:
Screenshot 2022-12-23 at 17 07 30

Collapsing:
Screenshot 2022-12-23 at 17 06 40

**Test Configuration**:
* Spark Version:
* Hadoop Version:
* Cluster or local?

### DIFF
--- a/core/src/main/scala/io/qbeast/core/model/RevisionClasses.scala
+++ b/core/src/main/scala/io/qbeast/core/model/RevisionClasses.scala
@@ -236,7 +236,7 @@ trait TableChanges {
   val isNewRevision: Boolean
   val isOptimizeOperation: Boolean
   val updatedRevision: Revision
-  val compressionMap: Map[CubeId, CubeId]
+  val compressedLeaves: Set[CubeId]
   val deltaReplicatedSet: Set[CubeId]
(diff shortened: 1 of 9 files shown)


Jiaweihu08:
The vanilla version of the DoublePass overestimates NormalizedWeights, resulting in cubes larger than the desiredCubeSize. The leaf nodes we desire to compress, or many leaves in general, may not exist in the resulting index because there are not enough elements to reach those levels. For this reason, this implementation of leaf compression won't work on many occasions, mainly when the data partitions differ significantly in distribution for the indexing columns. If desired, one can repartition the data by the indexing columns and hope for the best.

osopardo1:
After final TPC-DS tests over this PR, we determine that the implementation does not impact the reading performance, neither it behaves correctly with the vanilla version of the code. 
Closing this!

PR #152, opened 2023-01-25
Convert to Qbeast

### LINKED ISSUE #102: Add Convert To Qbeast

Jiaweihu08:
## Description

This PR adds the ability to read a hybrid `qbeast + delta` table using qbeast.

It also introduces the `ConverToQbeastCommand` to qbeast-spark to allow reading a `parquet` or `delta` table without indexing and rewriting. Partitioned table are not supported by this operation.

These features are achieved by putting the non-qbeast `AddFile`s in a staging Revision, which is created during the first qbeast write (including overwrites), or when running the conversion command. The non-qbeast files are at the moment characterized as having null tags, and are all placed in the root of the staging Revision.

During reads, these files are only processed in memory(for now), no filtering at the file level is done for all files are in the root.

```scala
import io.qbeast.spark.internal.commands.ConvertToQbeastCommand

val path = "/pathToTable/"
val tableIdentifier = s"parquet.`$path`"
val columnsToIndex = Seq("col1", "col2", "col3")
val desiredCubeSize = 50000

ConvertToQbeastCommand(tableIdentifier, columnsToIndex, desiredCubeSize).run(spark)
```

The converted table can be read using either `delta` or `qbeast`, and appending to the converted table using `delta` puts the data into the staging Revision without indexing. Appends that use `qbeast` follow the usual **qbeast indexing** procedure.

`Compaction` can be executed on the staging revision to reduce its number of files.

Fixes #102

This PR also makes the following changes:
- Compaction with dataChange=false as default
- No duplicated `RemoveFile` in `DeltaMetadataWriter`

## Type of change

- When first writing or when overwriting a table using qbeast, aside from the creation of the revision it also adds a staging revision. This has `EmptyTransformer`s and `EmptyTransformation`s, and most importantly `RevisionID = 0`
- Documentation update to `QbeastFormat`

## Checklist:
- [x] Add tests.

- The creation of the staging revision during first writes
- The same when running ConvertToQbeatCommand
- Test ConvertToQbeatCommand on delta and parquet tables, partitoned and not partitioned
- Test Compaction on converted table and revisionID = stagingID
- Appending on a converted table creates a now non-staging Revision

**Test Configuration**:
* Spark Version: `3.2.2`
* Hadoop Version: `3.3.1`
* Cluster or local? `Local`


src/main/scala/io/qbeast/spark/delta/IndexStatusBuilder.scala
```diff
@@ -35,41 +37,66 @@ private[delta] class IndexStatusBuilder(
...
   }
 
   /**
    * Returns the index state for the given space revision
    * @return Dataset containing cube information
    */
   def buildCubesStatuses: SortedMap[CubeId, CubeStatus] = {
-
     val spark = SparkSession.active
+    import spark.implicits._
```
osopardo1, line 77: If we didn't change the behaviour for this method, I think it's best to preserve the original state 🙈

src/main/scala/io/qbeast/spark/delta/OTreeIndex.scala
```diff
@@ -62,6 +80,8 @@ case class OTreeIndex(index: TahoeLogFileIndex) extends FileIndex {
         /* modificationTime */ qbeastBlock.modificationTime,
         absolutePath(qbeastBlock.path))
     }.toArray
+    val stagingStats = stagingFiles
```
osopardo1, line 83: Two concerns that I have (we should resonate them here first):

1. Could the `stagingFiles` be partitioned? If that's the case, maybe we should check it and return a sequence of `PartitionDirectory` directly.
2. Should we find a way of let Delta (or the underlying format) deal with the filtering of those files? Since we are collecting twice the log, maybe we could improve the number of files read at least.

src/main/scala/io/qbeast/spark/delta/OTreeIndex.scala
```diff
@@ -80,17 +100,17 @@ case class OTreeIndex(index: TahoeLogFileIndex) extends FileIndex {
...
- * @param sparkSession the spark session
- * @param path the path to the delta log
- * @return the OTreeIndex
- */
 object OTreeIndex {
 
   def apply(spark: SparkSession, path: Path): OTreeIndex = {
     val deltaLog = DeltaLog.forTable(spark, path)
-    val tahoe = TahoeLogFileIndex(spark, deltaLog, path, deltaLog.snapshot, Seq.empty, false)
+    val tahoe = TahoeLogFileIndex(
```
osopardo1, line 107: Is necessary to specify `isTimeTravelQuery`? Since we are not using that for the other parameters, we can leave it as in the main version.

src/main/scala/io/qbeast/spark/delta/QbeastMetadataOperation.scala
```diff
@@ -78,16 +80,28 @@ private[delta] class QbeastMetadataOperation extends ImplicitMetadataOperation {
...
-
     val newRevisionID = newRevision.revisionID
 
-    // Qbeast configuration metadata
-    baseConfiguration
-      .updated(MetadataConfig.lastRevisionID, newRevisionID.toString)
-      .updated(
-        s"${MetadataConfig.revision}.$newRevisionID",
-        mapper.writeValueAsString(newRevision))
+    // Add staging revision if necessary
```
osopardo1, line 85: Adding more comments would help to understand the steps 😄

src/main/scala/io/qbeast/spark/utils/Params.scala
```diff
@@ -40,3 +40,7 @@ object MetadataConfig {
   final val replicatedSet = "qbeast.replicatedSet"
   final val lastRevisionID = "qbeast.lastRevisionID"
 }
+
+object Staging {
```
osopardo1, line 44: Do we need to create an object for this? Or could we place it in another existing class like `QbeastDeltaSnapshot`? 
Because probably each format would have their own isStagingFile predicate.

docs/QbeastFormat.md
```diff
@@ -153,6 +153,36 @@ In Revision, you can find different information about the tree status and config
 
 In this case, we index columns `user_id` and `product_id` (which are both `Integers`) with a linear transformation. This means that they will not suffer any transformation besides the normalization.
 
+### Staging Revision and ConvertToQbeastConvertToQbeastCommand
```
osopardo1, line 156: Careful, ConvertToQbeast is duplicated.

osopardo1:
Hello! I did the first review, feel free to discuss any of my comments in the thread.

One more suggestion: I think it's better to call it Convert To Qbeast (Or incremental conversion to qbeast). Compatibility might be a confusing word (since we are already compatible) and the scope of this changes is actually being able to convert the table. The hybrid support is an inevitable change for the final result.

core/src/main/scala/io/qbeast/core/model/RevisionClasses.scala
```diff
@@ -193,8 +224,10 @@ case class IndexStatus(
     cubesStatuses: SortedMap[CubeId, CubeStatus] = SortedMap.empty)
     extends Serializable {
 
-  def addAnnouncements(newAnnouncedSet: Set[CubeId]): IndexStatus =
-    copy(announcedSet = announcedSet ++ newAnnouncedSet)
+  def addAnnouncements(newAnnouncedSet: Set[CubeId]): IndexStatus = {
```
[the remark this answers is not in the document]
Jiaweihu08: When calling QbeastTable.analyze() on a converted table.

core/src/main/scala/io/qbeast/core/transform/EmptyTransformation.scala
```diff
@@ -0,0 +1,13 @@
+package io.qbeast.core.transform
+
+/**
+ * An empty Transformation meant for empty revisions
+ */
+case class EmptyTransformation() extends Transformation {
```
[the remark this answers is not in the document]
Jiaweihu08: Should we use an invalid value for its transform like -1 or we can leave it to 0?

src/main/scala/io/qbeast/spark/delta/OTreeIndex.scala
Jiaweihu08: It'd leave the part of improving the READ efficiency to another issue

core/src/main/scala/io/qbeast/core/transform/EmptyTransformation.scala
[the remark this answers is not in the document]
Jiaweihu08: Leaving it to 0 for the moment

src/main/scala/io/qbeast/spark/delta/OTreeIndex.scala
osopardo1: Then I will open one, thanks!

core/src/main/scala/io/qbeast/core/model/RevisionClasses.scala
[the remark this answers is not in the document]
osopardo1: Better to skip running `analyze()` on Revision number 0 (as well as .optimize())

osopardo1 (approved):

PR #156, opened 2023-02-06
Simple domain-driven OTree data analyzer

### LINKED ISSUE #155: Cube sizes increase with level
As currently constructed, the cube size of the OTree index does not strictly follow that of the indicated value - `desiredCubeSize`, and generally increases with cube depth. Indexing a large dataset can lead to files with unpredictable sizes and element counts.

The following figure depicts the issue. The dataset used is the `store_sales` table from `TPC-DS`, with a scale factor of 1000. Indexing columns: `ss_sold_date_sk` and `ss_item_sk`. desiredCubeSize: `500.000`.

Average cube size per level(desiredCubeSize = 500k)

The dissimilarity in distribution between data partition is suspected to be the cause. For the issue is alleviated by repartition the data, ideally using the indexing columns, before indexing. We believe a weight estimation algorithm agnostic to partition data distribution can address the issue effectivety.

Jiaweihu08:
## Description
Fixes #155

The modifications to the DoublePass in this PR is the simplified version from #141 - one that doesn't consider the contributions of the missing cubes.

The primary benefit is a more straightforward computation of the cube domains, and consequently that of the estimated cube normalized weights. The results from the previous experiments indicate that, ignoring the missing cubes can potentially lead to slight cube size increases at the deeper levels of the index, e.g. level 10 and onwards.

In comparison with the vainilla double pass, it adds two extra tree traversals - one for local tree size computation done in each worker, and another to estimate the global NormalizedWeights due to the parent - child weight dependency: `Wc = Wp + dcs / domain`.

In comparison with the complete version from #141, it removes the computing and broadcasting of the unique cubes, missing cubes are not taken into consideration. No reference is needed between the cubes from a single tree, so no need to keep tree cubes in a single data structure.

The changes are tested on the `store_sales` table from `TPC-DS` with scale factor = 1000. Indexing on `ss_sold_date_sk` and `ss_item_sk` using a desirec cube size of `50.000` shows promising results. `dddp` stands for domain-driven double pass from this PR, and `dp` is the vainilla version, on the **y-axis** we have the average cube size of the inner cubes on each level.

Avg  inner cube size - TPC-DS

**Questions to answer:**
- [x] **How much overhead are we incurring by computing weights in a top-down fashion?**

We are measuring the time taken to compute the estimated cube weights from a sequence of (CubeId, Domains). It involved grouping values by cube depth and traversing the tree from top to bottom.

The experimental results have shown that, for a final index with about 300 cubes, it takes around 3ms. And for around 4.5 million cubes, the process took approximately 4s.

Generally, we won't encounter cases where such a large number of cubes are being used; we can safely conclude that the overhead in this particular part of the algorithm is negligible.

- [x] **How much is the problem of increasing cube sizes mitigated?**

Aside from the TPC-DS dataset shown above, four additional datasets were indexed, and the results were compared with the vanilla double pass. Except for the GitHub-pr dataset, all other three showed perfect alignment of the cube sizes at each level with the desired value.

dddp-vs-dp-4

- [x] **How much slower(or faster) is this implementation in writing compared to the vanilla version?**

As pointed out earlier, the overhead of the top-down weight estimation from dddp is relatively small, and neglible if the number of cubes is not too large.

The point where the two implementations diverge in time consumption is during row-cube mapping and block writes, for the number of cubes to write and the size of these cubes are different. 

The cubes grow in size with depth for the vanilla version, so the blocks are larger in size and fewer in number. On the other hand, the dddp predicts better max weight for cubes, so inner cubes, in theory, will have similar sizes. The leaf cubes can have sizes ranging from the desired cube size to 1 in both cases, but the number of leaves is more significant for dddp.

Indexing a dataset with **288M** rows using desired cube sizes 5M, 3M, and 1M shown no difference in time for the two implementations. Although cube counts are wildly different for dcs=3M and dcs=1M:
dddp vs  dp for TPC-DS sc=100

When dcs is very small, for instance, 5000, the time taken to index the same dataset is twice as slow for dddp - the size of each cube is so tiny that the preparation for each block write was almost more expensive than the writing itself, and since dddp produced twice as much cubes, the slowdown isn't surprising.

The same experiment is repeated for the same TPC-DS store_sales dataset but with **2880M** records. As we can see in the image below, dddp consistently output more cubes than dp, the different in time is negligible with dddp slightly outperform its vanilla counterpart.

dddp vs  dp - TPC-DS sf=1000

## Type of change
Improve alignment of the cube size with the desired value.

## Checklist:

- [x] Add tests.

## Tests
- Integration with Analyze and Optimize
- Monotonically decreasing cube domains along any tree branch
- Monoonically decreasing estimated weights
- Average cube size per level

**Test Configuration**:
* Spark Version: `3.2.2`
* Hadoop Version: `3.3.1`
* Cluster or local? `local`


core/src/main/scala/io/qbeast/core/model/CubeWeights.scala
```diff
@@ -78,21 +78,29 @@ class CubeWeightsBuilder protected (
...
-   * Builds the resulting cube weights sequence.
+   * Builds the resulting cube domain sequence.
    *
-   * @return the resulting cube weights map
+   * @return the resulting cube domain map
    */
-  def result(): Seq[CubeNormalizedWeight] = {
-    resultInternal() ++ resultBuffer
+  def result(): Seq[CubeDomain] = {
+    resultBuffer ++ resultInternal()
```
osopardo1, line 91: Is this change relevant?

core/src/main/scala/io/qbeast/core/model/CubeWeights.scala
```diff
@@ -144,9 +199,19 @@ private class WeightAndCount(var weight: Weight, var count: Int)
...
- * @param cubeBytes        the cube
- * @param normalizedWeight the weight
+ * NormalizedWeight and tree size of a given cube, with tree size defined as the
+ * sum of its own payload and all that of its descendents. The tree size is
+ * initialized with cube size and updated to the tree size during computation
+ * before being used.
+ * @param weight NormalizedWeight
+ * @param treeSize Cube tree size
+ */
+protected class WeightAndTreeSize(val weight: NormalizedWeight, var treeSize: Double)
```
osopardo1, line 209: Could you review the use of `var` in this class @alexeiakimov ? Do you think is safe?

cugni: I think it is safe if it is a class used only internally. However, why is it protected and not private? I understand **protected** allows access to subclasses. Is that case? Is there any subclass accessing this method? If not, it is better to put it private, right?

cugni:
@victorcuevasv could also run a benchmark with this and the main? I understand here the main things to check are:
1. that writing is not slower when we analyze the tree
2. That we compare this version fairly with the main, as they main will most likely generate much bigger files in the leaf nodes. Shuld we compare both version with the same cubeSize or use a bigger one for this PR? @Jiaweihu08 @victorcuevasv what do you think?

I'm really looking forward to merge this PR!

core/src/main/scala/io/qbeast/core/model/CubeWeights.scala
cugni: By the way, we can replace all protected with private in this class.

Jiaweihu08: I'll change all protected variables and classes to private.

Jiaweihu08:
> @victorcuevasv could also run a benchmark with this and the main? I understand here the main things to check are:
> 
> 1. that writing is not slower when we analyze the tree
> 2. That we compare this version fairly with the main, as they main will most likely generate much bigger files in the leaf nodes. Shuld we compare both version with the same cubeSize or use a bigger one for this PR? @Jiaweihu08 @victorcuevasv what do you think?
> 
> I'm really looking forward to merge this PR!

Writing is not slower when the `desiredCubeSize` used is not too small as mentioned in the description, in fact if the `desiredCubeSize` is 'reasonable' in comparison to the total number of records in the dataset, the write is faster. The same is demonstrated by the last benchmark run by @victorcuevasv.

When it comes to reading, however, it was shown that the new implementation is about 5% slower. This could be due to several factors, like the one you mentioned - the cube size is not homogeneous for the vanilla version but dddp sticks to the `desiredCubeSize`.

Here are two ways of comparing the two versions for reading we may want to explore:
- Use the median inner cube size from the vanilla version for the domain-driven version and then compare the read speed
- Compare the best read speed from the two implementations each with their own best `desiredCubeSize`, similar to hyperparameter fine tuning from ML model training

cugni:
@Jiaweihu08 did you check the reading speed with bigger cubeSize? What values have you used? Were are the results for the benchmarks? You can share it as a Gist.

Jiaweihu08:
> @Jiaweihu08 did you check the reading speed with bigger cubeSize? What values have you used? Were are the results for the benchmarks? You can share it as a Gist.


The vanilla and dddp implementations were compared by writing a denormalized tpc-ds table (2.6B records!) with sf=1000, and running 10 queries on the table four times.

The `desiredCubeSize` for the vanilla version was **8M**, which resulted in an index with a median inner cube size of **10701590**; this value is used for the dddp implementation as the `desiredCubeSize`.

In total the vanilla index has 533 cubes, 133 of which are inner cubes, while dddp created 469 cubes with 117 inner cubes.

Inner cube size quartiles:
imp | min | 1st | 2nd | 3rd | max
-----|-----|-----|-----|-----|-----
vanilla | 7.96M | 10,43M| 10.70M | 11.32M | 12.61M
dddp | 10.67M | 10.70M | 10.70M | 10.71M | 10.73M

Load test:
load-test

Read tests:
read-tests

As shown in the two graphs above, the domain-driven implementation out performs its vanilla counterpart in both write and read tests by 1.65% and 2.25% respectively.

The inner cube sizes for dddp are aligned with the desired value, while for the vanilla implementation the inner cubes range from 8 to 12.5M.

osopardo1:
I think we are ready to merge this changes! Results are good, not only on reading but also the writing is improved.

osopardo1 (approved):

Issue #157, opened 2023-02-06
Add option to set Min/Max values before indexing


Adricu8:
It is an important feature to add to the format to be able to control how the range of the cubes are formed when dealing with streaming or event-based data. The reason is that otherwise we have to create a new revision every time new data arrives (assuming a timestamp column is indexed), and also rewriting the index too frequently can be very expensive. 

We could add the option to customize the min/max values for specific columns.
The following snippet code is just an example on what would be needed:

### Example code

```scala
val df = spark.read.parquet("event-based-data.parquet")
df.write.format("qbeast").options(Map("columnsToIndex" -> "timestamp,other_column", "cubeSize" -> "5000", "min:timestamp" -> "1919191919", "max:timestamp" -> "2020202020")).save("tmpTable")
```

In this way, we can control how often the index will create a new revision on specific use-cases.

osopardo1:
Hello @Adricu8 !!

I propose a different API for adding min and max values:

```scala
df.write.format("qbeast").options(Map("columnsToIndex" -> "timestamp(min:19191919,max:202020),other_column", "cubeSize" -> "5000")).save("tmpTable")
```

Let me know which one do you prefer 😄

Adricu8:
Hi Paola, I think your proposal is nice! :)

osopardo1:
After some research, I will try to implement the option you propose in the beginning!

Because to calculate the `Transformations` (which are the **objects that contain min and max data**) we use an Spark Row that has the following structure:

```scala
"min(columnName)=1, max(columnName)=2"
```

**We can use the same structure in another optional parameter** that users can copy and paste from their spark applications.

```scala
df.write.format("qbeast").options(Map("columnsToIndex" -> "timestamp,other_column", "cubeSize" -> "5000", "stats" -> "(min(timestamp)=1, max(timestamp)=3)").save("tmpTable")
``` 

I am currently exploring this option, sorry for any possible confusion! 

I will come back with the final resolution soon.

PR #160, opened 2023-02-07
API for passing stats

osopardo1:
## Description

Fixes issue #157 . 

## Type of change

This PR includes a new feature on the DataFrame options. 

This new key/value pair allows the user to pass a set of statistics of the columns to index in a JSON formatted string. 

It will allow **a more flexible creation of `Revision`**, in order to **include values that might not be in the newly indexed `Dataframe`**, but that covers more space. For example, if you already know in advance the limits of the values you want to index, you can specify them through the `.option` method.

```scala

df.write.format("qbeast")
.option("columnsToIndex", "a,b")
.option("stats","""{"a_min":0,"a_max":10,"b_min":20.0,"b_max":70.0}""").save("/tmp/table")
```

The enforced structure of the `JSON` is:

```json
{
    "columnName_min" : value
    "columnName_max" : value

}
```

## Checklist:

- [ ] Add tests.

I've created some test under `SparkRevisionFactoryTest` class, which checks if the created `Revision` corresponds to the values introduced by the user.

**Test Configuration**:
* Spark Version: 3.2.2
* Hadoop Version: 3.3.0
* Cluster or local? Local

### DIFF
--- a/src/main/scala/io/qbeast/spark/internal/QbeastOptions.scala
+++ b/src/main/scala/io/qbeast/spark/internal/QbeastOptions.scala
@@ -6,23 +6,30 @@ package io.qbeast.spark.internal
 import io.qbeast.core.model.QTableID
 import io.qbeast.spark.index.ColumnsToIndex
 import org.apache.spark.qbeast.config.DEFAULT_CUBE_SIZE
-import org.apache.spark.sql.AnalysisExceptionFactory
+import org.apache.spark.sql.{AnalysisExceptionFactory, DataFrame, SparkSession}
 
 /**
  * Container for Qbeast options.
  * @param columnsToIndex value of columnsToIndex option
  * @param cubeSize value of cubeSize option
  */
-case class QbeastOptions(columnsToIndex: Seq[String], cubeSize: Int)
+case class QbeastOptions(columnsToIndex: Seq[String], cubeSize: Int, stats: DataFrame)
 
 /**
  * Options available when trying to write in qbeast format
  */
+
 object QbeastOptions {
   val COLUMNS_TO_INDEX = "columnsToIndex"
   val CUBE_SIZE = "cubeSize"
   val PATH = "path"
+  val STATS = "stats"
 
+  /**
+   * Gets the columns to index from the options
+   * @param options the options passed on the dataframe
+   * @return
+   */
   private def getColumnsToIndex(options: Map[String, String]): Seq[String] = {
     val encodedColumnsToIndex = options.getOrElse(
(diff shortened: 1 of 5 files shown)


osopardo1:
I'm adding you @Adricu8 and @eavilaes as reviewers for trying out the feature and understanding if there's any bugs :) Thanks!

src/main/scala/io/qbeast/spark/internal/QbeastOptions.scala
```diff
@@ -33,13 +40,38 @@ object QbeastOptions {
...
       case Some(value) => value.toInt
       case None => DEFAULT_CUBE_SIZE
     }
   }
 
+  /**
+   * Get the column stats from the options
+   * This stats should be in a JSON formatted string
+   * with the following schema
+   * {min(a):value, max(a):value, min(b):value...}
```
Adricu8, line 60: IS this correct? In the tests the json is passed as `"min_columnName":10,"max_columnName":20`

src/main/scala/io/qbeast/spark/internal/QbeastOptions.scala
```diff
@@ -33,13 +40,38 @@ object QbeastOptions {
...
+   * @param options the options passed on the dataframe
+   * @return
+   */
+  private def getStats(options: Map[String, String]): DataFrame = {
+    val spark = SparkSession.active
+
+    options.get(STATS) match {
+      case Some(value) =>
+        import spark.implicits._
+        spark.read.json(Seq(value).toDS)
```
Adricu8, line 70: Does it make sense to use spark here or is it possible to parse the options without it? (just thinking)

Adricu8 (approved):
Hi, @osopardo1 thanks for the fast implementation! Changes look great :) 
I left a few comments but overall look good!

src/main/scala/io/qbeast/spark/internal/QbeastOptions.scala
osopardo1: You are right! Thank you, I will change it for the correct format.

src/main/scala/io/qbeast/spark/internal/QbeastOptions.scala
osopardo1: Sure, probably we could think of a better approach that only uses `JSON` classes 🤔 . This is a workaround for now, that is not really "dangerous" or problematic, because those options that we are trying to parse are Spark DataFrame options (so they imply an existent `SparkSession` that we can use)

Issue #153, opened 2023-02-06
Better query management [Spark]


osopardo1:
For doing queries against Qbeast Format, we filter pushdown all the information into `OTreeIndex`.  In this class, we create a `QuerySpec`, which gives the context of the query (percentage of sampling and filters) and a `QueryExecutor`, which is in charge of filtering the files needed to satisfy the predicates in the spec. 

In the `QueryExecutor`, we pass by all the revision and within each one we build a `IndexStatus`. This IndexStatus gathers information of the files and re-constructs the status of the index at a certain point in time. To do so, it has to call the `collect()` operation over the DeltaLog, which implies memory and computation resources which can drawback the reading performance. 

To improve the existing implementation, we can think of different approaches:
- **Remove the qbeast filtering over pushdown predicates**. Thanks to the stats collection of min and max over the files, Data Skipping strategy already does a great job. 
- **Try to build the sampling algorithm over DataFrame**. This would allow Qbeast-Spark to improve the access and processing of metadata, taking advantage of Spark distributed computing. We would need to build predicates that filters the Delta Log searching for those cubes that matches the weight filters.

osopardo1:
Another possible improvement to use spark benefits it's to keep a **cached df with Qbeast transformations** over the Delta Log to make it more accessible. We need to explore that possibility, still just an idea. 

(Similar to how Delta handles the snapshot files DataFrame).

osopardo1:
Close it because results did not show major improvement and for implement option 2 we will need more research on Spark DataFrame.

PR #161, opened 2023-02-08
Apply Delta data skipping when needed

osopardo1:
## Description

Partially fixes issue #153 

## Type of change

This is a small change in order to improve the way we are processing the queries.

Even though we collect per-file statistics, we do not use them for filtering log files. (We use our own Qbeast algorithm for processing the cubes). Due to the improvement in data skipping techniques from the formats, this part can be avoided when the query does not apply `Sampling`.

## Checklist:

- [x] Add tests.

This is not tested properly yet.

### DIFF
--- a/core/src/test/scala/io/qbeast/core/model/QuerySpaceFromToTest.scala
+++ b/core/src/test/scala/io/qbeast/core/model/QuerySpaceFromToTest.scala
@@ -29,6 +29,18 @@ class QuerySpaceFromToTest extends AnyFlatSpec with Matchers {
     querySpaceFromTo.intersectsWith(cube) shouldBe true
   }
 
+  it should "intersect correctly with non-existing coordinates" in {
+    val cube = CubeId.root(2)
+    val transformation = Seq(
+      LinearTransformation(1, 2, IntegerDataType),
+      LinearTransformation(1, 2, IntegerDataType))
+    val from = toSpaceCoordinates(Seq(None, Some(0)), transformation)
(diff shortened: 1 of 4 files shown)


osopardo1:
TODO: Need to filter out the Replicated files. If we do not filter those data out, we are going to read duplicated information.

osopardo1:
> TODO: Need to filter out the Replicated files. If we do not filter those data out, we are going to read duplicated information.

Done!

osopardo1:
This one needs to be tested to in order to publish new version 0.3.2. @victorcuevasv could you take care of it?

victorcuevasv:
Initial results running the 99-query TPC-DS benchmark at the 100 GB scale factor were generated for the PR161 and for the main branch latest commit (a425a22). The results can be found in:

https://github.com/Qbeast-io/qbeast-benchmarking/blob/master/writing-benchmark-scala/results-analysis-notebook/PR161_TPCDS100GB/TPCDS100GB.ipynb

for total, mean, and geomean execution time of the power tests; and in

https://github.com/Qbeast-io/qbeast-benchmarking/blob/master/writing-benchmark-scala/results-analysis-notebook/PR161_TPCDS100GB/TPCDS100GBqueries.ipynb

for individual queries.

The results show essentially the same execution times. Further analysis is still needed to determine if there are changes in the data scanned and tuples processed. This will be done later.

osopardo1:
Since the performance it's almost the same and the issue had lost relevance, I will close this. Hopefully we can come back to it after closing other priorities.

PR #167, opened 2023-02-16
Spark 3.3.0 delta 2.1.0

### LINKED ISSUE #166: Update to Delta version 2.x [Delta]
Delta release a couple months ago newest stable version 2.x
Since Qbeast is an extension of Delta Lake, we should maintain the compatibility versions aligned to avoid maintaining legacy code and keep the features up-to-date. 

This issue is to test if v2.x is causing any errors and make a list of changes/possible problems that we need to tackle.

cugni:
## Description
The idea of this PR is to update Qbeast-spark to the latest versions of Apache Spark and Delta Lake. More precisely, we'll try to align our builds to the releases of the Databricks' runtime and EMR, so that we can target the latest stable releases. In particular, for this PR, I'm targeting the Databricks' runtime [11.3 LTS](https://docs.databricks.com/release-notes/runtime/11.3.html) and EMR [6.9.0](https://docs.aws.amazon.com/emr/latest/ReleaseGuide/emr-release-app-versions-6.x.html). 
This would require updating our dependencies to: 

- Apache Spark 3.3.0 
- Delta Lake 2.1.0
- Hadoop 3.3.4 (Databricks) or 3.3.3 (EMR)

This PR will fix #166. 

As part of porting the code to the latest versions and making them work, we should also investigate how we can take advantage of the new features in both Spark and Delta. At first glance, it seems that we should investigate: 
1. The new improvements in the [DataSourceV2 API](https://spark.apache.org/releases/spark-release-3-3-0.html#data-source-v2-api]), in particular:
    1. DS V2 Sample Push Down ([SPARK-37038](https://issues.apache.org/jira/browse/SPARK-37038))
    2. DS V2 LIMIT push down ([SPARK-37020](https://issues.apache.org/jira/browse/SPARK-37020))
2. Check what new features in  (Delta 2.1.0)[https://github.com/delta-io/delta/releases/tag/v2.1.0] and (Delta 2.0.0)[https://github.com/delta-io/delta/releases/tag/v2.0.0] can be ported or used in Qbeast. 

## Checklist:

Here is a quick list of things we should get done before merging the PR. We can, however, split this in 2, one adding support to the new version of Spark and Delta, and another that takes benefit of the new functionalities. 

- [x] Make it compile.
- [x] Pass all tests
- [x] Check for performance regressions using  the TPC-DS benchmark. 

Future PR's would contain:
- Add support to the new functionalities of the DataSuorceV2
- Study additional features from Delta 2.1.0 and implement them. 
- Test with TPC-DS and compare results with the previous version.


project/build.properties
```diff
@@ -1 +1 @@
-sbt.version = 1.5.8
\ No newline at end of file
+sbt.version = 1.6.2
```
osopardo1, line 1: why do you change sbt version?

src/main/scala/io/qbeast/spark/internal/sources/catalog/QbeastCatalog.scala
```diff
@@ -254,4 +255,10 @@ class QbeastCatalog[T <: TableCatalog with SupportsNamespaces]
     } else throw new IllegalArgumentException("Invalid session catalog: " + delegate)
   }
 
+  override def listFunctions(namespace: Array[String]): Array[Identifier] =
+    getDelegatedCatalog().listFunctions(namespace)
```
osopardo1, line 259: I think here is getSessionCatalog() instead of getDelegatedCatalog().

project/build.properties
osopardo1: [info] [launcher] getting org.scala-sbt sbt 1.6.2  (this may take some time)...
[info] [launcher] getting Scala 2.12.15 (for sbt)...

Maybe instead of using scala 2.12.14 we can put scala 2.12.15 and it wouold solve the compilation errors?

cugni:
Now that the tests work, we have to check for any performance regression with the TPC-DS. @victorcuevasv, can you take care of it?

victorcuevasv:
Initially, I compared the earlier results we have of the main branch with commit 08fa349 with this pull request, on the appends benchmark with extended read tests.

The results of the appends benchmark are here:
https://github.com/Qbeast-io/qbeast-benchmarking/blob/master/writing-benchmark-scala/results-analysis-notebook/PR167/AnalyticsPlotsDomainDrivenDP.ipynb

and the extended read tests of the appends benchmark here:
https://github.com/Qbeast-io/qbeast-benchmarking/blob/master/writing-benchmark-scala/results-analysis-notebook/PR167/ReadTestsDomainDrivenDP.ipynb

The writing time is essentially the same, whereas the read time is slightly lower for PR 167, but mainly  due to query 9, which is slower in the earlier results.

I will run the benchmark again with the most up to date version of the main branch to see if the difference on query 9 persists.

cugni:
Great job, @victorcuevasv! 

If I understand correctly, writing is OK, as it goes at the same speed, if not even faster. 

On the other hand, it seems the new PR is slower than the master in all queries but the 9, especially if we consider the single worse query time. 
Maybe we should create a baseline benchmarking Spark 3.2.2 using Delta 1.2.0 (the combination used in the master) vs. Spark 3.3.0 or Delta Lake 2.1.0 as in this PR. We can test only the reading with Delta. 
In this way, we can understand better what caused the performance change: the new Spark/Delta or something in Qbeast. 

What do you think, @victorcuevasv? Is the different significative enough?

victorcuevasv:
I have updated the notebooks with the results from the main branch latest commit (a425a22).

Regarding the writing results presented in:

https://github.com/Qbeast-io/qbeast-benchmarking/blob/master/writing-benchmark-scala/results-analysis-notebook/PR167/AnalyticsPlotsDomainDrivenDP.ipynb

PR167 shows a small but significant improvement over the current main (around 10%). I executed the main branch twice yielding similar results.

It is important to remark that the read tests in this notebook are not sufficiently reliable under small differences. This is because the cluster state is affected by the append operations. Furthermore, each query is executed only once after each data insertion. Since the amount of data changes with the data insertions, the query instances are not comparable among each other. For this reason, 4 additional read tests on a new cluster are performed over the final generated data. These results are in the following notebook:

https://github.com/Qbeast-io/qbeast-benchmarking/blob/master/writing-benchmark-scala/results-analysis-notebook/PR167/ReadTestsDomainDrivenDP.ipynb

The large differences for query 9 arise because Spark version 3.3 used by PR167 generates a more optimal plan. For this reason, query 9 is excluded from the final sum of the read tests times. Excluding query 9, the main branch is faster than PR167 in the final time sum.

However, this occurs mainly because of a slow query instance in read test 4 and two slow query instances in read test 5. The execution times of read tests 6 and 7 are very similar. In addition, the best execution times for each query, as well as the average of the 3 best instances for each query are close. Therefore, there does not appear to be a significant slowdown. A new execution of PR167 could help to clarify if there is. According to the cube statistics the generated file structure is identical for both configurations.

Later I will apply the proposed baseline to these new results.

osopardo1:
Currently investigating about how to integrate the Sample Pushdown (DS V2 Sample Push Down ([SPARK-37038](https://issues.apache.org/jira/browse/SPARK-37038))) in our V2 Reader 🤔

src/main/scala/io/qbeast/spark/delta/OTreeIndex.scala
```diff
@@ -109,3 +109,19 @@ object OTreeIndex {
   }
 
 }
+
+case class EmptyIndex() extends FileIndex {
```
alexeiakimov, line 113: Looks like a singleton, probably it make sense to implement it as object.

src/main/scala/io/qbeast/spark/delta/OTreeIndex.scala
```diff
@@ -109,3 +109,19 @@ object OTreeIndex {
   }
 
 }
+
```
alexeiakimov, line 112: Scaladoc is missing.

alexeiakimov (changes requested):

src/main/scala/io/qbeast/spark/delta/OTreeIndex.scala
osopardo1: Done in 0028ee4

src/main/scala/io/qbeast/spark/delta/OTreeIndex.scala
osopardo1: Done in 0028ee4

cugni:
Two questions: 
1. Do we still need to support the DataSource API V1? If not, which Spark version would we not be able to support?
2. As long as adding the support to the Pushdown will most likely require significant changes, maybe it is best to split this PR in two. Now we can merge with the essential support to Spark 3.3.0, and then we can add more. What do you think, @osopardo1 @alexeiakimov?

osopardo1:
Yes, we were discussing this on #175 . I'm going to put this PR as ready for review now. Thanks.

alexeiakimov (approved):

PR #174, opened 2023-03-15
Revision compaction command

### LINKED ISSUE #171: Implement Staging Area + Level Compaction Strategy
The implementation of the append strategy may depend on the indexed column and the way we ingest the data:

- **Incremental**. Small appends characterise this type of writing. Those small appends usually contain a monotonic column, which increases its value quickly over time. In the vanilla version, this would cause the creation of a large number of revisions with mutually exclusive domains.
- **Batch**. In this case, data is written in a huge batch, and the chunks already contain an extensive range of values from each dimension.

We are trying to find the best way of appending data incrementally without creating many small files or deteriorating the index. 

A possible solution combines a write implementation and a compaction implementation. 

### Write: Staging Area + New Revision

The main idea is to **write files in a staging folder without indexing,** and when reaching a predefined number of insertions, we can index the data. 
Untitled (3)

The data saved in the staging area would be listed in the Delta Log but **not indexed with Qbeast** and not efficiently organised (neither skipped during readings). This would allow the system to perform **quick insertions** and produce a **less-changing tree structure** for the qbeast files.

For this implementation, it makes sense to implement a **single `Revision` for each write operation.** **In the current writing process, the trigger of a new `Revision` it's done if the space changes. It's not mandatory that each write creates a new one, but _it would be likely if the upcoming data is increasing over time._** 

### Compaction: Level Compaction

Based on [LSM compaction mechanism](https://www.alibabacloud.com/blog/an-in-depth-discussion-on-the-lsm-compaction-mechanism_596780) used in Cassandra database 

**Each level is an ordered run that consists of multiple `Revisions`**. The size of each level is maintained at T times that of the previous level.

Untitled (2)

This method combines multiple runs of the level to one, **reducing the read amplification and space amplification**. 

This policy on the Level Compaction **needs to be defined and described better** so the user can tune the parameters (e.g size of the levels) and understand the behaviour of the operation. 

### To-Do List

- [ ] Describe the policy for Level Compaction.
- [ ] Implement both the writing and compaction strategies on a Pull Request.
- [ ] Test this implementation against the current. Run TPC-DS benchmark with an small appends workload and analyze the query results and the final index structure.

Jiaweihu08:
## Description
Fixes #171 
This PR introduces the Level-Compaction for Qbeast Revisions, attempting to reduce the number of revisions created by appending data with doundaries not included in the latest revision.

More descriptions pending.

## Type of change
New Feature

## Checklist:
- [x] Add tests.

Pending.

**Test Configuration**:
* Spark Version:
* Hadoop Version:
* Cluster or local?


Jiaweihu08:
PR is temporarily closed until we have a clearer view of this feature's working scenarios and its technical requirements.

PR #177, opened 2023-03-17
Fix column stats

### LINKED ISSUE #176: Writing with column stats create the first revision with id = 0
## What went wrong?
When using the columnStats option to give a larger indexing column boundaries we end up creating a Revision with RevisionID = 0.

## How to reproduce?
Indexing the ecommerce dataset with the following configurations:
```
(df
  .write
  .format("qbeast")
  .mode("overwrite")
  .option("cubeSize", "10000")
  .option("columnsToIndex", "user_id,price")
  .option("columnStats", """{"price_min":-1.0,"price_max":999999.0, "user_id_min":-1, "user_id_max": 999999999 }""")
  .save("/tmp/test1/")
)
```
### 2. Branch and commit id:
`main`, [e874051](https://github.com/Qbeast-io/qbeast-spark/commit/e8740512acca21bbe8de90502df089c66310e6f5)

### 3. Spark version:
`3.2.2`

### 4. Hadoop version:
`3.3.1`

Local spark shell

Jiaweihu08:
## Description

Fixes #176

#176 is caused by `calculateRevisionChanges` not being able to detect a change of revision because the revision used is already updated with valid columnStats.

The solution provided here is to directly update the `RevisonID` when dimension boundaries are provided for all indexing columns before any further processing. So, if there's no `SpaceChanges` detected, the revision we use would still have the proper value: `latestRevision.revisionID + 1`.

An unintended consequence of this approach would be incrementing the `RevisionID` twice if column boundaries are provided for all indexing columns but some or all are smaller than that from the actual data - once during revision creation, and again during `spaceChanges.createNewRevision`. This 'issue' has no affect on the data or the integrity of the index at all, though it'd be nice to remove it.

When the `columnStats` provided are valid, that is, larger than the min/max of the actual data, we face with an inconsistency for the value of `isNewRevision` from `BroadcastedTableChanges`; it'd be set to `false` when it should be `true`. Currently this leads to no harm by the way `isNewRevision` is defined in `updateQbeastMetadata`, but it is something to take into account.
 
## Type of change

**Bug fix**
## Checklist:
- [x] Add tests.

- Test RevisionID = 0 is not created when valid columnStats are used during the first write.

### DIFF
--- a/src/main/scala/io/qbeast/spark/index/OTreeDataAnalyzer.scala
+++ b/src/main/scala/io/qbeast/spark/index/OTreeDataAnalyzer.scala
@@ -61,6 +61,10 @@ object DoublePassOTreeDataAnalyzer extends OTreeDataAnalyzer with Serializable {
   private[index] def calculateRevisionChanges(
       row: Row,
       revision: Revision): Option[RevisionChange] = {
+    // TODO: When all indexing columns are provided with a boundary, a new revision is
+    //  created directly. If the actual data boundaries are not contained by those
+    //  values, the RevisionID would then increase again by 1, leaving a discontinued
+    //  sequence of RevisionIDs in the metadata.
 
     val newTransformation =
       revision.columnTransformers.map(_.makeTransformation(colName => row.getAs[Object](colName)))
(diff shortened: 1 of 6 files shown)


osopardo1:
LGTM! Once checks are passed, ready to merge.

Issue #175, opened 2023-03-16
Support for Spark 3.3.x Sampling Pushdown and Limit Pushdown


osopardo1:
Related to #166 .

Qbeast-Spark should be compatible with latest versions of Delta Lake and Apache Spark, to benefit from any new features and major upgrades.
The change to Delta version 2.1.0 and Spark 3.3.0, reveal a set of interesting Pushdown operation that could be empowered with the Qbeast Metadata.
- DS V2 Sample Push Down ([SPARK-37038](https://issues.apache.org/jira/browse/SPARK-37038))
- DS V2 LIMIT push down ([SPARK-37020](https://issues.apache.org/jira/browse/SPARK-37020))

We should:

- [ ] Integrate those pushdown properties into `QbeastTableImpl`.
- [ ] Eliminate the Sample optimisation rule from `QbeastSparkSessionExtension`.
- [ ] The deletion of Sample Optimisation would also affect #68 . This requieres some more insights.*
- [ ] Plan the Limit Pushdown approach.
- [ ] Merge with PR #167 

### Thoughts on #68 

Ideally, **we would no longer need to compute the hash for the indexed columns** in order to filter the records in memory. But then, we need to ensure the sample returned from `OTreeIndex` (or from any other class that is involved, such as `ParquetFileFormat`) is correct.
Since for each file we know the minWeight and maxWeight, we can determine how many rows we can read from it. The only thing we need to find out is where those records are filtered. Here's a detailed blog that explains how Spark process Parquet files: https://animeshtrivedi.github.io/spark-parquet-reading

osopardo1:
In DatasourceV2 there's also the possibility to **build your own scan of the table**, with more options than the Datasource V1 (which we are currently using). 

Maybe it's worth to explore the DV2 API.

alexeiakimov:
IMHO we need to explore the Datasource V2 API, possibly we will end to drop the V1. To support both can be too much conditional logic.

osopardo1:
Yes, I agree. Do you think this can be done in the same PR #167 or it is best to do workaround first for Sampling and Limit Pushdown and migrate everything to V2 in a separate issue?

alexeiakimov:
[...]

PR #178, opened 2023-03-24
#175 initial implementation of Read API V2

alexeiakimov:
## Description

This PR provides implementation for the read part of the Spark DataSource API V2. Once merged it should close the issue #175.

At the moment the following items are pending:
1. Filter push down which should be ported from DataSource V1 implementation
2. Sampling push down
3. Limit push down
4. Tests for QbeastDataSource and dependent classes.

This PR starts as a draft to make it possible to review it ASAP, the pending items mentioned above
 will be added to the same PR later as soon they are ready.

## Type of change

This change is internal process of migrating the Qbeast code to the Spark DataSource API V2. After the changes are applied the library should be backward compatible modulo it requires version of Spark 3.+.

## Checklist:

1. Implement the Read API V2.
2. Port filter push down from the DataSource V1 implementation.
3. Implement sampling push down.
4. Implement limit push down.
5. Fix QbeastDataSource tests.
6. Create tests for Read API V2 implementation.
7. Probably remove Read API V1 implementation.

Since the change is internal migration process it should be enough to use existing high level integration tests and to add unit tests for the new code.


src/main/scala/io/qbeast/spark/internal/sources/QbeastDataSource.scala
```diff
@@ -47,39 +47,35 @@ class QbeastDataSource private[sources] (private val tableFactory: IndexedTableF
...
+    QbeastContext.metadataManager.loadCurrentSchema(tableId)
+  }
 
   // Used to get the table of an existing one
   override def getTable(
       schema: StructType,
       partitioning: Array[Transform],
       properties: util.Map[String, String]): Table = {
-    val tableId = QbeastOptions.loadTableIDFromParameters(properties.asScala.toMap)
+    val tableProperties = new util.HashMap[String, String](properties).asScala
```
cugni, line 61: why is it necessary to do all these transformations on this map? From java Map, to HashMap, to Scala Map, and then a few line down other toMap? Can't it just be converted once?

src/main/scala/io/qbeast/spark/internal/sources/v2/QbeastBatch.scala
```diff
@@ -0,0 +1,27 @@
+/*
+ * Copyright 2021 Qbeast Analytics, S.L.
```
cugni, line 2: I guess we should update the copyright to 2023 😅

src/main/scala/io/qbeast/spark/internal/sources/v2/QbeastPartitionReader.scala
```diff
@@ -0,0 +1,27 @@
...
+ */
+package io.qbeast.spark.internal.sources.v2
+
+import org.apache.spark.sql.catalyst.InternalRow
+import org.apache.spark.sql.connector.read.PartitionReader
+
+/**
+ * Implementation of [[PartitionReader]].
+ */
+class QbeastPartitionReader private[v2] (private val iterator: Iterator[InternalRow])
```
cugni, line 12: This partition reader is only Rows, not for Columnar Reading (which is the one that has better performance). We have to implement both a reader that PartitionReader[InternalRow] and one that does PartitionReader[ColumnBatch]. So, it would probably be best to call this class QbeastPartitionRowReader, or something similar

src/main/scala/io/qbeast/spark/internal/sources/v2/QbeastPartitionReaderFactory.scala
```diff
@@ -0,0 +1,59 @@
...
+import org.apache.spark.sql.SparkSession
+import org.apache.spark.sql.execution.PartitionedFileUtil
+import org.apache.spark.sql.execution.datasources.{FileIndex, PartitionDirectory, PartitionedFile}
+import org.apache.spark.sql.execution.datasources.parquet.ParquetFileFormat
+import org.apache.spark.sql.types.StructType
+
+/**
+ * Implementation of [[PartitionReaderFactory]].
+ */
+class QbeastPartitionReaderFactory private[v2] (
```
cugni, line 22: I think this class should also override the methods for creating a columnar reader (PartitionReader<ColumnarBatch> createColumnarReader(InputPartition partition)). 
In the PartitionReaderFactory interface, these methods return exceptions by default, but we should support columnar readying.

cugni (commented):
This is a work in progress, but we should consider already supporting columnar reading. 
Is it really necessary/desirable to create single files for each class, even if they are very small? I know it is a requirement in Java but in Scala? I can see the benefits of both approaches; I'm wondering what our is 😅

src/main/scala/io/qbeast/spark/internal/sources/QbeastDataSource.scala
alexeiakimov: Unfortunately yes. The `util.Map` passed as a parameter turned out to be immutable (`java.util.collections.Collections.unmodifiableMap()`), and `asScala` does not copy the values it is just a wrapper, so without making a copy, the modification below would fail.

src/main/scala/io/qbeast/spark/internal/sources/v2/QbeastBatch.scala
alexeiakimov: I will fix it, thank you.

alexeiakimov:
> This is a work in progress, but we should consider already supporting columnar reading. Is it really necessary/desirable to create single files for each class, even if they are very small? I know it is a requirement in Java but in Scala? I can see the benefits of both approaches; I'm wondering what our is sweat_smile

IMHO the project is mainly in Scala, so I would prefer to use Scala as much as possible to avoid mixing the languages. Regarding the code style I tried to follow the "official" recommendation https://docs.scala-lang.org/style/files.html. Honestly I think that Java and Scala use too fine grained approach, so if you believe that the official strategy is not optimal, please let me know, it is easy to place those classes in a single file. Thank you.

src/main/scala/io/qbeast/spark/internal/sources/v2/QbeastBatch.scala
```diff
@@ -0,0 +1,27 @@
...
+package io.qbeast.spark.internal.sources.v2
+
+import org.apache.spark.sql.connector.read.{Batch, InputPartition, PartitionReaderFactory}
+import org.apache.spark.sql.execution.datasources.FileIndex
+import org.apache.spark.sql.types.StructType
+
+/**
+ * Implementation of org.apache.spark.sql.connector.read.Batch.
+ */
+class QbeastBatch private[v2] (
```
osopardo1, line 13: Do we need a particular implementation of `Batch`? Could we delegate some of the operations to Parquet built-in classes?

Just to avoid re-implementing a lot of the features, and to be compatible with other possible optimisations for Parquet files.

osopardo1:
Just one more comment: I am not sure if we are using the V2 implementation right now. 

To make a fallback to V1 for reading, we made a rule on the Spark Plan that removes the calls to `DataSourceV2Relation` and replaces it to a `LogicalRelation`. Could you remove this rule and run the test? 

It is placed here: https://github.com/Qbeast-io/qbeast-spark/blob/cb19cd40e229be9d9783d18de09822c10712ffbd/src/main/scala/io/qbeast/spark/internal/rules/QbeastAnalysis.scala#L17
And initialised on the SessionExtension: https://github.com/Qbeast-io/qbeast-spark/blob/cb19cd40e229be9d9783d18de09822c10712ffbd/src/main/scala/io/qbeast/spark/internal/QbeastSparkSessionExtension.scala#LL20C4-L20C4

src/main/scala/io/qbeast/spark/internal/sources/v2/QbeastScanBuilder.scala
```diff
@@ -0,0 +1,23 @@
...
+package io.qbeast.spark.internal.sources.v2
+
+import org.apache.spark.sql.connector.read.{Scan, ScanBuilder}
+import org.apache.spark.sql.execution.datasources.FileIndex
+import org.apache.spark.sql.types.StructType
+
+/**
+ * Implementation of org.apache.spark.sql.connector.read.ScanBuilder.
+ */
+class QbeastScanBuilder private[v2] (
```
osopardo1, line 13: This scan builder should support at least `SupportsPushDownFilters` and `SupportsPushDownRequiredColumns`.

src/main/scala/io/qbeast/spark/internal/sources/v2/QbeastBatch.scala
osopardo1: I know we can do it, but I am not sure if it's the best option for the Qbeast reading. 

I am thinking about the special case of Sampling and Limit: since we know what are the min and max weight values of a file, we could read just a part of it.  I don't know if Parquet classes had parameters to infer how many records should be read or if we need to re-implement some of the methods. We can investigate. 

What do you think @cugni ?

src/main/scala/io/qbeast/spark/internal/sources/v2/QbeastPartitionReader.scala
osopardo1: There is trait called `PartitionReaderFactory` that has both methods: createColumnarReader and `createReader` (and `supportColumnarReads` to verify it can read in columnar batches). We can extend that.

src/main/scala/io/qbeast/spark/internal/sources/v2/QbeastPartitionReaderFactory.scala
osopardo1: Agree. There's also the method `supportColumnarReads` which tells the Plan that the Format supports columnar batch.

osopardo1:
Do you think we can close this PR for the moment? @cugni

osopardo1:
Hello! Is this PR relevant at the moment? Should we close it and pause until other priorities are resolved? @alexeiakimov

alexeiakimov:
I have no idea when we continue working on this, probably the codebase will be different, so let's close it.

osopardo1:
Closing until the codebase is stable to continue the work.

PR #180, opened 2023-04-06
PR179: add logging of filtered files

victorcuevasv:
## Description

- You may include "Fixes #(issue)": A reference to a related [issue](https://github.com/Qbeast-io/qbeast-spark/issues).

- A clear, concise description of the changes proposed in the pull request.

- (Optional) @mentions of the person or team responsible for reviewing the proposed changes.

## Type of change

Describe the change you're making: how it affects the API, user experience...

- Is it a **Bug fix**? (non-breaking change which fixes an [issue](https://github.com/Qbeast-io/qbeast-spark/issues)).
- Have you developed a **New feature**? (non-breaking change, which adds new functionality).
- Are you introducing a **Breaking change**? (fix or feature that would cause existing functionality not to work or do it differently than previously).
- **Documentation** update (updates regarding the [Users docs](https://github.com/Qbeast-io/qbeast-spark/tree/main/docs) or the [Developer docs](https://docs.qbeast.io/), or similar).

## Checklist:

- [ ] Add tests.

- Test A: _(e.g., I have created the class FooTest.scala...)_
- Test B: _(e.g., Running some benchmarks...)_

**Test Configuration**:
* Spark Version:
* Hadoop Version:
* Cluster or local?

### DIFF
--- a/src/main/scala/io/qbeast/spark/delta/OTreeIndex.scala
+++ b/src/main/scala/io/qbeast/spark/delta/OTreeIndex.scala
@@ -13,6 +13,8 @@ import org.apache.spark.sql.delta.actions.AddFile
 import org.apache.spark.sql.delta.files.TahoeLogFileIndex
 import org.apache.spark.sql.execution.datasources.{FileIndex, PartitionDirectory}
 import org.apache.spark.sql.types.StructType
+import org.apache.spark.internal.Logging
+import org.apache.spark.sql.execution.SQLExecution
 
 import java.net.URI
 
@@ -20,8 +22,11 @@ import java.net.URI
  * FileIndex to prune files
  *
  * @param index the Tahoe log file index
+ * @param spark spark session
  */
-case class OTreeIndex(index: TahoeLogFileIndex) extends FileIndex {
+case class OTreeIndex(index: TahoeLogFileIndex, spark: SparkSession)
+    extends FileIndex
+    with Logging {
 
   /**
    * Snapshot to analyze
@@ -69,7 +74,6 @@ case class OTreeIndex(index: TahoeLogFileIndex) extends FileIndex {
   override def listFiles(
       partitionFilters: Seq[Expression],
       dataFilters: Seq[Expression]): Seq[PartitionDirectory] = {
-
     val qbeastFileStats = matchingBlocks(partitionFilters, dataFilters).map { qbeastBlock =>
       new FileStatus(
         /* length */ qbeastBlock.size,
@@ -81,7 +85,18 @@ case class OTreeIndex(index: TahoeLogFileIndex) extends FileIndex {
     }.toArray
     val stagingStats = stagingFiles
     val fileStats = qbeastFileStats ++ stagingStats
-
+    val execId = spark.sparkContext.getLocalProperty(SQLExecution.EXECUTION_ID_KEY)
+    partitionFilters.foreach(f => {
+      logInfo(s"OTreeIndex partition filter (exec id ${execId}): ${f.toString}")
+    })
+    dataFilters.foreach(f => {
+      logInfo(s"OTreeIndex data filter (exec id ${execId}): ${f.toString}")
+    })
+    val allFilesCount = snapshot.allFiles.count
+    val nFiltered = allFilesCount - fileStats.length
+    val filteredPct = (((nFiltered * 1.0) / allFilesCount) * 100.0).round
+    val filteredMsg = s"${nFiltered} of ${allFilesCount} (${filteredPct}%)"
+    logInfo(s"OTreeIndex filtered files (exec id ${execId}): ${filteredMsg}")
     Seq(PartitionDirectory(new GenericInternalRow(Array.empty[Any]), fileStats))
 
   }
@@ -108,7 +123,7 @@ object OTreeIndex {
   def apply(spark: SparkSession, path: Path): OTreeIndex = {
     val deltaLog = DeltaLog.forTable(spark, path)
     val tahoe = TahoeLogFileIndex(spark, deltaLog, path, deltaLog.snapshot, Seq.empty, false)
-    OTreeIndex(tahoe)
+    OTreeIndex(tahoe, spark)
   }
 
 }


src/main/scala/io/qbeast/spark/delta/OTreeIndex.scala
```diff
@@ -81,7 +85,18 @@ case class OTreeIndex(index: TahoeLogFileIndex) extends FileIndex {
...
+      logInfo(s"OTreeIndex partition filter (exec id ${execId}): ${f.toString}")
+    })
+    dataFilters.foreach(f => {
+      logInfo(s"OTreeIndex data filter (exec id ${execId}): ${f.toString}")
+    })
+    val allFilesCount = snapshot.allFiles.count
+    val nFiltered = allFilesCount - fileStats.length
+    val filteredPct = (((nFiltered * 1.0) / allFilesCount) * 100.0).round
+    val filteredMsg = s"${nFiltered} of ${allFilesCount} (${filteredPct}%)"
+    logInfo(s"OTreeIndex filtered files (exec id ${execId}): ${filteredMsg}")
```
osopardo1, line 99: Instead of `OTreeIndex` could we print Qbeast?

src/main/scala/io/qbeast/spark/delta/OTreeIndex.scala
```diff
@@ -81,7 +85,18 @@ case class OTreeIndex(index: TahoeLogFileIndex) extends FileIndex {
     }.toArray
     val stagingStats = stagingFiles
     val fileStats = qbeastFileStats ++ stagingStats
-
+    val execId = spark.sparkContext.getLocalProperty(SQLExecution.EXECUTION_ID_KEY)
+    partitionFilters.foreach(f => {
```
osopardo1, line 89: I think it's best if we print the sequence directly in a single message.

src/main/scala/io/qbeast/spark/delta/OTreeIndex.scala
```diff
@@ -13,15 +13,20 @@ import org.apache.spark.sql.delta.actions.AddFile
...
 import java.net.URI
 
 /**
  * FileIndex to prune files
  *
  * @param index the Tahoe log file index
+ * @param spark spark session
  */
-case class OTreeIndex(index: TahoeLogFileIndex) extends FileIndex {
+case class OTreeIndex(index: TahoeLogFileIndex, spark: SparkSession)
```
osopardo1, line 27: If we add `spark` to the parameters, we need to change it in any test or place where OTreeIndex is called. Maybe we can use `SparkSession.active ` to retrieve the context and use logging.

src/main/scala/io/qbeast/spark/delta/OTreeIndex.scala
victorcuevasv: OK, done.

src/main/scala/io/qbeast/spark/delta/OTreeIndex.scala
victorcuevasv: Using SparkSession.active instead.

src/main/scala/io/qbeast/spark/delta/OTreeIndex.scala
victorcuevasv: Ok, modified to print the sequence as a single message.

src/main/scala/io/qbeast/spark/delta/OTreeIndex.scala
```diff
@@ -81,9 +83,18 @@ case class OTreeIndex(index: TahoeLogFileIndex) extends FileIndex {
     }.toArray
     val stagingStats = stagingFiles
     val fileStats = qbeastFileStats ++ stagingStats
-
+    val sc = SparkSession.active.sparkContext
```
cugni, line 86: Instead of getting the global spark session, which could give some issues if we have multiple spark session (e.g. in testing), you can use the spark session that is in the TahoeLogFileIndex
```scala
val sc = index.spark.sparkContext
```

src/main/scala/io/qbeast/spark/delta/OTreeIndex.scala
```diff
@@ -81,9 +83,18 @@ case class OTreeIndex(index: TahoeLogFileIndex) extends FileIndex {
...
+    val execId = sc.getLocalProperty(SQLExecution.EXECUTION_ID_KEY)
+    val pfStr = partitionFilters.map(f => f.toString).mkString(" ")
+    logInfo(s"OTreeIndex partition filters (exec id ${execId}): ${pfStr}")
+    val dfStr = dataFilters.map(f => f.toString).mkString(" ")
+    logInfo(s"OTreeIndex data filters (exec id ${execId}): ${dfStr}")
+    val allFilesCount = snapshot.allFiles.count
+    val nFiltered = allFilesCount - fileStats.length
+    val filteredPct = (((nFiltered * 1.0) / allFilesCount) * 100.0).round
```
cugni, line 94: I think it would be best to have percentage with at least 2 decimal digits (e.g. 0.23% or 53.23%). 

You can use the f instead of the s to format properly numbers. For instance:

```
val filteredPct = (nFiltered * 100.0) / allFilesCount
val filteredMsg = f"${nFiltered} of ${allFilesCount} (${filteredPct}%.2f%%)"
```

cugni (approved):
A few minor changes were suggested to reduce spaghetti code and better data printing. 

In general, it would be nice to add some tests, but I'm not sure how to structure them, and it might be over-killing for this case. 
It would be nice to have in the future a more structured way to measure what is happening within Qbeast.

osopardo1:
> A few minor changes were suggested to reduce spaghetti code and better data printing.
> 
> In general, it would be nice to add some tests, but I'm not sure how to structure them, and it might be over-killing for this case. It would be nice to have in the future a more structured way to measure what is happening within Qbeast.

I think we should at least test that something has been printed on the logs.

src/main/scala/io/qbeast/spark/delta/OTreeIndex.scala
victorcuevasv: Ok, changed to print two decimals.

osopardo1:
I think is ready to merge!

PR #181, opened 2023-04-12
Qbeast-Photon module

osopardo1:
## Description

In this PR, we are adding support for querying Qbeast Format in [Photon](https://cs.stanford.edu/~matei/papers/2022/sigmod_photon.pdf), the new Query Engine for Spark. 

## Type of change

It is a **new feature** that includes the following changes:

- New module named `qbeast-photon` that depends on `qbeast-core` and has it's own datasource implementation. This Datasource/Table Provider includes the necessary capabilities for **`Sample`, `Filter` and `Aggregation` Pushdown**. 
- Reorganisation of the code: we moved some parts of the Query Management into core module, to avoid having code duplication. After some analysis, we still have Spark parts that can be shared with `qbeast-spark` main project, maybe we can decouple Spark code to be able to reuse as much as possible. We can discuss it in this PR. 
- `README` and instructions are uploaded in the independent module.

## Checklist:

- [ ] Add tests.


osopardo1:
TODO:

- I commented some test on `qbeast-spark `module. Need to double check them.
- Add more test on the `qbeast-photon` module.
- Test the full solution on platforms.

osopardo1:
Morning!

Updates on the module:

- At first, we implement `OTreePhotonIndex` as an extension of `InMemoryFileIndex`. This class returns a list of `FileStatus` that contains the files to read. Unfortunately, this `FileStatus` is not compatible with internal Databricks class `SerializableFileStatus` on customised implementations (like ours), and an error appears on reading.
- To solve this problem, we can:
    1. initialise an `InMemoryFileIndex` with the filtered files already. This would be converted in a `PhotonScan` directly and can fully benefit from Photon vectorial reading. This solution is proved to be working on DB. 
    2. Use an extension of `PartitioningAwareFileIndex`. Not sure if this would work on DB clusters. Needs to be tested.
    
- The tests on `qbeast-spark` are modified accordingly.
- Some small test are added on `qbeast-photon`. 

Next steps:

- Run TPC-DS queries to verify compatibility. 
- Review this PR. Clean the code, improve the usability and readability.

osopardo1:
Few comments:

- It's better to separate this PR in two, so code review it's easier. One of them would contain only Photon module (with no refactoring) and the other would include the package changes.
- Need to include CI with Databricks for this part. We would discuss the best way to do so and come back with the changes.

osopardo1:
Closing this PR to separate the code in two. Sorry for the inconvenience.

PR #186, opened 2023-04-25
Support for OR and IN operator

### LINKED ISSUE #183: Support OR operator to filter dataframes
## What went wrong?
Having a dataset indexed by a string column it is able to push down a filter operator and take advantage of the index to read less data. Though, having a filter containing **OR** operators for the same column requires to read all data.

I am not sure this is a wanted feature in Qbeast but I leave it here in case you consider it useful. There is a workaround to do this where you make a single filter per value and finally union the results.
## How to reproduce?

Index a dataset with a string column.
Apply filter operator **OR** 

```scala
val strings = Seq("aaa", "bbb", "ccc", "ddd", "eee", "fff").toDF
strings.write.format("qbeast").option("columnsToIndex","value").save("/tmp/test")
val df_test = spark.read.format("qbeast").load("/tmp/test")
df_test.filter("value = 'aaa' or value = 'ccc'").show
```

### 3. Spark version:
3.2.1

### 4. Hadoop version:
3.2.1

local

osopardo1:
## Description

This PR fixes #183 . 

## Type of change

It includes additional functionalities on the `query` package that processes the OR query operators, creating several query spaces to iterate and union. 

The pipeline is the following:
1. SparkFilters (a sequence of `Expression`) are passed to the `OTreeIndex` from Spark Query Plan. 
2. We create a `QuerySpecBuilder`.
3. To filter the files, we load all the revisions from `QbeastSnapshot` and we build a `QuerySpec` for each of them. 
4. Now, we can have **several `QuerySpecs` for a single `Revision`**. Since the predicates contain OR's, our queries are divided into smaller filters that can be run independently. To create the QuerySpecs, we do:

    i. **Split Query Filters and Weight Filters**. Weight filters are those which indicates the size of the sample. Query Filters are those that involve any of the indexed columns. 
    ii. **Split the disjunctive and conjunctive predicates**. The predicates are parsed as a sequence of ANDs. The OR's are parsed into a single Expression. 
   iii. **Create a single `QuerySpec` with the conjunctive predicates** (AND). 
   iv. **For each disjunctive predicate (OR), create a different `QuerySpec`.** 

6. We run each query and we union the results.

For the IN predicate, we pre-process the filters in `QuerySpecBuilder` as follows:

1. Match the `IN` filter passed to the datasource.
2. We can manage IN as a sequence of OR filtering, each one retrieving one slice of the space. But this is not scalable when we have too many values in the set.
3. A solution is to **find a range `(min, max)`** and retrieve the blocks belonging to that `QuerySpaceFromTo`. When it comes to a **String indexed column, finding a range could be difficult because Strings are hashed** and, as far as I know, **`Murmur3Hash` does not respect ordering**. To solve that, we must:
    - Hash the values inside the IN set.
    - Select the `min` and `max` of those values.
    - Initialise the `QuerySpaceFromTo`. 

## Checklist:

- [x] Add tests.

This has been tested in a separate class `io.qbeast.spark.index.query.DisjunctiveQuerySpecTest `, but also on `io.qbeast.spark.utils.QbeastFilterPushdownTest`

**Test Configuration**:
* Spark Version: 3.3.0
* Hadoop Version: 3.3.4
* Cluster or local? Local

### DIFF
--- a/core/src/main/scala/io/qbeast/core/model/QuerySpace.scala
+++ b/core/src/main/scala/io/qbeast/core/model/QuerySpace.scala
@@ -55,7 +55,8 @@ case class EmptySpace() extends QuerySpace {
  * @param from inclusive starting range
  * @param to   exclusive ending query range
  */
-case class QuerySpaceFromTo(from: Seq[Option[Double]], to: Seq[Option[Double]]) extends QuerySpace {
+case class QuerySpaceFromTo(from: Seq[Option[Double]], to: Seq[Option[Double]])
+    extends QuerySpace {
 
   private def intersects(f: Double, t: Double, cube: CubeId, coordinate: Int): Boolean = {
     val cf = cube.from.coordinates(coordinate)
@@ -77,7 +78,7 @@ case class QuerySpaceFromTo(from: Seq[Option[Double]], to: Seq[Option[Double]]) 
 
   override def contains(other: QuerySpace): Boolean = {
     other match {
-      case q : QuerySpaceFromTo =>
+      case q: QuerySpaceFromTo =>
         q.from.zip(from).forall {
           case (Some(otherFrom), Some(f)) => otherFrom >= f
           case (Some(otherFrom), None) => otherFrom == 0.0
@@ -118,11 +119,22 @@ object QuerySpace {
     var isPointStringSearch = false
     var (isOverlappingSpace, isAllSpace) = (true, true)
 
+    // TODO This implementation may cause problems when a query
+    //  with (>=, <=) on HashTransformation is issued
+    //  without transforming the values first
     from.indices.foreach { i =>
       val (isOverlappingDim, isAllDim) = (from(i), to(i), transformations(i)) match {
-        case (Some(f), Some(t), _: HashTransformation) if f == t =>
-          isPointStringSearch = true
-          (true, true)
+        case (
+              Some(f),
+              Some(t),
+              _: HashTransformation
+            ) => // Accept Range Filters for HashTransformation only if they have min and max values
+          if (f == t) {
+            isPointStringSearch = true
+            (true, true)
+          } else {
+            (f <= t && f <= 1d && t >= 0d, f <= t && f <= 0d && t >= 1d)
+          }
         case (_, _, _: HashTransformation) | (None, None, _) =>
           (true, true)
         case (Some(f), Some(t), _) =>
@@ -138,7 +150,7 @@ object QuerySpace {
     }
 
     if (isAllSpace && !isPointStringSearch) AllSpace()
-    else if (isPointStringSearch || isOverlappingSpace) new QuerySpaceFromTo(from, to)
+    else if (isPointStringSearch || isOverlappingSpace) QuerySpaceFromTo(from, to)
     else EmptySpace()
   }
 
--- a/core/src/main/scala/io/qbeast/core/model/QuerySpace.scala
+++ b/core/src/main/scala/io/qbeast/core/model/QuerySpace.scala
@@ -11,9 +11,9 @@ import io.qbeast.core.transform.{HashTransformation, Transformation}
 trait QuerySpace {
 
   /**
-   * Checks if other query space is contained in the current
+   * Checks if this QuerySpace contains other QuerySpace
    * @param other the other query space
-   * @return true if they overlap
+   * @return true if this QuerySpace contains the other QuerySpace
    */
 
   def contains(other: QuerySpace): Boolean
@@ -44,7 +44,9 @@ case class EmptySpace() extends QuerySpace {
 
   override def intersectsWith(cube: CubeId): Boolean = false
 
-  override def contains(other: QuerySpace): Boolean = false
+  // The only case in which EmptySpace contains other QuerySpace
+  // is when other QuerySpace is an EmptySpace
+  override def contains(other: QuerySpace): Boolean = other.isInstanceOf[EmptySpace]
 }
 
 /**
--- a/core/src/main/scala/io/qbeast/core/model/QuerySpace.scala
+++ b/core/src/main/scala/io/qbeast/core/model/QuerySpace.scala
@@ -11,6 +11,14 @@ import io.qbeast.core.transform.{HashTransformation, Transformation}
 trait QuerySpace {
 
   /**
+   * Checks if other query space is contained in the current
+   * @param other the other query space
+   * @return true if they overlap
+   */
+
+  def contains(other: QuerySpace): Boolean
+
+  /**
    * Returns whether the space intersects with a given cube.
    *
    * @param cube the cube
@@ -25,6 +33,8 @@ trait QuerySpace {
 case class AllSpace() extends QuerySpace {
 
   override def intersectsWith(cube: CubeId): Boolean = true
+
+  override def contains(other: QuerySpace): Boolean = true
 }
 
 /**
@@ -33,6 +43,8 @@ case class AllSpace() extends QuerySpace {
 case class EmptySpace() extends QuerySpace {
 
   override def intersectsWith(cube: CubeId): Boolean = false
+
+  override def contains(other: QuerySpace): Boolean = false
 }
 
 /**
@@ -43,8 +55,7 @@ case class EmptySpace() extends QuerySpace {
  * @param from inclusive starting range
  * @param to   exclusive ending query range
  */
-class QuerySpaceFromTo(private val from: Seq[Option[Double]], private val to: Seq[Option[Double]])
-    extends QuerySpace {
+case class QuerySpaceFromTo(from: Seq[Option[Double]], to: Seq[Option[Double]]) extends QuerySpace {
 
   private def intersects(f: Double, t: Double, cube: CubeId, coordinate: Int): Boolean = {
     val cf = cube.from.coordinates(coordinate)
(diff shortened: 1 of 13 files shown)


Adricu8:
Does operator ISIN behave in the same way and is also pushed-down?
In theory, they should be treated in the same way under the hood.
We could add some test to verify it

osopardo1:
> Does operator [ISIN](https://spark.apache.org/docs/1.6.0/api/java/org/apache/spark/sql/Column.html#isin(scala.collection.Seq)) behave in the same way and is also pushed-down? In theory, they should be treated in the same way under the hood. We could add some test to verify it
> 
> Example:
> 
> ```sql
> SELECT *
> FROM my_table
> WHERE column1 = 'value1' OR column2 = 'value2' OR column3 = 'value3'
> 
> -- AND
> 
> SELECT *
> FROM my_table
> WHERE column1 IN ('value1', 'value2', 'value3')
> ```
> 
> should be equivalent

Yep, thanks for the suggestion!

osopardo1:
After a quick test I can confirm that the IN predicate is also pushdown to the `OTreeIndex`. I would write a function to transform the IN to a AND/OR predicate before building the `QuerySpec`.

Adricu8 (commented):
Changes seem good overall! Thanks for the fast work @osopardo1 ! 💯

src/main/scala/io/qbeast/spark/index/query/QuerySpecBuilder.scala
```diff
@@ -84,49 +96,54 @@ private[spark] class QuerySpecBuilder(sparkFilters: Seq[Expression])
...
+
+        val columnTo = columnFilters
+          .collectFirst {
+            case LessThan(_, Literal(value, _)) => sparkTypeToCoreType(value)
+            case LessThanOrEqual(_, Literal(value, _)) => sparkTypeToCoreType(value)
+            case GreaterThan(Literal(value, _), _) => sparkTypeToCoreType(value)
+            case GreaterThanOrEqual(Literal(value, _), _) => sparkTypeToCoreType(value)
+            case EqualTo(_, Literal(value, _)) => sparkTypeToCoreType(value)
+            case EqualTo(Literal(value, _), _) => sparkTypeToCoreType(value)
+            case IsNull(_) => null
```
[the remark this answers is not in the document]
osopardo1: Because of the Literal placement. It can be placed at the left or right of the operator, and it can reflect different from or to clauses. 

Take as an example` id < 3` and `3 < id`. For the same operator (<) would have different cases. 
In the first one (id < 3) we should filter all the values from 3 to Min(id). In the second one (3 < id), we should filter all the values from 3 to Max(id). 

Probably we can transform those filters before to have a more structured an easy way to parse `from` and `to`, at the same time we parse the IN predicate (or other complex ones).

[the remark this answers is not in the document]
Adricu8: Okay, that makes sense! Good catch :)

osopardo1:
Just a quick summary of the last updates:

- We encountered some bugs when doing pushdown on the IN predicate in case of String columns.
- IN predicates are processed as follows:
    - Match the IN filter passed to the datasource.
    - We can manage IN as a sequence of OR filtering, each one retrieving one slice of the space. But this is not scalable when we have too many values in the set. A solution is to find a range (min, max) and retrieve the blocks belonging to that `QuerySpaceFromTo`. 

Problem: **String Columns are Hashed before indexing**. **Hash do not preserve ordering on Strings**, so when retrieving a space `>=` and/or `<=` than the column, we might miss some records. 

Partial solution: pre-process the IN sequence by hashing before retrieving (min, max). This solves the problem with IN predicate, but does not guarantee that filters that explicitely involve >= and <= would retrieve the correct set of rows.

Let's see an example.

1. `domain IN (a, b, c) `-> We do the transformation. Let's imagine that each hash corresponds to: (a = 1, b = 5, c = 2) and we choose min and max following the numbers. In this case would be a and b (1 and 5). It's ok because it also includes `c`.
2. `domain > a` and `domain < c` -> We do not transform the space. Min and max in this case would be `a` and `c` (1 and 2). `b` is not included in the range values. A possible solution is to pre-process those predicates too. 
3. `domain > b`. In this case, we would search all the values > 5. **This would not include `c`**, which has a hash value of 2. I do not have a solution for that, except to return the whole set of files.

osopardo1:
Another quick update:

- After discussion, we think that we should discard Range Predicates on Strings in the moment. We should filter out those Spark Filters that contain any range expression involving strings before initialising the `QuerySpace`. 

- We would provide an Spark Option to enhance and use those filters if the users can tolerate approximate results.

src/main/scala/io/qbeast/spark/index/query/QueryFiltersUtils.scala
```diff
@@ -0,0 +1,200 @@
...
+        /**
+         * Compare (Any, Int) by using Int comparison
+         */
+        override def compare(x: (Any, Int), y: (Any, Int)): Int =
+          Ordering[Int].compare(x._2, y._2)
+      }
+
+      val min = Literal(valuesHashed.min(ordering)._1, dataType)
+      val max = Literal(valuesHashed.max(ordering)._1, dataType)
+      Seq(LessThanOrEqual(column, max), GreaterThanOrEqual(column, min))
```
alexeiakimov, line 159: Does it make sense? Hash(S1) < Hash(S2) does not imply S1 < S2 and vice versa. Let alone, that string comparison is a subtle matter, in many cases it depends on the locale.

osopardo1: No, it does not, but in HashTransformation (string mapping), we hash the value of the string to determine the position in the space. 

What we are trying to evaluate here, is the range of space hashed that we need to retrieve with Qbeast. 

But today we discussed a bit and we found out that:
- Having a fixed size length for the strings helps to ensure that H(S1) < H(S2) imply S1 < S2. 
- Since we transform the value to a double, we must fit in 2^52. So the maximum number of characters to encode would be 11. (26^11 < 2^52)
- If the dataset has different values with length < 11, we will add 0s at the end of the string to fit the size. 
- If the dataset has different values with length > 11, they would be mapped in the same position, making the algorithm to read more data than the necessary. But at least, no records are lost.

Do you think our last approach makes sense?

core/src/main/scala/io/qbeast/core/model/QuerySpace.scala
```diff
@@ -33,6 +43,8 @@ case class AllSpace() extends QuerySpace {
 case class EmptySpace() extends QuerySpace {
 
   override def intersectsWith(cube: CubeId): Boolean = false
+
+  override def contains(other: QuerySpace): Boolean = false
```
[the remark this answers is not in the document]
osopardo1: Then it should be contained I believe. Thanks.

core/src/main/scala/io/qbeast/core/model/QuerySpace.scala
```diff
@@ -10,6 +10,14 @@ import io.qbeast.core.transform.{HashTransformation, Transformation}
  */
 trait QuerySpace {
 
+  /**
+   * Checks if other query space is contained in the current
+   * @param other the other query space
+   * @return true if they overlap
```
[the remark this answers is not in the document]
osopardo1: Thanks, solved.

src/main/scala/io/qbeast/spark/index/query/QueryFiltersUtils.scala
alexeiakimov: The approach you described above probably makes sense under the following assumptions:
1. All the strings are in Latin, if you have to include Cyrillic or Japanese symbols it will not work.
2. Even if you use Latin it is not completely clear how to compare strings, in general it depends on locale. Just a question which string is greater `esta` or `está`? The rule can depend on the national standards, i.e. locale. A more exotic examples is https://en.wikipedia.org/wiki/%C3%9F in German, as I remember it is a substitute for `ss` and in sorting should be treated as two letters, you may ask Clements for details.
At the same time I can imagine clients or which the string comparison you proposed is still useful, because it is better than nothing.

osopardo1: Ideally the user should be aware of the constraints of their data. We can add an optional parameter to encode the strings with a certain length. 
So, by default the behaviour would be the current one: only a hash, with no possibility to execute range queries over that transformation. If the user finds the solution useful, she or he could configure it for each use case.

osopardo1: Just one comment to start closing things:

Do you think we can merge this PR **with only the necessary changes** to include IN and OR predicate? So no touching anything related to Hash, and then open a new one with #188

alexeiakimov: Feel free to merge anything you need to be merged. I have not started writing the code yet, so not problems.

osopardo1:
Could any of you approve this PR? So we can move forward on #188 

@alexeiakimov @Adricu8 

Thank you!

src/main/scala/io/qbeast/spark/index/query/QueryExecutor.scala
```diff
@@ -19,23 +19,25 @@ class QueryExecutor(querySpecBuilder: QuerySpecBuilder, qbeastSnapshot: QbeastSn
    * Executes the query on each revision according to their QuerySpec
    * @return the final sequence of blocks that match the query
    */
-  def execute(): Seq[QbeastBlock] = {
+  def execute(): Iterable[QbeastBlock] = {
```
Adricu8, line 22: Why this change? The function returns a Seq type, could it lead to issues?

Adricu8 (approved):
I dont have more comments on this PR. Thanks Paola

src/main/scala/io/qbeast/spark/index/query/QueryExecutor.scala
osopardo1: We need the sequence of blocks to be unique. Since OR's queries are a union of queries, it could lead to duplicated blocks. Iterable is a more flexible interface to implement it.

osopardo1: Don't think it could lead to issues. Seq is an Iterable object.

Adricu8: Okay, thanks for the explanation

PR #194, opened 2023-05-26
68 overhead of qbeast hash filtering when doing a sample

### LINKED ISSUE #68: Overhead of qbeast_hash filtering when doing a Sample

alexeiakimov:
## Description
This PR is a work in progress in order to fix #68. It changes the internal implementation of sampling to avoid calling the hash function for every returned row. The key changes are

* Query class which unifies the querying of the staging area and the indexed revisions. Finally it should replace the QueryExecutor class
* QuerySpacesFactory the factory to construct a collection of QuerySpace instance to improve handling of AND and OR predicates
* QbeastSample is a fake filter which is always evaluated to true and is only used to push the sampling information down to the FileIndex inplementation.
* SampleRule is reworked to add QbeastSample filter and for a better support of AND and OR predicated

## Type of change

This is a backward compatible change of the implementation which should not affect client data already stored with the previous version. The goal is to provide a better user experience by making the sampling faster.

## Checklist:

- [ ] Review the correctness of the proposed algorithm
- [ ] Make the existing tests passed
- [ ] Add new tests for the new code where it is necessary
- [ ] Check that the new code really speeds up the sampling

To be done later while testing the performance


osopardo1:
Closing it while we focus on #207 

Afterward, we are going to solve qbeast_hash.

PR #197, opened 2023-06-13
#196 2 files based optimization

alexeiakimov:
## Description

- You may include "Fixes #(issue)": A reference to a related [issue](https://github.com/Qbeast-io/qbeast-spark/issues).

- A clear, concise description of the changes proposed in the pull request.

- (Optional) @mentions of the person or team responsible for reviewing the proposed changes.

## Type of change

Describe the change you're making: how it affects the API, user experience...

- Is it a **Bug fix**? (non-breaking change which fixes an [issue](https://github.com/Qbeast-io/qbeast-spark/issues)).
- Have you developed a **New feature**? (non-breaking change, which adds new functionality).
- Are you introducing a **Breaking change**? (fix or feature that would cause existing functionality not to work or do it differently than previously).
- **Documentation** update (updates regarding the [Users docs](https://github.com/Qbeast-io/qbeast-spark/tree/main/docs) or the [Developer docs](https://docs.qbeast.io/), or similar).

## Checklist:

Here is the list of things you should do before submitting this pull request:

- [ ] New feature / bug fix has been committed following the [Contribution guide](https://github.com/Qbeast-io/qbeast-spark/blob/main/CONTRIBUTING.md).
- [ ] Add comments to the code (make it easier for the community!).
- [ ] Change the documentation.
- [ ] Add tests.
- [ ] Your branch is updated to the main branch (dependent changes have been merged).

## How Has This Been Tested? (Optional)

Please describe the tests that you ran to verify your changes.

- Test A: _(e.g., I have created the class FooTest.scala...)_
- Test B: _(e.g., Running some benchmarks...)_

**Test Configuration**:
* Spark Version:
* Hadoop Version:
* Cluster or local?

PR #200, opened 2023-07-05
199 rework otree algorithm to treat cube files separately

### LINKED ISSUE #199: Rework OTree algorithm to treat the cube files separately
For the needs of new optimization/compaction algorithm #196 and sampling optimization #68 it is necessary to change the part of OTree algorithm which treats the cube file depending on the cube state.

In the current algorithm:
1. Cubes have state, which is one of the following: Flooded, Announced, Replicated
2. Cube files also have state which reflects the cube state at the moment when the file was written.
3. Given a sample query the algorithm checks if the cube has enough data to fulfill the sample, if yes then all the cube files are read regardless their state and the algorithm does not proceed to the child cubes. Otherwise the algorithm checks the cube state, reads the cube files depending on the cube state and proceeds to the child cubes. If the cube is Flooded, then all the cube files are read, if the cube is Announced then all the non-replicated cube files are read, if the cube is Replicated no files are read.
4. Both cube and cube file states are persisted in the Delta transaction log.
5. During optimization the replicated cube identifiers are saved in the transaction log metadata, from which they are read while getting the index snapshot.

In the algorithm to implement:
1. Cube files state is replaced with a boolean attribute "replicated", which has the value true if the file data is replicated to the child cubes.
2. Cube state is computed from the its files, if no files are replicated then it is Flooded, if all the files are replicated when it is Replicated, otherwise it is Announced.
3. Only cube file attributes are persisted in the transaction log.
4. Given a sample query the algorithm checks if the cube has enough data to fulfill the sample, if yes, then all the files are read, otherwise the algorithm only reads the files with the replicated attribute set to true and proceeds to the child cubes.
5. During optimization the replication is applied to cube files separately, for the replicated files the algorithm adds an entry into the transaction log which sets replicated attribute to true.
6. The invariant "a cube can only be replicated if all its ancestors are already replicated" is retired.

More details on the changes and the work plan can be found in https://docs.google.com/document/d/11SNS-dAgDlLt9wyJ_AJzcqQDOUZwXmsX-hKUpgT509o/edit#heading=h.om4an6sc3v0p

alexeiakimov:
## Description

The present PR fixes #199. This is the first step towards new optimization algorithm #197 and query sampling optimization #68.
 
## Type of change

The main change is the different index file format, in particular the string file tag "state" is replaced with the boolean tag "replicated", and the replicated cube identifiers are not stored any longer in the transaction log metadata. Now the cube state is computed from the replicated status of its files.
**Warning!** The implementation of the change is not backward compatible, i.e. the datasets written by the previous versions cannot be read correctly. This incompatibility is planned because the change is planned for the next major release.


src/main/scala/io/qbeast/spark/delta/IndexStatusBuilder.scala
```diff
@@ -91,13 +95,34 @@ private[delta] class IndexStatusBuilder(
...
+        normalizeWeight(col("maxWeight"), col("elementCount"), lit(rev.desiredCubeSize))
+          .as("normalizedWeight"),
         col("files"))
       .as[CubeStatus]
       .collect()
       .foreach(row => builder += row.cubeId -> row)
     builder.result()
   }
 
+  def buildReplicatedAndAnnouncedSets(
```
cugni, line 107: Do we still need to distinguish between replicated and announced? isn't better to manage them together?

src/test/scala/io/qbeast/spark/keeper/ProtocolMockTest.scala
```diff
@@ -4,7 +4,7 @@ import io.qbeast.core.keeper.{Keeper, LocalKeeper}
 
 class ProtocolMockTest extends ProtocolMockTestSpec {
   "the qbeast-spark client" should
-    "throw an execution when an inconstant state is found" in withContext(LocalKeeper) {
+    "throw an execution when an inconstant state is found" ignore withContext(LocalKeeper) {
```
cugni, line 7: I think we can delete these tests now, as we don't have conflicts anymore (just a replication will need to be re-executed). What do you think @osopardo1 ?

alexeiakimov: This (and another one similar) test checks if two concurrent operations lead to conflict, however with the new code the conflict does not occur because we do not write the replicated cubes into the metadata any longer.

src/main/scala/io/qbeast/spark/delta/IndexStatusBuilder.scala
osopardo1: Certainly, it could be some hidden code. As far as I know and remember, the Announced Set was retrieved from the `Keeper` and used for writing new data. The "ANNOUNCED" state on the files was only processed by the `QueryExecutor` in order to filter files.

Since there is only one state for the cube, I think it's best to align the code and not keep ghosts.

src/test/scala/io/qbeast/spark/keeper/ProtocolMockTest.scala
osopardo1: If the conflict those not occur, let's delete them.

src/main/scala/io/qbeast/spark/delta/IndexStatusBuilder.scala
alexeiakimov: It seems that we cannot simply merge them. Let's consider the following scenario:

1. IndexStatus is loaded from the log, so some cubes have all their files replicated and some have some their files replicated. The first ones go to the replicated set and the second ones to the announced set
2. The announced set is extended by the cubes reported by the keeper as announced.
3. Since the keeper is considered to be unreliable it can report some already replicated cubes as if they were still announced, so the cubesToOptimize set is computed as a difference between the announced set and replicated set.

If merge announced and replicated sets then there is a risk to replicate some cube twice. So suggest to leave it as is.

src/main/scala/io/qbeast/spark/delta/writer/BlockStats.scala
```diff
@@ -11,14 +11,14 @@ import io.qbeast.core.model.Weight
  * @param cube the cube
  * @param maxWeight the maximum weight
  * @param minWeight the minimum weight
- * @param state the cube state
+ * @param replicated the cube is replicated
```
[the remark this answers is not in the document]
osopardo1: Nevermind, I saw that it's for the Block stats. Then maybe we should write "the block is replicated" instead.

src/main/scala/io/qbeast/spark/table/IndexedTable.scala
```diff
@@ -257,25 +262,18 @@ private[table] class IndexedTableImpl(
       while (tries > 0) {
         val announcedSet = write.announcedCubes.map(indexStatus.revision.createCubeId)
         val updatedStatus = indexStatus.addAnnouncements(announcedSet)
-        val replicatedSet = updatedStatus.replicatedSet
-        val revisionID = updatedStatus.revision.revisionID
         try {
```
osopardo1, line 265: Then is it necessary to keep the try, catch?

src/test/scala/io/qbeast/spark/keeper/ProtocolMock.scala
```diff
@@ -141,15 +140,11 @@ class WritingProcess(context: ProtoTestContext)(implicit keeper: Keeper) extends
...
-            case cme: ConcurrentModificationException
-                if SparkDeltaMetadataManager.hasConflicts(
-                  tableID,
-                  rev.revisionID,
-                  knownAnnounced,
-                  Set.empty) || tries == 0 =>
-              succeeded = Some(false)
-              throw cme
-            case _: ConcurrentModificationException =>
+            case e: ConcurrentModificationException =>
```
osopardo1, line 143: Same here with the try, catch statement.

src/main/scala/io/qbeast/spark/table/IndexedTable.scala
alexeiakimov: I can imagine a conflict if two concurrent transactions try to create a new revision while writing the data.

src/main/scala/io/qbeast/spark/delta/IndexStatusBuilder.scala
osopardo1: But I think that we discussed that there were no "Replicated" cubes. Since cubes are either declared Announced or not, we don't have a "fully replicated" set of cubes and avoid this type of conflict. 

The problem is to have two sets of "announced" cubes, the one for the Keeper and the one from the Log. The one retrieved from the Keeper should contain the one from the Log, so yes, it could be duplicated. But even if `cubesToOptimize` contains an "already replicated" cube, it does not matter because we are going to take a look at the files before replicating. Is it correct?

src/test/scala/io/qbeast/spark/keeper/ProtocolMockTest.scala
alexeiakimov: done

src/main/scala/io/qbeast/spark/delta/IndexStatusBuilder.scala
alexeiakimov: Yes, if all the files are already replicated then there is nothing to do during the replication. Yes, it can be checked at the later stage when we try to replicate a cube, however the existing code does it earlier and it works, so I prefer to leave this particular part as is. However if you @cugni , @osopardo1 prefer the approach with late check I will rework it.

osopardo1: Oh, I see. For me it's a matter of understandability of what's proposed: to change the number of states of the cubes and the files. 

No hard constraints. If you think is safer to leave the code as it is, we can acknowledge and rework it another time.

osopardo1 (approved):

Issue #201, opened 2023-07-17
Broken Medium Link


osopardo1:
In the main `README.md` we have links to different resources. 

One of them is our blog. Now it appears to be broken, we need to **change it to the following URL**: https://qbeast.io/academy-courses-index/ for the moment.

osopardo1:
Fixed on #203

PR #203, opened 2023-07-25
Fix time-series filtering

### LINKED ISSUE #202: Timestamp filtering returns empty DataFrame
## What went wrong?
Since version 0.3.x we are able to index `Timestamp` and `Date`. But when we try to filter but those Timestamps, the Query is not well parsed and returns an empty DataFrame.

After a few prints and tests, I discovered that the `d.getTime` we use in `LinearTransformer` to initialize `min` and `max` of `LinearTransformation`, returns a Long X. In contrast, the filtering from the user returns a different Long Y. 

This difference does not allow the `QuerySpecBuilder` to initialize the correct space. `QuerySpaceFromTo` returns an `EmptySpace`, which contains 0 files to read. 

Using some code in the `TransformerIndexingTest`:

```scala
    import spark.implicits._
    val df =
      Seq(
        "2017-01-03 12:02:00",
        "2017-01-02 12:02:00",
        "2017-01-02 12:02:00",
        "2017-01-02 12:02:00",
        "2017-01-01 12:02:00",
        "2017-01-01 12:02:00")
        .toDF("date")
        .withColumn("my_date", to_timestamp($"date"))

    df.write.format("qbeast").option("columnsToIndex", "my_date").save(tmpDir)

    val indexed = spark.read.format("qbeast").load(tmpDir)

    indexed.filter("my_date == '2017-01-02 12:02:00'").count() shouldBe 3
```

But the returned DataFrame contains 0 rows.

### 2. Branch and commit id:

I'm using photon branch: `photon-datasource-standalone`

### 3. Spark version:

Spark version 3.3.0 and Delta version 2.1.0

### 4. Hadoop version:

3.3.4

Local

osopardo1:
## Description

Fixes #202 (and #201 😃 )

## Type of change

It is a Bug Fix for the reading part, it does not include breaking changes on the algorithm or the writing of the files.

Now `TimestampType` and `DateType` are pre-processed on `QueryFilterUtils` to return a Millisecond representation that can be filtered properly by the index. 

## Checklist:

- [x] Add tests.

I've added a test on `src/test/scala/io/qbeast/spark/index/query` called `TimeSeriesQueryTest`. In this class, I've tested four different ways of creating an index on a Date / Timestamp column. Each one were failing differently, and should fail if executed with previous commits. 

**Test Configuration**:
 TBD.


osopardo1:
To have a fixed version for 0.3.x for compatibility with Spark 3.2.x, the commits of this PR should be merged into a new 0.3.5 version release, and updated accordingly.

osopardo1:
Hello @alexeiakimov ! Could you review this quick fix? Thank you 🙌

src/main/scala/io/qbeast/spark/index/query/QueryFiltersUtils.scala
```diff
@@ -91,15 +96,25 @@ private[query] trait QueryFiltersUtils {
   }
 
   /**
-   * Convert an Spark String type to a Scala core type
-   * @param value the value to convert
+   * Convert an Literal value from Spark to a Qbeast/Scala core type
```
alexeiakimov, line 99: Probably typo: a literal value

src/main/scala/io/qbeast/spark/index/query/QueryFiltersUtils.scala
```diff
@@ -91,15 +96,25 @@ private[query] trait QueryFiltersUtils {
...
-  def sparkTypeToCoreType(value: Any): Any = {
-    value match {
-      case s: UTF8String => s.toString
-      case _ => value
+  def sparkTypeToCoreType(l: Literal): Any = {
+
+    (l.value, l.dataType) match {
+      case (int: Integer, _: DateType) =>
+        // convert DateType to Milliseconds
+        lazy val zoneId = getZoneId(SQLConf.get.sessionLocalTimeZone)
```
alexeiakimov, line 109: I wonder that the data is always stored as the Spark session local timezone, right?

src/main/scala/io/qbeast/spark/index/query/QueryFiltersUtils.scala
```diff
@@ -91,15 +96,25 @@ private[query] trait QueryFiltersUtils {
...
+        val dateInMicros = daysToMicros(int, zoneId)
+        val dateInMillis = TimeUnit.MILLISECONDS.convert(dateInMicros, TimeUnit.MICROSECONDS)
```
alexeiakimov, line 111: Can be shorter:
```Scala
TimeUnit.MICROSECONDS.toMillis(dateInMicros)
```

src/main/scala/io/qbeast/spark/index/query/QueryFiltersUtils.scala
```diff
@@ -91,15 +96,25 @@ private[query] trait QueryFiltersUtils {
...
+        dateInMillis
+      case (long: Long, _: TimestampType) =>
+        // convert Timestamp from Microseconds to Milliseconds
+        TimeUnit.MILLISECONDS.convert(long, TimeUnit.MICROSECONDS)
```
alexeiakimov, line 115: This also can be shorter.

src/main/scala/io/qbeast/spark/index/query/QueryFiltersUtils.scala
alexeiakimov: Do you mean that the data is persisted on disk using the timezone of Spark?

src/main/scala/io/qbeast/spark/index/query/QueryFiltersUtils.scala
osopardo1: Thanks! Done 👍

src/main/scala/io/qbeast/spark/index/query/QueryFiltersUtils.scala
osopardo1: According to this post: https://www.databricks.com/blog/2020/07/22/a-comprehensive-look-at-dates-and-timestamps-in-apache-spark-3-0.html If no Timezone is specified, **Spark takes a time zone from the SQL configuration `spark.sql.session.timeZone` and applies it to function invocations**. 

So yes, data would be persisted on the Disk using the `spark.sql.session.timeZone`.

alexeiakimov: My bad, after reading the article I think the code is correct.

alexeiakimov (approved):

osopardo1:
Fixed #202 and #201

PR #211, opened 2023-08-30
Upgrade to Spark 3.4.1 and Delta 2.4.0

osopardo1:
This Draft PR is for updating versions of Spark and Delta to the latest available!

Here's a summary of some interesting changes that are included in the updates. 

## Spark 3.4.x

Read the full notes here: [Apache Spark Version 3.4 ](https://spark.apache.org/releases/spark-release-3-4-0.html), [Apache Spark Version 3.4.1](https://spark.apache.org/releases/spark-release-3-4-1.html)

- **Python Client for Spark Connect**.
- Timestamp without timezone support. (`TimestampNTZ` Type ).
- Support **timestamp in seconds** for TimeTravel using Dataframe options
- Better Spark UI scalability and Driver stability.
- **Catalog API compatible with 3 Layer Namespace**. 

## Delta 2.4.0

Read the full notes here: [Delta Lake Version 2.4.0](https://github.com/delta-io/delta/releases/tag/v2.4.0)
- Support for **Apache Spark 3.4**
- **Deletion Vectors implementation**. Improve the way UPDATES are handled and recovered, adopting a "Merge On Read" strategy.
- Support for writing and reading with Deletion Vectors and `PURGE` in case we want to deactivate the feature for the table (which will invoke a rewrite of some of the files).
- **Support `REPLACE WHERE` expressions** in SQL to selectively overwrite data.
- **Change on the `DeltaLog` API**: Now `snapshot` is no longer reliable to return the last Snapshot available. Use `DeltaLog.update()` instead. 
- NOTE: This version **does not include the [Iceberg to Delta converter](https://docs.delta.io/2.4.0/delta-utility.html#convert-an-iceberg-table-to-a-delta-table)**

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,9 +18,9 @@
 
 **Qbeast Spark** is an extension for [**Data Lakehouses**](http://cidrdb.org/cidr2021/papers/cidr2021_paper17.pdf) that enables **multi-dimensional filtering** and **sampling** directly on the storage
 
-[![apache-spark](https://img.shields.io/badge/apache--spark-3.3.x-blue)](https://spark.apache.org/releases/spark-release-3-2-2.html) 
+[![apache-spark](https://img.shields.io/badge/apache--spark-3.4.x-blue)](https://spark.apache.org/releases/spark-release-3-4-1.html) 
 [![apache-hadoop](https://img.shields.io/badge/apache--hadoop-3.3.x-blue)](https://hadoop.apache.org/release/3.3.1.html)
-[![delta-core](https://img.shields.io/badge/delta--core-2.1.0-blue)](https://github.com/delta-io/delta/releases/tag/v1.2.0)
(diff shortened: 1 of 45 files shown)


osopardo1:
The current status is the following: 

- New APIs of `Table` and `DeltaLog` adapted to the current code. 
- Introducing compatibility with TimestampNTZ.

osopardo1:
Following the previous comment, I investigated a little bit more about the problem and opened this [issue](#213)

I will discard all the code to solve the INSERT INTO with TIMESTAMP_NTZ here, so we can decouple both developments. 
And before continuing with the Upgrade, we should fix #213 first.

osopardo1:
Pending TODO: Support TIMESTAMP_NTZ. That would be for the next iteration. I would open an issue.

cugni (approved):

PR #217, opened 2023-09-15
Fix soc2 issues

joancarles-qbeast:
## Description

In Vanta vulnerabilities section, are identified some issues with different severities:

image

## Type of change

It's necessary to update the versions with the recommended by the CVE bulletins. The dependabot detected this issues and Vanta show them in the security dashboard.

The change basically is focused in the requirements.txt file , updating the versions of the dependencies.

There are minor changes in the built.sbt , on the Sonatype section, that are shown in the test phase due a deprecated syntax.

## How Has This Been Tested?

The changes has been tested running the test defined in the workflow.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -104,7 +104,8 @@ ThisBuild / publishTo := {
 }
 
 // Sonatype settings
-ThisBuild / publishMavenStyle := true
+//ThisBuild / publishMavenStyle := true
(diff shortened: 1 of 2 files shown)


osopardo1:
LGTM!! Let's wait for the test to pass and we can merge 👍

PR #221, opened 2023-10-24
String indexing via column histograms

Jiaweihu08:
## Description
This PR introduces a new **String** `Transformation` that uses String histograms for value mapping.

### How does it work?
A sorted sequence of distinct string values should be provided to map String values to the space.

The sequence is treated as an equal-width histogram for the column to index.

To transforma a given **String** value, we look for its **insertion position** within the sequence using binary search:
```scala
val hist = ["a", "b", "c", "d", "e"]
val coordinate = hist.search("b").insertionPoint.toDouble / (hist.length - 1) // 0.25
```

Values that are not contained within the limits of the histogram are mapped to their corresponding extremes, i.e. `0.0` or `1.0`. 

### How to use?
To use the feature, we build the histogram and provide it as part of `columnStats` when writing.

`:histogram` should also be specified for the String column to index.

1. Compute the histogram:
- Use all distinct String values:
```scala
import org.apache.spark.sql.DataFrame
import org.apache.spark.sql.functions.col

def getAllString(df: DataFrame, columnName: String): String = {	
  df
  .select(columnName)
  .distinct()
  .na.drop
  .orderBy(col(columnName).asc)
  .collect()
  .map { r => 
    val s = r.getAs[String](0)
    s"'$s'"
  }
  .mkString("[", ",", "]")
}

val histogram = getAllString(df, "test_col_name")
```

- Use a reduced number of String values:
```scala
import org.apache.spark.sql.delta.skipping.MultiDimClusteringFunctions
import org.apache.spark.sql.DataFrame
import org.apache.spark.sql.functions.{col, min}

def getStringHistogramStr(df: DataFrame, columnName: String, numBins: Int): String = {
  val binStarts = "__bin_starts"
  val stringPartitionColumn = MultiDimClusteringFunctions.range_partition_id(col(columnName), numBins)
	
  df
  .select(columnName)
  .distinct()
  .na.drop
  .groupBy(stringPartitionColumn)
  .agg(min(columnName).alias(binStarts))
  .select(binStarts)
  .orderBy(binStarts)
  .collect()
  .map { r => 
    val s = r.getAs[String](0)
    s"'$s'"
  }
  .mkString("[", ",", "]")
}

val histogram = getStringHistogramStr(df, "test_col_name", 100)
```

2. Index the data as follows:
```scala
val columnStats = s"""{"test_col_name_histogram":$histogram}"""

df
  .write
  .format("qbeast")
  .option("columnsToIndex", s"test_col_name:histogram")
  .option("columnStats", columnStats)
  .save(targetPath)
```

If no histogram is provided in `columnStats` during the first write, a default histogram("a" to "z") will be used. Subsequent `appends` will reuse the same histogram.

A new revision will be created when a different custom histogram is provided as "columnStats".

Note: It also fixes the bug where the first write has to have `overwrite` write mode.

## Checklist:

- [x] Add tests.
- [x] Branch is updated to the main branch.

### DIFF
--- a/core/src/main/scala/io/qbeast/core/transform/StringHistogramTransformer.scala
+++ b/core/src/main/scala/io/qbeast/core/transform/StringHistogramTransformer.scala
@@ -7,7 +7,7 @@ object StringHistogramTransformer extends TransformerType {
   override def transformerSimpleName: String = "string_hist"
 
   // "a" to "z"
-  val defaultHist: Array[String] = (97 to 122).map(_.toChar.toString).toArray
+  val defaultHist: IndexedSeq[String] = (97 to 122).map(_.toChar.toString)
 
   private val defaultHistStr: String = defaultHist.mkString("Array('", "', '", "')")
 }
@@ -35,7 +35,7 @@ case class StringHistogramTransformer(columnName: String, dataType: QDataType)
    */
   override def makeTransformation(row: String => Any): Transformation = {
     val hist = row(colHist) match {
-      case h: Seq[_] => h.map(_.toString).toArray
+      case h: Seq[_] => h.map(_.toString).toIndexedSeq
       case _ => defaultHist
     }
 
--- a/core/src/main/scala/io/qbeast/core/transform/StringHistogramTransformer.scala
+++ b/core/src/main/scala/io/qbeast/core/transform/StringHistogramTransformer.scala
@@ -1,10 +1,15 @@
 package io.qbeast.core.transform
 
 import io.qbeast.core.model.QDataType
+import io.qbeast.core.transform.StringHistogramTransformer.{defaultHist, defaultHistStr}
 
 object StringHistogramTransformer extends TransformerType {
   override def transformerSimpleName: String = "string_hist"
 
+  // "a" to "z"
+  val defaultHist: Array[String] = (97 to 122).map(_.toChar.toString).toArray
+
+  private val defaultHistStr: String = defaultHist.mkString("Array('", "', '", "')")
 }
 
 case class StringHistogramTransformer(columnName: String, dataType: QDataType)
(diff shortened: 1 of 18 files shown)


core/src/main/scala/io/qbeast/core/transform/StringHistogramTransformer.scala
```diff
@@ -0,0 +1,45 @@
+package io.qbeast.core.transform
+
+import io.qbeast.core.model.QDataType
+import io.qbeast.core.transform.StringHistogramTransformer.{defaultHist, defaultHistStr}
+
+object StringHistogramTransformer extends TransformerType {
+  override def transformerSimpleName: String = "string_hist"
+
+  // "a" to "z"
+  val defaultHist: IndexedSeq[String] = (97 to 122).map(_.toChar.toString)
```
osopardo1, line 10: Thinking if the defaultHist could be extracted from the data if the user allows it so...

docs/AdvancedConfiguration.md
```diff
@@ -115,6 +115,57 @@ val columnStats =
...
+2. support for range queries on String columns
+3. improved overall query speed
+
+The following code snippet demonstrates the extraction of a **String** histogram from the source data:
+```scala
+ import org.apache.spark.sql.delta.skipping.MultiDimClusteringFunctions
+ import org.apache.spark.sql.DataFrame
+ import org.apache.spark.sql.functions.{col, min}
+
+ def getStringHistogramStr(columnName: String, numBins: Int, df: DataFrame): String = {
```
osopardo1, line 133: Maybe not for this PR, but would be nice if the user could call this code or that could activate an option to let qbeast-spark build it internally (for example: spark.qbeast.analyze.histogram.enabled)

what do you think?

osopardo1: For next steps.

core/src/main/scala/io/qbeast/core/transform/StringHistogramTransformer.scala
osopardo1: For next steps.

cdelfosse:
To merge into the main-1.0.0a branch with Alexey's compaction work.

osopardo1 (dismissed):
All good!

osopardo1:
I am closing this PR, which is replaced by #234 

Since the feature is merged in `main-1.0.0`, we will use the commit there to push it to `main`.

PR #232, opened 2023-11-14
207 rollup compaction simplified

alexeiakimov:
## Description

The present PR provides initial implementation for indexing and optimization based on the rollup compaction as it is proposed in the #207. The most important changes are
1. A single index file can contain blocks from different cubes
2. The format of the file metadata stored in the tags of `AddFile` has changed to support multiple blocks.
3. Cube state is computed from the `replicated` flags of its blocks and id not persisted in the DeltaLog metadata anymore.
4. The Qbeast metadata appended to the metadata of the Delta has changed to exclude the information about the replicated cubes.
5. While indexing the cube blocks to be stored together in the same physical index file are grouped using the rollup compaction algorithm.
6. While compacting the cube blocks to be stored together in the same physical file are grouped using the rollup compaction algorithm. 

## Type of change

This PR introduces incompatible changes in the metadata format for both individual index files and for the DeltaLog as well. The tables written with the previous version cannot be read by the present version, and vice versa. However the existing tables can be converted (int theory) to the new format by transforming the AddFile tags and DeltaLog metadata without touching the actual files.

The correctness tests are provided in the project code. Also @Jiaweihu08 did some extra performance test comparing the proposed version with the previous one and pure Delta.


src/test/scala/io/qbeast/spark/utils/QbeastFilterPushdownTest.scala
```diff
@@ -37,7 +36,7 @@ class QbeastFilterPushdownTest extends QbeastIntegrationTestSpec {
           val matchingFiles =
             index.listFiles(f.partitionFilters, f.dataFilters).flatMap(_.files)
           val allFiles = index.inputFiles
-          matchingFiles.length shouldBe <(allFiles.length)
+          matchingFiles.length shouldBe <=(allFiles.length)
```
cugni, line 39: it makes no sense to have this test with <=. It will always <=, and it doesn't check if there is actually any push down filtering. I understand that with the rollup this test might fail now, but it is better to either fix the test or deleted.

cugni (approved):
A lot of files changes, most of them small refactoring, but it is a bit hard to review like this. But the central class seems well written and the logic the one discussed. I only have 1 doubt about a test.

src/test/scala/io/qbeast/spark/utils/QbeastFilterPushdownTest.scala
alexeiakimov: Fixed.

osopardo1:
We are finally merging the rollup!! Great job!! @alexeiakimov

PR #239, opened 2023-11-28
Add Delta Data Skipping on Staging Area

osopardo1:
## Description

Fixes #235

## Type of change

It is an enhancement, it does not affect the current writing algorithm/format.

## Checklist:

Here is the list of things you should do before submitting this pull request:

- [x] New feature / bug fix has been committed following the [Contribution guide](https://github.com/Qbeast-io/qbeast-spark/blob/main/CONTRIBUTING.md).
- [x] Add comments to the code (make it easier for the community!).
- [x] Add tests.
- [x] Your branch is updated to the `main-1.0.0` branch (dependent changes have been merged).

## How Has This Been Tested? (Optional)

Added test to `OTreeIndexTest` to check if staging files are filtered properly.

PR #245, opened 2023-12-04
Add Delta Data Skipping on Staging Area

### LINKED ISSUE #235: Add Delta File Skipping for non-indexed files
We can have hybrid tables with files from Delta and Qbeast. In order to efficiently filter files on queries, we can use the underlying format Data Skipping for those files that are not indexed.

osopardo1:
Cherry-pick for #239 

### Summary

Fixes #235 

One feature in Qbeast Format is that we can have hybrid tables with files from Delta and Qbeast.

Qbeast provides its own filtering code, which does not skip files without Qbeast Metadata (it returns the whole collection).  To efficiently filter files on queries, we can use the underlying format Data Skipping for joining the Qbeast result. This way we make sure we output the best of the file filtering, especially in files that are written in the staging area (those not indexed).


osopardo1:
Closing it because I will use the original commits to merg. Sorry for the inconvenience.

osopardo1:
This feature is merged on `main-1.0.0`, but needs to be on `main`.

Issue #244, opened 2023-12-04
Add AutoIndexing


osopardo1:
Qbeast Format allows indexing multiple columns to improve the file layout. Right now, the user should explicitly assign the set of columns that she/he wants to index through the `columnsToIndex` parameter. 

The idea is to **get rid of enforcing parameters in the configuration** and allow choosing the more valuable columns through a Correlation Matrix Analysis. We will **order the columns from least to max average correlation and then filter the top n columns to choose the `columnsToIndex`**.

There are two cases in which we should integrate the code for calling the AutoIndexer:

1. **If no `columnsToIndex` is specified**. Use a correlation matrix as a default strategy to choose which columns contain between the 50%-70% of the variance.
2. **If the option `spark.qbeast.index.autoIndexerEnabled`** (we can change the name, this is just a suggestion) is set to true. 

In either of those cases, the pipeline in `IndexedTable.save` should work as follows:

1. Check if `columnsToIndex` is empty / `spark.qbeast.index.autoIndexerEnabled ` is set.
2. If the result is positive, call `SparkAutoIndexer` code to choose the columns to index. The number of columns indexed should be specified with either: an option or a default parameter (`spark.qbeast.index.maxColumnsToIndex`). [TBD]
3. Create a new `Revision` with the output columns from the AutoIndexer. No other workflow is touched. 

We should:

- [x] Determine **which is the variance percentage desired to choose a column**.
- [x] Add the Correlation Matrix code to qbeast-spark.
- [x] Integrate with existing API.
- [x] Add tests.

osopardo1:
FYI: we don't use the full PCA. We just partly analyze which columns contain higher variance, without having to create new mapping columns within the dataframe.

osopardo1:
I will add the corresponding skeleton to call the PCA method, and tomorrow we can work on integrating @SrTangente code.

SrTangente:
WIP, we use a correlation matrix to order the columns from least to max avarega correlation and then filter the top n columns

osopardo1:
> WIP, we use a correlation matrix to order the columns from least to max avarega correlation and then filter the top n columns

Description updated, thanks!

osopardo1:
This feature is merged in 1.0.0-main. Waiting for addition into main when the release is made.

osopardo1:
Merged on https://github.com/Qbeast-io/qbeast-spark/pull/284

PR #247, opened 2023-12-05
Auto Indexing

osopardo1:
## Description

Adds new Auto Indexing functionality #244 

## Type of change

A new feature that enables indexing a DataFrame/Table without specifying the `columnsToIndex`. 

The feature is not enabled by default. If you want to use it, you should add the necessary configuration. 

```shell
spark.qbeast.index.columnsToIndex.auto=true
spark.qbeast.index.columnsToIndex.auto.max=10
```

### Code

The idea is to have an `ColumnsToIndexSelector` interface in the `core` project with the following information: 

```scala

trait ColumnsToIndexSelector[DATA] {

  /**
   * The maximum number of columns to index.
   * @return
   */
  def MAX_COLUMNS_TO_INDEX: Int

  /**
   * Selects the columns to index given a DataFrame
   * @param data
   *   the data to index
   * @return
   */
  def selectColumnsToIndex(data: DATA): Seq[String] =
    selectColumnsToIndex(data, MAX_COLUMNS_TO_INDEX)

  /**
   * Selects the columns to index with a given number of columns to index
   * @param data
   *   the data to index
   * @param numColumnsToIndex
   *   the number of columns to index
   * @return
   *   A sequence with the names of the columns to index
   */
  def selectColumnsToIndex(data: DATA, numColumnsToIndex: Int): Seq[String]

}
```
- `MAX_COLUMNS_TO_INDEX`: maximum number of columns to index for this implementation (could be a configurable parameter)
- `selectColumnsToIndex`: this method is called every time we build a new table, only once (if we do not want to update it).

## Checklist:

- [x] Add Skeleton for extensible AutoIndexing code (support for different algorithms and engines)
- [x] Add default PCA code 
- [x] Add tests.
- [x] Your branch is updated to the `main-1.0.0` branch (dependent changes have been merged).

This should be tested individually (each AutoIndexer should have unit tests) as well as Integrated with the Spark API (DataFrame reads and writes).


osopardo1:
Codecov is failing, but test pass and PR is ready to be reviewed.

core/src/main/scala/io/qbeast/core/model/AutoIndexer.scala
```diff
@@ -0,0 +1,36 @@
...
+   */
+  def MAX_COLUMNS_TO_INDEX: Int
+
+  /**
+   * Chooses the columns to index without limiting the number of columns.
+   * @param data
+   *   the data to index
+   * @return
+   */
+  def chooseColumnsToIndex(data: DATA): Seq[String]
```
alexeiakimov, line 22: Since MAX_COLUMNS_TO_INDEX is a part of the API, then this method can have a default implementation. It will reduce the work for those who is going to implement the trait.

core/src/main/scala/io/qbeast/core/model/AutoIndexer.scala
```diff
@@ -0,0 +1,36 @@
+package io.qbeast.core.model
+
+/**
+ * AutoIndexer interface to automatically choose which columns to index.
+ * @tparam DATA
+ *   the data to index
+ */
+trait AutoIndexer[DATA] {
```
alexeiakimov, line 8: The name is not what excites me. IMHO it sounds like it is responsible for indexing, that is not the case. However since I cannot provide anything better than `ColumnsToIndexChooser`, you can leave it as is.

src/main/scala/io/qbeast/spark/index/SparkAutoIndexer.scala
```diff
@@ -0,0 +1,122 @@
...
+   * Chooses the columns to index without limiting the number of columns.
+   *
+   * @param data
+   *   the data to index
+   * @return
+   */
+  override def chooseColumnsToIndex(data: DataFrame): Seq[String] =
+    chooseColumnsToIndex(data, MAX_COLUMNS_TO_INDEX)
+
+  override def chooseColumnsToIndex(data: DataFrame, numColumnsToIndex: Int): Seq[String] = {
```
alexeiakimov, line 37: The method is too long to my taste. I would decompose it into a series of calls using the comments you put as the names of the submethods. Again it is a matter of taste.

src/main/scala/io/qbeast/spark/index/SparkAutoIndexer.scala
```diff
@@ -0,0 +1,122 @@
...
+      return data.columns.take(numColumnsToIndex)
+    }
+
+    var updatedData = data
+    val inputCols = data.columns
+
+    // Convert timestamp columns to Unix timestamps and update the DataFrame
+    val timestampCols = inputCols.filter(colName =>
+      data.schema(colName).dataType == org.apache.spark.sql.types.TimestampType)
+    timestampCols.foreach { colName =>
```
alexeiakimov, line 50: Scala zealots like to chain two closures, so the local variable timestampCols is redundant.

src/main/scala/io/qbeast/spark/index/SparkAutoIndexer.scala
```diff
@@ -0,0 +1,122 @@
...
+    if (data.isEmpty) {
+      return data.columns.take(numColumnsToIndex)
+    }
+
+    var updatedData = data
+    val inputCols = data.columns
+
+    // Convert timestamp columns to Unix timestamps and update the DataFrame
+    val timestampCols = inputCols.filter(colName =>
+      data.schema(colName).dataType == org.apache.spark.sql.types.TimestampType)
```
alexeiakimov, line 49: I wonder why do we need to use fully qualified names for the Spark types like TimestampType, StringType, etc? Are there any import conflict with other code, from prelude or alike? If not, then I would suggest to import those symbols and to use short names, the readability of the code will be better.

src/main/scala/io/qbeast/spark/index/SparkAutoIndexer.scala
```diff
@@ -0,0 +1,122 @@
...
+    val transformedToOriginal = inputCols.flatMap { colName =>
+      val transformedName = s"${colName}_Vec"
+      Some(transformedName -> colName)
+    }.toMap
+
+    // Select the columns based on the selected indices
+    val selectedTransformedColumns = selectedIndices.map(columnNames(_))
+
+    // Map the selected transformed column names back to original column names
+    val selectedOriginalColumns = selectedTransformedColumns.map(transformedToOriginal)
```
alexeiakimov, line 117: This local variable is redundant here.

core/src/main/scala/io/qbeast/core/model/AutoIndexer.scala
alexeiakimov: This class is pretty technical, so IMHO the name can follow the name of its main member. Currently its duty is to `chooseColumsToIndex`, the verb is "choose", thus we can derive a noun for the class i.e. `...Chooser` (Chooser is a quite used word in software, for example FileChooser in UI). Since it chooses columns to index, I put these words to the beginning. This is how I got `ColumnsToIndexChooser`.
If you prefer "select", then same logic leads us to `ColumnsToIndexSelector`. I believe that it is a common practice to move the objective before the noun derived from the verb representing the main operation. At least for small technical classes which do just one thing. For bigger entities like services with many operations it is a bit different.

osopardo1: Yep, I understand. In my opinion the word `select` feels more elegant. I still do not like the whole composition as `ColumnsToIndexSelector`, but it is as descriptive as it can be. 

For the same reasoning argument, I would like to also change the configuration parameters. 

**Which one feels best?**

```scala
spark.qbeast.index.columnsToIndex.auto=true
spark.qbeast.index.columnsToIndex.max=10
```
 or

```scala
spark.qbeast.index.columnSelector.auto=true
spark.qbeast.index.columnSelector.max=10
```

or 

```scala
spark.qbeast.index.columnsToIndexSelector.auto=true
spark.qbeast.index.columnsToIndexSelector.max=10
```

src/main/scala/io/qbeast/spark/index/SparkAutoIndexer.scala
osopardo1: Mm, I would take a look. I would also check if some of the methods can be generalized in the `trait`.

core/src/main/scala/io/qbeast/core/model/AutoIndexer.scala
alexeiakimov: I think the first option is the best. Because they are for the end user, who knows that the columsToIndex can be chosen automatically. So columnsToIndex.auto is good. Unfortunately columnsToIndex.max is misleading, because one can think that it is the maximum number of columns to specified by columnsToIndex option.

src/main/scala/io/qbeast/spark/index/SparkAutoIndexer.scala
alexeiakimov: I do not have to extract a trait, i.e. introduce a new abstraction, I mean just to extract a few functions to be implemented in the same class next to the mentioned method.

core/src/main/scala/io/qbeast/core/model/AutoIndexer.scala
alexeiakimov: The following variants look good to me:
```
spark.qbeast.index.columnsToIndex.auto=true
spark.qbeast.index.columnsToIndex.auto.max=10
```
or
```
spark.qbeast.index.columnsToIndex.auto.enabled=true
spark.qbeast.index.columnsToIndex.auto.max=10
```
I would go with the first of the two above for brevity.

alexeiakimov: These options are for the end user at least in theory, so they should be named in terms the user can observe and understand. So the variants with "selector" expose and refer to the name of the internal abstraction which is not visible to the user.

osopardo1: Those seems appropiate: 
```
spark.qbeast.index.columnsToIndex.auto=true
spark.qbeast.index.columnsToIndex.auto.max=10
```

Ok, thanks!

src/main/scala/io/qbeast/spark/index/SparkAutoIndexer.scala
osopardo1: Done!

src/main/scala/io/qbeast/spark/index/SparkAutoIndexer.scala
osopardo1: Turned the column selection in a single step

src/main/scala/io/qbeast/spark/index/SparkAutoIndexer.scala
osopardo1: I've tried to separate the code in 3 methods:

- withUnixTimestamp (converts columns with timestamp to unix_timestamp)
- withPreprocessedPipeline (which executes the ML Pipeline and returns the DataFrame with correlation matrix)
- getTopNCorrelatedColumns (selects the top N correlated columns.)

docs/ColumnsToIndexSelector.md
```diff
@@ -0,0 +1,73 @@
...
+
+```scala
+spark.sql("CREATE TABLE table_name USING qbeast LOCATION 'path/to/table'")
+```
+### Interface
+
+The Auto Indexer is an interface that can be implemented by different classes. The interface is defined as follows:
+
+```scala
+trait AutoIndexer[DATA] {
```
alexeiakimov, line 30: Please correct the trait name. Thank you.

docs/ColumnsToIndexSelector.md
```diff
@@ -0,0 +1,73 @@
...
+   * Chooses the columns to index.
+   * @param data
+   *   the data to index
+   * @param numColumnsToIndex
+   *   the number of columns to index if not specified, the default value is
+   *   MAX_NUM_COLUMNS_TO_INDEX
+   * @return
+   *   A sequence with the names of the columns to index
+   */
+  def chooseColumnsToIndex(data: DATA, numColumnsToIndex: Int): Seq[String]
```
alexeiakimov, line 56: Please correct the method name.

alexeiakimov (approved):

PR #248, opened 2023-12-05
CREATE EXTERNAL TABLE without OPTIONS

### LINKED ISSUE #246: CREATE EXTERNAL TABLE with LOCATION should load existing configuration
## What went wrong?

If we have an existing table in a location `L`, the only way of populating the Table in the Catalog without rewriting the data is to CREATE EXTERNAL TABLE with LOCATION L. The current implementation forces us to add `columnsToIndex` parameters to the `OPTION`  SQL clause because we use the **same interface for creating a new table and linking a Table to an external directory**. 

```scala
      val location = tmpDir + "/external_student/"
      val data = createTestData(spark)
      data.write.format("qbeast").option("columnsToIndex", "id,name").save(location)

      spark.sql(
        s"CREATE EXTERNAL TABLE student (id INT, name STRING, age INT) " +
          s"USING qbeast " +
          s"LOCATION '$location'")
```

### 2. Branch and commit id:

`main`

### 3. Spark version:

3.4.1

### 4. Hadoop version:

3.3.1

Local computer

osopardo1:
## Description

Fixes #246 

## Type of change

Allows the creation of an already physically formatted Qbeast Table without the need to specify the options `columnsToIndex` or `cubeSize`. **In case the table is external, exists, and has Qbeast Metadata, it loads the required parameters from the `QbeastSnapshot`.** 

Let's see an example:

```scala

val df = Seq(1, 2, 3).toDF("id")
df.write.format("qbeast").option("columnsToIndex", "id").save("/tmp/qbeast_table")

spark.sql("CREATE EXTERNAL TABLE t USING qbeast LOCATION '/tmp/qbeast_table'")
```

## Checklist:

- [x] Add tests.

It is tested in `QbeastSQLIntegrationTest`.


src/main/scala/io/qbeast/spark/table/IndexedTable.scala
```diff
@@ -30,13 +30,26 @@ trait IndexedTable {
    */
   def exists: Boolean
 
+  /**
+   * Returns whether the table is converted to Qbeast format.
+   * @return
+   */
+  def isConverted: Boolean
```
alexeiakimov, line 37: IMHO this name could be better. The problem with "Converted" is that it seems that the table had a different format, but later was converted to Qbeast. What if the table was Qbeast from the very beginning. What question does the method tries to answer?

osopardo1: It tries to answer if this particular table contains Qbeast Metadata of any type (staging, revision..). 

If the precondition of creating an `IndexedTable` is that this Table is already indexed (or converted), I will remove the condition and add checks before initializing the particular IndexedTable.

osopardo1:
Adding some tests I've found a weird behavior when passing Options/Paramaters/Properties back and forth. 

For example, if a Table already exists with Options A and B, the user can:
- **Not specify options**, and A and B should be considered.
- **Specify options**, and A and B should be verified and overriden if necessary. 

In the second case, options A and B weren't passed to the save methods, and subsequently not overridden. 

Fixing them!!

src/main/scala/io/qbeast/spark/table/IndexedTable.scala
alexeiakimov: Then I would suggest to name the method as `hasQbeastMetadata` or alike.

osopardo1: Done!

alexeiakimov (approved):

PR #249, opened 2023-12-05
243 investigate streaming options

alexeiakimov:
## Description

The set of supported options is extended to let the user to control the transaction identity and thus to make the writes idempotent.

## Type of change

This is a new feature borrowed from Delta which adds two supported options `txnAppId` and `txnVersion` to be used to identify the write transaction. Once committed any other transaction with the same properties will be ignored as already committed. This feature allows to make the writes idempotent.

No breaking changes have been introduced.

## Checklist:

- [x] Add tests.

A new integration test `io.qbeast.spark.utils.QbeastSparkTxnTest` is added to check the transaction handling.

### DIFF
--- a/core/src/main/scala/io/qbeast/core/model/MetadataManager.scala
+++ b/core/src/main/scala/io/qbeast/core/model/MetadataManager.scala
@@ -8,8 +8,10 @@ import io.qbeast.IISeq
  *   type of data schema
  * @tparam FileDescriptor
  *   type of file descriptor
+ * @tparam QbeastOptions
+ *   type of the Qbeast options
  */
-trait MetadataManager[DataSchema, FileDescriptor] {
+trait MetadataManager[DataSchema, FileDescriptor, QbeastOptions] {
   type Configuration = Map[String, String]
 
   /**
@@ -36,13 +38,16 @@ trait MetadataManager[DataSchema, FileDescriptor] {
    *   the QTableID
    * @param schema
    *   the schema of the data
+   * @param options
+   *   the update options
    * @param append
    *   the append flag
-   * @param writer
-   *   the writer code to be executed
    */
-  def updateWithTransaction(tableID: QTableID, schema: DataSchema, append: Boolean)(
-      writer: => (TableChanges, IISeq[FileDescriptor])): Unit
+  def updateWithTransaction(
+      tableID: QTableID,
+      schema: DataSchema,
+      options: QbeastOptions,
+      append: Boolean)(writer: => (TableChanges, IISeq[FileDescriptor])): Unit
 
   /**
    * Updates the table metadata by overwriting the metadata configurations with the provided
--- a/src/main/scala/io/qbeast/spark/delta/SparkDeltaMetadataManager.scala
+++ b/src/main/scala/io/qbeast/spark/delta/SparkDeltaMetadataManager.scala
@@ -4,6 +4,7 @@
 package io.qbeast.spark.delta
 
 import io.qbeast.core.model._
+import io.qbeast.spark.internal.QbeastOptions
 import io.qbeast.IISeq
 import org.apache.spark.sql.delta.actions.FileAction
 import org.apache.spark.sql.delta.DeltaLog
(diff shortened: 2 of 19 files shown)


core/src/main/scala/io/qbeast/core/model/MetadataManager.scala
```diff
@@ -8,8 +8,10 @@ import io.qbeast.IISeq
  *   type of data schema
  * @tparam FileDescriptor
  *   type of file descriptor
+ * @tparam QbeastOptions
+ *   type of the Qbeast options
  */
-trait MetadataManager[DataSchema, FileDescriptor] {
+trait MetadataManager[DataSchema, FileDescriptor, QbeastOptions] {
```
osopardo1, line 14: Should QbeastOptions be an interface in `core` instead of a type?

src/main/scala/io/qbeast/spark/delta/SparkDeltaMetadataManager.scala
```diff
@@ -15,16 +16,29 @@ import org.apache.spark.sql.SparkSession
...
 
     val deltaLog = loadDeltaQbeastLog(tableID).deltaLog
     val mode = if (append) SaveMode.Append else SaveMode.Overwrite
-    val options =
-      new DeltaOptions(Map("path" -> tableID.id), SparkSession.active.sessionState.conf)
-    val metadataWriter = DeltaMetadataWriter(tableID, mode, deltaLog, options, schema)
+    val conf = SparkSession.active.sessionState.conf
+    val deltaOptions = Map.newBuilder[String, String]
+    deltaOptions += "path" -> tableID.id
+    for (txnAppId <- options.txnAppId; txnVersion <- options.txnVersion) {
```
osopardo1, line 32: Could there be more than one `txnAppId` and `txnVersion` set in the options?

osopardo1:
BTW, I think we can also merge this feature on `main` right away, so we publish an official version 0.5.0 before 1.0.0.

src/main/scala/io/qbeast/spark/delta/SparkDeltaMetadataManager.scala
alexeiakimov: No, there cannot be more than one. I guess you are confused with `for` here. I borrowed this approach from Delta code where they handle the same options. The idea behind is that `for` allows to iterate over the Cartesian product of the specified generators. It is convenient because the syntax automatically unpack the elements, they are available under the variables declared tho the left of arrow. So the code in the body only works if and only if all the generators are not empty, and this is exactly what I need. An imperative analogue could be like the following:

```scala
if (options.txnAppId.notEmpty && options.txnVersion.notEmpty) {
  ...
}
```

Or if you like lambdas then

```scala
options.txnAppId.foreach(txnAppId => options.txnVersion.foreach(txnVersion => ...))
```

IMHO too many variants to my taste.

core/src/main/scala/io/qbeast/core/model/MetadataManager.scala
alexeiakimov: I just follow the approach chosen by @cugni. IMHO the core as a separate project with generic interfaces is overdesign here.

osopardo1:
> If we really need to have this functionality in the `main` branch, then I suggest to create a separate issue, because a simple cherry pick cannot help here, because the changes affect the files already modified in the `main-1.0.0`.

Okei, let's do it in a separate task. This can be merged on `main-1.0.0` in the meantime, and test it on user's environment.

core/src/main/scala/io/qbeast/core/model/MetadataManager.scala
cugni: QbeastOptions does not contain any specific Spark class, so it could be put directly in the core, with no need for generics. Indeed, DataSchema and FileDescriptor are generics because they are a specific class of Spark. The argument was that in the future, we might want to write in Qbeast without using Spark, which is something that now, with the Delta Kernels, is also possible and it might be useful. IMHO, the QbeastOptions relates only to Delta, nor to Spark, so they should be in core.

cugni (approved):

PR #251, opened 2023-12-15
Upgrade Apache Spark 3.5 and Delta Lake 3.0

osopardo1:
This PR is for upgrading to Spark 3.5 and Delta Lake 3.0. 

Here is what's changed:

### Spark 3.5

- Scala and Go Client to Support SparkConnect.
- Structured Streaming support for Spark Connect in Python and Scala.
- Introduce Arrow Python UDFs.
- Pandas API support for the Python Spark Connect Client. Easier management of Pandas DataFrames. 
- Avoid unnecessary task rerun on decommissioned executor lost if shuffle data migrated. 

[Here](https://spark.apache.org/releases/spark-release-3-5-0.html) are the full release notes.

### Delta 3.0
- **Delta Universal Format (UniForm)**. Now you can write on Delta and Iceberg within a single process.
- **Delta Kernel**. Set of Java Libraries that decouples the Log processing for both read (single and multi-threaded) and (TBD) write.
- Ability to specify which columns need to collect statistics via the table property `delta.dataSkippingStatsColumns`
- Support for Apache Spark 3.5.

[Here](https://github.com/delta-io/delta/releases/tag/v3.0.0) are the full release notes.

### Changes in Qbeast Spark
In terms of code, some APIs changed with the upgrade of versions.

- Spark `FileStatus` changed to `FileStatusWithMetadata`. Now the `FileStatus` returned to the Spark Catalyst from the underlying `FileIndex` can contain some extra Metadata to show in the table or to be used by Optimization rules. 

```scala

/**
 * A file status augmented with optional metadata, which tasks and file readers can use however they
 * see fit. For example, a custom [[FileIndex]] and [[FileFormat]] working together could expose
 * this extra metadata as file-constant fields of the file source metadata column.
 */

```
This can be useful for us, although we need further investigation. 

- Delta `withNewTransaction`. This method is used to commit information to the Delta Transaction Log. Before 3.0, it did not require any input parameter, but now it allows passing a `CatalogTable` or a `Snapshot` as an option for initializing the transaction. 
- Delta `StatisticsCollection`. This class constructs expressions that can be used to collect global and column-level statistics. The extended implementation includes methods like `columnMappingMode`, `outputTableStatsSchema` which is computed with the Table Schema, or `statsColumnSpec` which specifies the set of columns to be used for stats collection on a table. Keep in mind that if `delta.dataSkippingStatsColumns` is present, it takes precedence over `delta.dataSkippingNumIndexedCols` config.
- Spark replaces `LogicalPlan` for `QueryContext` when AnalysisException is thrown.


osopardo1:
Currently working on solving the issues when writing the Stats (max, min, nullCount of the columns for each file).

osopardo1:
The failing test was not finding the Delta File Stats on the first write. 

```scala

      val data = Seq(("A", 1), ("B", 2), ("C", 3)).toDF("a", "b")
      data.write
        .format("qbeast")
        .option("columnsToIndex", "a,b")
        .save(tmpDir)

      val stats =
        DeltaLog.forTable(spark, tmpDir).unsafeVolatileSnapshot.allFiles.collect().map(_.stats)
      stats.length shouldBe >(0)
      stats.head shouldBe "{\"numRecords\":3,\"minValues\":{\"a\":\"A\",\"b\":1}," +
        "\"maxValues\":{\"a\":\"C\",\"b\":3}," +
        "\"nullCount\":{\"a\":0,\"b\":0}}"

```

The problem was on `DeltaStatsCollectionUtils`, which **was initializing the `StatsCollector` from an empty schema** (since the Metadata of the Snapshot was empty because the was the first write of a Table). 

Hope a small fix can solve the issue.

src/main/scala/org/apache/spark/sql/delta/DeltaStatsCollectionUtils.scala
```diff
@@ -19,50 +22,115 @@ import org.apache.spark.sql.SparkSession
...
+      snapshot: Snapshot,
+      outputSchema: StructType,
+      dataFrameOutput: Seq[Attribute],
+      partitionSchema: StructType): (Seq[Attribute], Seq[Attribute]) = {
+
+    // Column mapping mode
+    val columnMappingMode = snapshot.columnMappingMode
+    // Stats Schema: if the table is not initialized, we use the output schema to collect stats;
+    // otherwise, we use the table schema to collect stats
+    val statsSchema =
```
osopardo1, line 44: The failing test was not finding the Delta File Stats on the first write. 

```scala

      val data = Seq(("A", 1), ("B", 2), ("C", 3)).toDF("a", "b")
      data.write
        .format("qbeast")
        .option("columnsToIndex", "a,b")
        .save(tmpDir)

      val stats =
        DeltaLog.forTable(spark, tmpDir).unsafeVolatileSnapshot.allFiles.collect().map(_.stats)
      stats.length shouldBe >(0)
      stats.head shouldBe "{\"numRecords\":3,\"minValues\":{\"a\":\"A\",\"b\":1}," +
        "\"maxValues\":{\"a\":\"C\",\"b\":3}," +
        "\"nullCount\":{\"a\":0,\"b\":0}}"

```

The problem was on `DeltaStatsCollectionUtils`, which **was initializing the `StatsCollector` from an empty schema** (since the Metadata of the Snapshot was empty because it was the first write of a Table). 

Hope this small fix can solve the issue.

osopardo1:
See some Codecov errors with DeltaStatsCollectionUtils. I will add unit testing for proper checks.

alexeiakimov (approved):

PR #254, opened 2024-01-12
253 add file skipping with delta

alexeiakimov:
## Description

This is a rework of the query implementation. This PR uses internal Delta query engine always
 except the queries with sampling clause. For the later the Qbeast engine is used.

## Type of change

This is an internal change, neither API nor data format are affected.

## Checklist:

- [x] Add tests.

DefaultFileFormatTest is added to test the new logic in the query engine


src/main/scala/io/qbeast/spark/delta/DefaultFileIndex.scala
```diff
@@ -0,0 +1,86 @@
...
+import org.apache.spark.sql.types.StructType
+import org.apache.spark.sql.SparkSession
+
+/**
+ * Default implementation of the FileIndex.
+ *
+ * @param target
+ *   the target file index implemented by Delta
+ */
+class DefaultFileIndex private (target: TahoeLogFileIndex)
```
osopardo1, line 24: Should we include Qbeast on the name of the class? Like "QbeastDefaultFileIndex"?

src/main/scala/io/qbeast/spark/delta/DefaultFileIndex.scala
```diff
@@ -0,0 +1,86 @@
...
+    logFilters(partitionFilters, dataFilters)
+    val strategy = if (haveQbeastWeightExpression(dataFilters)) {
+      SamplingListFilesStrategy
+    } else {
+      DefaultListFilesStrategy
+    }
+    strategy.listFiles(target, partitionFilters, dataFilters)
+  }
+
+  private def logFilters(
```
osopardo1, line 44: Instead of having a separate method in the class, wouldn't be better to implement an object for Logging where we can also add other types? (Like logging for writing, optimizing....)

I know it might be out of the scope of this PR, but if it's necessary we can open an issue and another PR for that.

src/main/scala/io/qbeast/spark/delta/SamplingListFilesStrategy.scala
```diff
@@ -0,0 +1,100 @@
...
+          file.size,
+          false,
+          0,
+          1,
+          file.modificationTime,
+          getAbsolutePath(target, file)))
+      .map(file => new FileStatusWithMetadata(file, Map.empty))
+  }
+
+  private def isStagingAreaFile(file: AddFile): Boolean = file.getTag(TagUtils.revision) match {
```
osopardo1, line 59: I think there is some trait or object called `StagingUtils` where this method can be more useful.

src/main/scala/io/qbeast/spark/delta/SamplingListFilesStrategy.scala
```diff
@@ -0,0 +1,100 @@
...
+      dataFilters: Seq[Expression]): Seq[PartitionDirectory] = {
+    val snapshot = target.getSnapshot
+    val path = target.path
+    val files = listStagingAreaFiles(target, partitionFilters, dataFilters) ++
+      listIndexFiles(snapshot, path, partitionFilters, dataFilters)
+    logFilteredFiles(target, snapshot, files)
+    Seq(PartitionDirectory(new GenericInternalRow(Array.empty[Any]), files))
+  }
+
+  private def listStagingAreaFiles(
```
osopardo1, line 41: Shouldn't this method be common for `ListFilesStrategy`?

alexeiakimov: No, it should not, it only selects the staging area files, it cannot be reused in DefaultListFileStrategy, because there I just use Delta for all the files (it is assumed that there is no replication)

src/main/scala/io/qbeast/spark/delta/DefaultFileIndex.scala
alexeiakimov: IMHO it is too early, do you know other pieces of code which need such an utility object? Anyway feel free to open a separate issue.

src/main/scala/io/qbeast/spark/delta/DefaultFileIndex.scala
alexeiakimov: I do not think so, here everything is Qbeast, adding Qbeast does not add any new information to what the abstraction is dedicated to. IMHO adding Qbeast only makes sense for public classes to be used by the end-user, because for the end- user is important to distinguish them from similar concepts from other libraries. For internal API we can use a shorter name.

src/main/scala/io/qbeast/spark/delta/SamplingListFilesStrategy.scala
alexeiakimov: There are two traits StagingUtils in the qbeast-core, and DeltaStagingUtils in qbeast-spark. The first one knows nothing about AddFile, and the second one has method I need ´isStagingFile´ however implementing DeltaStagingUtils in SamplingListFileStrategy is hard, because it requires to implement ´snapshot´ method returning the Delta Snapshot, but the design of ListFileStrategy is that there is no state, the snapshot is computed on the stack. Probably we need to redesign DeltaStagingUtils first.

src/main/scala/io/qbeast/spark/delta/DefaultFileIndex.scala
osopardo1: I will open an issue with some other processes that might need Logging information, and which type of information could be good to show to the developer. Thanks.

osopardo1:
I will approve this PR, but before merging I am waiting to understand if benchmarking the solution is necessary.

fpj:
I gave a shot at this PR. I tested it locally with the NYC Yellow Cab trip data:

https://www.nyc.gov/site/tlc/about/tlc-trip-record-data.page

I have fetched the parquet files from 2022 and 2023, and ingested them as a qbeast table. There are changes to the schema across the files, so I ended up executing the following to ingest the data:

```
def readParquetFile(filePath: String): DataFrame = {
        spark.read.parquet(filePath)
     }

parquetFiles.foreach { parquetFile =>
        val df = readParquetFile(parquetFile)
        val dfCasted = df.withColumn("VendorID", df("VendorID").cast(LongType)).
                                     withColumn("passenger_count", df("passenger_count").cast(LongType)).
                                     withColumn("RatecodeID", df("RatecodeID").cast(LongType)).
                                     withColumn("PULocationID", df("PULocationID").cast(LongType)).
                                     withColumn("DOLocationID", df("DOLocationID").cast(LongType))
        dfCasted.write.format("qbeast").mode("append").
                                  option("columnsToIndex", "fare_amount,trip_distance,passenger_count").
                                  option("cubeSize", "50000").
                                  save("/path/nyc-qbeast-table")
      }
```

I executed the following test query without the changes in this PR:

```
"SELECT count(*) FROM table WHERE (fare_amount > 100) AND (passenger_count = 3)"
```

and I got bad numbers:

```
scala> spark.time(nyc_qbeast_query.show)
+--------+
|count(1)|
+--------+
|   12282|
+--------+

Time taken: 20671 ms

scala> spark.time(nyc_qbeast_query.show)
+--------+
|count(1)|
+--------+
|   12282|
+--------+

Time taken: 19528 ms
```

with the changes in this PR, the execution time becomes a lot better:

```
scala> spark.time(nyc_qbeast_query.show)
+--------+                                                                      
|count(1)|
+--------+
|   12282|
+--------+

Time taken: 1283 ms

scala> spark.time(nyc_qbeast_query.show)
+--------+
|count(1)|
+--------+
|   12282|
+--------+

Time taken: 1158 ms
```

osopardo1:
Great test @fpj !!! Thanks for giving it a try 💯 
I would suggest to also output some performance numbers for the sampling operation. Something like:

`SELECT count(*) FROM table TABLESAMPLE(1 PERCENT)`

fpj:
I'm not seeing any significant difference when sampling between the execution time without the changes here and with the changes. Here is what I'm seeing:

```
// Query: "SELECT count(*) FROM nyc_qbeast_table TABLESAMPLE(1 PERCENT)"

//
// Runs without the changes in this PR
//
scala> spark.time(nyc_qbeast_query.show)
+--------+                                                                      
|count(1)|
+--------+
|  830160|
+--------+

Time taken: 24262 ms

scala> spark.time(nyc_qbeast_query.show)
+--------+                                                                      
|count(1)|
+--------+
|  830160|
+--------+

Time taken: 10625 ms

scala> spark.time(nyc_qbeast_query.show)
+--------+
|count(1)|
+--------+
|  830160|
+--------+

Time taken: 8200 ms

scala> spark.time(nyc_qbeast_query.show)
+--------+
|count(1)|
+--------+
|  830160|
+--------+

Time taken: 9059 ms

//
// Runs with the changes in this PR
//

scala> spark.time(nyc_qbeast_query.show)
+--------+                                                                      
|count(1)|
+--------+
|  830160|
+--------+

Time taken: 25136 ms

scala> spark.time(nyc_qbeast_query.show)
+--------+                                                                      
|count(1)|
+--------+
|  830160|
+--------+

Time taken: 10750 ms

scala> spark.time(nyc_qbeast_query.show)
+--------+
|count(1)|
+--------+
|  830160|
+--------+

Time taken: 8774 ms

scala> spark.time(nyc_qbeast_query.show)
+--------+                                                                      
|count(1)|
+--------+
|  830160|
+--------+

Time taken: 8725 ms

```

@osopardo1 Is this behavior expected?

osopardo1:
Yes @fpj , the behaviour is expected since the filters of sampling should be applied in the same way as before. 

Now, if this PR has shown improvement in WHERE clauses and no damage at SAMPLING, I think it's ready for merge 👍

osopardo1 (approved):

PR #265, opened 2024-02-05
cubeStatuses should be a SortedMap

Jiaweihu08:
Change IndexMetrics.cubeStatuses type from `Map` to `SortedMap`.


osopardo1:
Does it affect the sorting or it just affects how the elements are searched through the Map?

osopardo1 (approved):

PR #277, opened 2024-02-26
Upgrade delta lake version

### LINKED ISSUE #275: Upgrade to Delta 3.1.0
Within the past weeks, Delta Lake released its version 3.1.0 with some exciting opportunities for integrations with the ecosystem. 

Some of the highlights are:

- Preview of [Liquid Clustering](https://github.com/delta-io/delta/issues/1874). Now Delta allows clustering the data in a Delta table for better data skipping.
- [Auto-Compaction](https://github.com/delta-io/delta/pull/2414) to address the small files problem during table writes.
- [Support for optimizing min/max aggregations](https://github.com/delta-io/delta/issues/2092) using the table metadata
- [Fixed](https://github.com/delta-io/delta/commit/61dd5d16) some issues with column renames in Merge. 

Find the full notes [here](https://github.com/delta-io/delta/releases).

We should:

- [ ] Check which APIs have changed and if they affect our code. 
- [ ] Agreed on which version of Qbeast-Spark should we add this change. The initial idea is to upgrade it on top of 1.0.0, but then there will be no version for Delta 3.0.0 (unless we publish within a different version number). We agree to constantly support the latest versions, and for the size of the project, it's ok.
- [ ] Upgrade Delta Library on the branch. 
- [ ] Update the APIs.
- [ ] Run tests
- [ ] Publish SNAPSHOT with the last commit ID.
- [ ] Update other deployments that might contain the Qbeast and Delta Lake images.

Jiaweihu08:
Fixes #275

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -6,7 +6,7 @@ import sbt._
 object Dependencies {
   lazy val sparkVersion: String = sys.props.get("spark.version").getOrElse("3.5.0")
   lazy val hadoopVersion: String = sys.props.get("hadoop.version").getOrElse("3.3.4")
-  lazy val deltaVersion: String = "3.0.0"
+  lazy val deltaVersion: String = "3.1.0"
 
   val sparkCore = "org.apache.spark" %% "spark-core" % sparkVersion
   val sparkSql = "org.apache.spark" %% "spark-sql" % sparkVersion


osopardo1 (commented):
all gooooooood

osopardo1 (approved):

PR #283, opened 2024-03-13
Road to 0.6.0!

### LINKED ISSUE #226: Appends that don't update cube weights
## What went wrong?
The current **weight merge** can push append records to levels beyond estimation, and there are scenarios where new **blocks** are added to cubes without updating their `maxWeights`.

When the existing index is relatively large in comparison to that of the appending data, the estimated cube weights wil reduce significantly because of the merge. This pushes records down to the subtrees, and can reach levels that are not present in the estimation index.

For example, say that `c1` is a leaf cube with a `NormalizedWeight` of 2.0 for the appending data. After merging its weight is reduced to say 0.8. Some records from `c1`'s domain, specifically those with weights between 0.8 and 1.0, will be distributed among its descendants in the following fashin:
- If `c1` in the existing index is also a leaf cube, child cubes will be created for the allocation of these records, creating one extra level at max.
- If the **existing index** has a subtree for `c1`, however, these records will be indexed accordingly but the blocks will be added without updating the index, for the weight merge is already done.

During an experiment with 30 appends, **~45%** of the append blocks experience this issue.

### 2. Branch and commit id:
`main`, [f066acf](https://github.com/Qbeast-io/qbeast-spark/commit/f066acfa640e795459bb9ff815b641377660bd43)

### 3. Spark version:
`3.4.1`

### 4. Hadoop version:
`3.3.4`

`locally`

osopardo1:
We published release 0.5.0 last January 9th. The last version included changes such as String Indexing and Updates to Spark and Delta 2.4.0. 

Although those changes **significantly improved the distribution of text columns**, we wanted to work on algorithm optimizations that could balance the file layout for all kinds of data.

# What's changed?

## Roll-Up

One of the key operations for distributing the files evenly is the Roll-Up.

Roll-Up compaction solves the small files problem by **storing the blocks of closely related cubes in a single physical file**. Here “closely related” means that there is a high probability that the cubes contribute data to the same query. 

image

## New Protocol
We've been working on upbreaking changes in the algorithm, **which affects the Metadata written in the Commit Log**.

In summary, instead of having one single file containing one single cube, now we allow **one file to contain multiple cubes stored in `blocks`**. 

image

#### Original protocol metadata:

```json
"tags": {
  "state": "FLOODED",
  "cube": "w",
  "revision": "1",
  "minWeight": "2",
  "maxWeight": "3",
  "elementCount": "4" 
}


```

#### NEW protocol metadata:

```json
"tags": {
  "revision": "1",
  "blocks": [
    {
      "cube": "w",
      "minWeight": 2,
      "maxWeight": 3,
      "replicated": false,
      "elementCount": 4
    },
    {
      "cube": "wg",
      "minWeight": 5,
      "maxWeight": 6,
      "replicated": false,
      "elementCount": 7
    },
  ]
}


```

## Domain-Driven Appends

Another of the upgrades we made in the new code, is using Cube Domains for appending data incrementally. The change uses the existing index during partition-level domain estimation to help reducing the number of cubes with **outdated max weights from 45% to 0.16%**. Or in other words, more balanced files!

Fixes #226. Full details in #227 

## Auto-Indexing

Sometimes the `.option("columnsToIndex", "a,b")` was too hot to handle... That's why we add functionality to **automatically choose the best columns to organize the data**. 

The feature is not enabled by default. If you want to use it, you should add the necessary configuration. 

```shell
spark.qbeast.index.columnsToIndex.auto=true
spark.qbeast.index.columnsToIndex.auto.max=10
```


osopardo1:
Uep, closing this PR because it should be merged the changed to main, not to 1.0.0 to resolve conflicts. Sorry!!

PR #284, opened 2024-03-13
Issue 292: Merge main-1.0.0 into main

osopardo1:
We published release 0.5.0 last January 9th. The last version included changes such as String Indexing and Updates to Spark and Delta 2.4.0. 

Although those changes **significantly improved the distribution of text columns**, we wanted to work on algorithm optimizations that could balance the file layout for all kinds of data.

Closes #292 

# What's changed?

## Roll-Up

One of the key operations for distributing the files evenly is the Roll-Up.

Roll-Up compaction solves the small files problem by **storing the blocks of closely related cubes in a single physical file**. Here “closely related” means that there is a high probability that the cubes contribute data to the same query. 

image

## New Protocol
We've been working on upbreaking changes in the algorithm, **which affects the Metadata written in the Commit Log**.

In summary, instead of having one single file containing one single cube, now we allow **one file to contain multiple cubes stored in `blocks`**. 

image

#### Original protocol metadata:

```json
"tags": {
  "state": "FLOODED",
  "cube": "w",
  "revision": "1",
  "minWeight": "2",
  "maxWeight": "3",
  "elementCount": "4" 
}


```

#### NEW protocol metadata:

```json
"tags": {
  "revision": "1",
  "blocks": [
    {
      "cube": "w",
      "minWeight": 2,
      "maxWeight": 3,
      "replicated": false,
      "elementCount": 4
    },
    {
      "cube": "wg",
      "minWeight": 5,
      "maxWeight": 6,
      "replicated": false,
      "elementCount": 7
    },
  ]
}


```

## Domain-Driven Appends

Another of the upgrades we made in the new code, is using Cube Domains for appending data incrementally. The change uses the existing index during partition-level domain estimation to help reducing the number of cubes with **outdated max weights from 45% to 0.16%**. Or in other words, more balanced files!

Fixes #226. Full details in #227 

## Auto-Indexing

Sometimes the `.option("columnsToIndex", "a,b")` was too hot to handle... That's why we add functionality to **automatically choose the best columns to organize the data**. 

The feature is not enabled by default. If you want to use it, you should add the necessary configuration. 

```shell
spark.qbeast.index.columnsToIndex.auto=true
spark.qbeast.index.columnsToIndex.auto.max=10
```

PR #286, opened 2024-03-19
Issue 281 - Schema Merge

### LINKED ISSUE #281: Appending data should not merge schemas at rename
## What went wrong?

The expected behavior is that:

- if I create a table with 5 columns, and then I try to append to it a dataset with 6 columns, it should fails. Delta fails, but Qbeast doesn't and merge the schema.
- if I create a table with 5 columns, and then we append a dataset that contains only part of them (let's say the first 4), the operation should succeed, and the last column is full of null. In this, both Delta and Qbeast behave correctly.
- if I create a a table with 5 columns, and then I try to append a dataset with 5 columns, but one of them it is renamed, it should fail. Here Qbeast fails to raise an exception and instead, merges the schema.

## How to reproduce?

```scala

val df = Seq((2073090619, 773622413, "PRT76gRvYP", "{8RxqrEd9TY}")).toDF("a", "b", "field_with_forbidden_characters_", "another_field_with_forbidden_characters")
df.write.format("qbeast").option("columnsToIndex", "a,b").save("/tmp/dir")

val renamed_df = df.withColumn("rename_field", col("another_field_with_forbidden_characters")
renamed_df.write.format("qbeast").mode("append").save("/tmp/dir")


```

The resulting schema is

`| a         | b         | field_with_forbidden_characters_ | another_field_with_forbidden_characters | renamed_field |
`

### 2. Branch and commit id:

`1.0.0-main` 

### 3. Spark version:

`3.5.0`

### 4. Hadoop version:

Locally

### 6. Stack trace:

osopardo1:
## Description

Fixes #281 

## Type of change

Bug fix. No breaking changes. Change only qbeast-spark package.

## Checklist:

- [x] Add tests.

Added test to `QbeastSchemaTest`.

**Test Configuration**:
* Spark Version: 3.5.0
* Hadoop Version: 3.3.3
* Cluster or local? Local

### DIFF
--- a/src/main/scala/io/qbeast/spark/delta/DeltaMetadataWriter.scala
+++ b/src/main/scala/io/qbeast/spark/delta/DeltaMetadataWriter.scala
@@ -54,6 +54,11 @@ private[delta] case class DeltaMetadataWriter(
 
   private def isOverwriteOperation: Boolean = mode == SaveMode.Overwrite
 
+  override protected val canMergeSchema: Boolean = options.canMergeSchema
+
+  override protected val canOverwriteSchema: Boolean =
+    options.canOverwriteSchema && isOverwriteOperation && options.replaceWhere.isEmpty
+
   private val sparkSession = SparkSession.active
 
   private val deltaOperation = {
--- a/src/main/scala/io/qbeast/spark/delta/QbeastMetadataOperation.scala
+++ b/src/main/scala/io/qbeast/spark/delta/QbeastMetadataOperation.scala
@@ -104,17 +104,14 @@ private[delta] trait QbeastMetadataOperation extends ImplicitMetadataOperation w
     val dataSchema = StructType(schema.fields.map(field =>
       field.copy(nullable = true, dataType = asNullable(field.dataType))))
 
-    println("SCHEMA" + dataSchema.json)
     val mergedSchema =
       if (isOverwriteMode && canOverwriteSchema) dataSchema
       else SchemaMergingUtils.mergeSchemas(txn.metadata.schema, dataSchema)
-    println("MERGED SCHEMA" + mergedSchema)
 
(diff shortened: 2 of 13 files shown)


osopardo1:
Different parts of the code were affected by the Schema Evolution and Schema Enforcement features:

- `QbeastMetadataOperation` parameter initialization of `canMergeSchema` and `canOverwriteSchema`.
- `QbeastCatalogUtils` creates a Table if:
    - No data has been committed previously. If there's no initial log, we would need to create one with the schema specified. A quick fix for that is to call CREATE DELTA TABLE COMMAND + CONVERT TO QBEAST.
    - Data would be written in the operation. The first commit would be done as usual when using SAVE TABLE AS (SELECT * FROM...). 
    - Data is written already. In this case, we would normally read the properties and the schema from the existing log, no need to add extra info. 
- User Options that contain modifications to the `Metadata`, such as `mergeSchema` are not passed to the `QbeastMetadataOperation`.

osopardo1:
Changes on `QbeastCatalogUtils` seem to broke the creation of new tables somehow 🐒 

Need some extra time to find the cause and solve it.

osopardo1:
> Changes on `QbeastCatalogUtils` seem to broke the creation of new tables somehow 🐒
> 
> I think it might be because `CreateDeltaTableCommand` creates the table in a different SessionCatalog than the one that Qbeast manages. Need some extra time to find the cause and solve it.

It was something to do with global SparkSession initialization for the object QbeastCatalogUtils.... solved 👌

osopardo1:
After merging #284 we will add this feature to `main`.

osopardo1:
⛔  Need to rebase to `main` and change the base branch!

osopardo1:
Cherry-picked those changes in a new branch. I'm going to open a new PR!

PR #298, opened 2024-03-28
Issue 297: Reduce overhead for CubeDomainsBuilder instantiation

### LINKED ISSUE #297: Overhead during partition cube domain estimation
## What went wrong?
Sending `IndexStatus` to each partition for the creation of a `CubeDomainsBuilder` instance incurs a large amount of overhead.

`IndexStatus` has a list of `Block`s for each associated `CubeId` in `IndexStatus.cubeStatuses`, and the number of `Block`s grows with the number of appends. Meanwhile, this information is unnecessary during domain computation and should be avoided.

Branch: `main`, `main-1.0.0`

Spark version: `3.5.0`

Hadoop version: `3.3.4`

How are you running Spark? `locally`

Jiaweihu08:
This PR Fixes #297 by

1. As the overhead comes from `CubeStatus`, we pass the required objects directly instead of `IndexStatus`
2. Use `Broadcast` so they persist in each worker so we don't have to pass them for each task

### DIFF
--- a/core/src/main/scala/io/qbeast/core/model/CubeDomainsBuilder.scala
+++ b/core/src/main/scala/io/qbeast/core/model/CubeDomainsBuilder.scala
@@ -47,47 +47,17 @@ object CubeDomainsBuilder {
     Math.max(minGroupCubeSize, groupCubeSize.toInt)
   }
 
-  /**
-   * Compute cube NormalizedWeights to properly reflect leaf cube elementCounts.
-   */
-  private def computeNormalizedWeights(
-      desiredCubeSize: Int,
-      groupCubeSize: Int,
-      cubeStatuses: Map[CubeId, CubeStatus]): Map[CubeId, NormalizedWeight] = {
-    val isLeaf = (cubeId: CubeId) => !cubeId.children.exists(cubeStatuses.contains)
-    val multiplier = groupCubeSize / desiredCubeSize.toDouble
-
-    cubeStatuses.map {
-      case (cubeId, status) if isLeaf(cubeId) =>
-        val cubeSize = status.blocks.map(_.elementCount).sum
-        val groupSize = (cubeSize * multiplier).toLong.max(1L)
-        val nw = NormalizedWeight(groupCubeSize, groupSize).max(1.0)
-        cubeId -> nw
-      case (cubeId, status) => cubeId -> status.normalizedWeight
-    }
-  }
-
   def apply(
-      indexStatus: IndexStatus,
+      existingCubeWeights: Map[CubeId, Weight],
+      replicatedOrAnnouncedSet: Set[CubeId],
+      desiredCubeSize: Int,
       numPartitions: Int,
       numElements: Long,
       bufferCapacity: Long): CubeDomainsBuilder = {
-    val desiredCubeSize = indexStatus.revision.desiredCubeSize
-    val groupCubeSize = estimateGroupCubeSize(
-      indexStatus.revision.desiredCubeSize,
-      numPartitions,
-      numElements,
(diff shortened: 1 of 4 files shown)


src/main/scala/io/qbeast/spark/index/OTreeDataAnalyzer.scala
```diff
@@ -145,13 +145,21 @@ object DoublePassOTreeDataAnalyzer extends OTreeDataAnalyzer with Serializable {
       val numPartitions: Int = weightedDataFrame.rdd.getNumPartitions
       val bufferCapacity: Long = CUBE_WEIGHTS_BUFFER_CAPACITY
 
+      // Broadcast large objects for CubeDomainsBuilder
+      val broadcastExistingCubeWeights = SparkSession.active.sparkContext.broadcast(
```
osopardo1, line 149: I would use the same SparkSession.active for both broadcasts

core/src/main/scala/io/qbeast/core/model/CubeDomainsBuilder.scala
```diff
@@ -235,22 +204,18 @@ class CubeDomainsBuilder protected (
...
- * @param existingNormalizedWeights
+ * @param existingCubeWeights
  *   Cube Weight of the existing index
- * @param groupCubeSize
- *   partition-level cube size for indexing
  */
-private class WeightAndCountFactory(
-    existingNormalizedWeights: Map[CubeId, NormalizedWeight],
-    groupCubeSize: Int) {
+private class WeightAndCountFactory(existingCubeWeights: Map[CubeId, Weight]) {
```
osopardo1, line 210: Why Weight instead of NormalizedWeight? Could this affect calculations?

core/src/main/scala/io/qbeast/core/model/CubeDomainsBuilder.scala
```diff
@@ -235,22 +204,18 @@ class CubeDomainsBuilder protected (
...
 
   def create(cubeId: CubeId): WeightAndCount =
-    existingNormalizedWeights.get(cubeId) match {
-      case Some(nw: NormalizedWeight) if nw < 1.0 =>
-        new InnerCubeWeightAndCount(Weight(nw))
-      case Some(nw: NormalizedWeight) =>
-        val existingLeafSize = ((groupCubeSize - 1) / nw).toInt
-        new LeafCubeWeightAndCount(existingLeafSize)
+    existingCubeWeights.get(cubeId) match {
```
osopardo1, line 213: Adding some comments to the code would be great 🙏

osopardo1 (commented):
Some comments!

core/src/main/scala/io/qbeast/core/model/CubeDomainsBuilder.scala
Jiaweihu08: Instead of computing `NormalizedWeight` and then using it to compute the `elementCount` for each existing leaf cube, we assume leaf cubes are empty.

This would place more elements in these leaf cubes during indexing. Still, as we are only interested in cube domains, this should be fine with the calculations.

osopardo1: Let's test that this change does not compromise index layout with bigger workloads before introducing it 🙏

osopardo1:
Seems that the change in the code is not affecting the structure of the index in larger appends. 

When @fpj adds his review, I will wrap up and prepare the merge.

fpj (approved):
+1, lgtm

osopardo1 (approved):

PR #301, opened 2024-04-02
Issue 281: Schema Merge

### LINKED ISSUE #281: Appending data should not merge schemas at rename

osopardo1:
## Description

Fixes #281 . Schema changes in the data where merged by default when appending it to an existing table.

Two main problems are solved in the PR:

1. The `QbeastMetadataOperation` is no longer setting up `canMergeSchema` and `canOverwriteSchema` parameters to `true` by default. Now they are initialized according to the properties specified in `DeltaOptions` / `QbeastOptions`.
3. When Creating a Table, the behavior of Delta is to do a **first commit in the path with only the Metadata** (schema, table id...). In the Qbeast case, no initial schema was committed. This is solved by introducing a **first write with an empty data frame and a Convert To Qbeast operation**.

## Type of change

Bug fix. No breaking changes. Change only qbeast-spark package.

## Checklist:

- [x] Add tests.

Added test to `QbeastSchemaTest`.

**Test Configuration**:
* Spark Version: 3.5.0
* Hadoop Version: 3.3.3
* Cluster or local? Local

### DIFF
--- a/src/main/scala/io/qbeast/spark/delta/DeltaMetadataWriter.scala
+++ b/src/main/scala/io/qbeast/spark/delta/DeltaMetadataWriter.scala
@@ -66,6 +66,11 @@ private[delta] case class DeltaMetadataWriter(
 
   private def isOverwriteOperation: Boolean = mode == SaveMode.Overwrite
 
+  override protected val canMergeSchema: Boolean = options.canMergeSchema
+
+  override protected val canOverwriteSchema: Boolean =
+    options.canOverwriteSchema && isOverwriteOperation && options.replaceWhere.isEmpty
+
   private val sparkSession = SparkSession.active
 
   private val deltaOperation = {
--- a/src/main/scala/io/qbeast/spark/internal/QbeastOptions.scala
+++ b/src/main/scala/io/qbeast/spark/internal/QbeastOptions.scala
@@ -38,6 +38,10 @@ import org.apache.spark.sql.SparkSession
  *   the application identifier
  * @param userMetadata
  *   user-provided metadata for each CommitInfo
+ * @param mergeSchema
+ *   whether to merge the schema
+ * @param overwriteSchema
+ *   whether to overwrite the schema
  */
 case class QbeastOptions(
     columnsToIndex: Seq[String],
@@ -45,7 +49,34 @@ case class QbeastOptions(
     stats: Option[DataFrame],
     txnAppId: Option[String],
     txnVersion: Option[String],
-    userMetadata: Option[String])
+    userMetadata: Option[String],
+    mergeSchema: Option[String],
+    overwriteSchema: Option[String]) {
+
+  def toDeltaOptions(): DeltaOptions = {
+
+    val spark = SparkSession.active
+    val conf = spark.sessionState.conf
+    val deltaOptions = Map.newBuilder[String, String]
+    for (txnAppId <- txnAppId; txnVersion <- txnVersion) {
+      deltaOptions += DeltaOptions.TXN_APP_ID -> txnAppId
+      deltaOptions += DeltaOptions.TXN_VERSION -> txnVersion
+    }
+    if (userMetadata.nonEmpty) {
+      deltaOptions += DeltaOptions.USER_METADATA_OPTION -> userMetadata.get
+    }
+    if (mergeSchema.nonEmpty) {
+      deltaOptions += DeltaOptions.MERGE_SCHEMA_OPTION -> mergeSchema.get
+    }
+    if (overwriteSchema.nonEmpty) {
+      deltaOptions += DeltaOptions.OVERWRITE_SCHEMA_OPTION -> overwriteSchema.get
+    }
+
+    new DeltaOptions(deltaOptions.result(), conf)
+
+  }
+
+}
 
 /**
  * Options available when trying to write in qbeast format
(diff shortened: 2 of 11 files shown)


src/test/scala/io/qbeast/spark/utils/QbeastSchemaTest.scala
```diff
@@ -131,4 +131,90 @@ class QbeastSchemaTest extends QbeastIntegrationTestSpec {
...
+
+      import spark.implicits._
+
+      spark.sql(
+        "CREATE TABLE student (id INT) USING qbeast " +
+          "OPTIONS ('columnsToIndex'='id')")
+
+      val dfExtraCol = Seq((1, "John"), (2, "Doe")).toDF("id", "name")
+
+      // EXRA COLUMN
```
Jiaweihu08, line 154: EXTRA

src/test/scala/io/qbeast/spark/utils/QbeastSchemaTest.scala
```diff
@@ -131,4 +131,90 @@ class QbeastSchemaTest extends QbeastIntegrationTestSpec {
...
+      val path = s"$tmpDir/student"
+      df.write
+        .format("qbeast")
+        .mode("overwrite")
+        .option("columnsToIndex", "id")
+        .save(path)
+
+      val dfExtraCol = Seq((1, "John"), (2, "Doe")).toDF("id", "name")
+
+      // EXRA COLUMN
```
Jiaweihu08, line 188: EXTRA

Jiaweihu08 (approved):

osopardo1:
Fixing the comments and adding coverage to `QbeastOptions` case class.

src/main/scala/io/qbeast/spark/internal/QbeastOptions.scala
```diff
@@ -38,14 +40,44 @@ import org.apache.spark.sql.SparkSession
...
+      options += DeltaOptions.TXN_VERSION -> txnVersion
+    }
+    if (userMetadata.nonEmpty) {
+      options += DeltaOptions.USER_METADATA_OPTION -> userMetadata.get
+    }
+    if (mergeSchema.nonEmpty) {
+      options += DeltaOptions.MERGE_SCHEMA_OPTION -> mergeSchema.get
+    }
+    if (overwriteSchema.nonEmpty) {
+      options += DeltaOptions.OVERWRITE_SCHEMA_OPTION -> overwriteSchema.get
```
fpj, line 71: It sounds like codecov is not happy about this line 71. Where do we exercise this `toMap` method in the tests?

src/main/scala/io/qbeast/spark/internal/QbeastOptions.scala
```diff
@@ -135,14 +175,24 @@ object QbeastOptions {
     val txnAppId = getTxnAppId(options)
     val txnVersion = getTxnVersion(options)
     val userMetadata = getUserMetadata(options)
-    QbeastOptions(columnsToIndex, desiredCubeSize, stats, txnAppId, txnVersion, userMetadata)
+    val mergeSchema = getMergeSchema(options)
+    val overwriteSchema = getOverwriteSchema(options)
+    QbeastOptions(
```
fpj, line 180: Perhaps not for this PR, but I'm wondering whether we should implement QbeastOptions following the builder pattern. It makes the creation of the object easier to maintain and visualize.

fpj (changes requested):
Looks good, I only have a few questions.

src/main/scala/io/qbeast/spark/delta/DeltaMetadataWriter.scala
```diff
@@ -66,6 +66,11 @@ private[delta] case class DeltaMetadataWriter(
 
   private def isOverwriteOperation: Boolean = mode == SaveMode.Overwrite
 
+  override protected val canMergeSchema: Boolean = options.canMergeSchema
+
+  override protected val canOverwriteSchema: Boolean =
+    options.canOverwriteSchema && isOverwriteOperation && options.replaceWhere.isEmpty
```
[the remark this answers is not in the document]
osopardo1: This is part of Delta's code, but basically the option `canOverwriteSchema` is just an argument that allows to overwrite it, but unless the other requirements are met (which needs to be explicit), the enforcement of the schema is applied.

src/main/scala/io/qbeast/spark/internal/QbeastOptions.scala
osopardo1: Fixed on bd53be0 !

src/main/scala/io/qbeast/spark/internal/QbeastOptions.scala
osopardo1: Makes sense. Let me open an issue.

src/main/scala/io/qbeast/spark/delta/DeltaMetadataWriter.scala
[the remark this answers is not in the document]
fpj: Thank you, @osopardo1

fpj (approved):
+1, lgtm

PR #304, opened 2024-04-05
Issue 274: Fix erroneous idempotent writes

### LINKED ISSUE #274: Unexpected idempotent write behavior
The [delta lake protocol](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#transaction-identifiers) states that for the same `txnAppId,` any `txnVersion` **smaller than or equal to** the latest one in the log is to be ignored. At the moment, we are checking for equality only.

In `snapshot.setTransactions`, only the largest `txnVersion` is saved.

```
Map(testApp1 -> 30, testApp2 -> 10)
```

- Branches: `main`, `main-1.0.0`

Jiaweihu08:
## Description

Fixes #274

## Type of change

**Bug fix**

## Checklist:

- [x] Add tests.

Modified existing tests to test idempotent writes according to the [delta protocol](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#transaction-identifiers).

**Test Configuration**:
* Spark Version: `3.5.0`
* Hadoop Version: `3.3.4`
* Cluster or local? `local`


osopardo1:
LGTM 👍

osopardo1 (approved):

osopardo1:
After this squash and merge, I will prepare the main for a 0.6.0-SNAPSHOT release.

src/test/scala/io/qbeast/spark/utils/QbeastSparkTxnTest.scala
```diff
@@ -91,14 +97,20 @@ class QbeastSparkTxnTest extends QbeastIntegrationTestSpec {
         .option(QbeastOptions.COLUMNS_TO_INDEX, "id")
         .option(QbeastOptions.CUBE_SIZE, 1)
         .option(QbeastOptions.TXN_APP_ID, "test")
-        .option(QbeastOptions.TXN_VERSION, "1")
+        .option(QbeastOptions.TXN_VERSION, "2")
```
fpj, line 100: A quick question, looking at this test scenario, it is not clear to me how we know that txn version 1 is a version already committed. I'm making the assumption that we are changing to 2 because with 1 it fails, but expanding the lines and looking at the case, I'm not sure why that holds. Please help me understand it.

fpj (changes requested):
Looks good, but I have a small request, can we please say where precisely in the delta protocol doc is the requirement around txn.version described in #274?

src/test/scala/io/qbeast/spark/utils/QbeastSparkTxnTest.scala
Jiaweihu08: @fpj Step 2 specifies when a `txnVersion` is ignored, and only the latest version for a given `appId` is stored in the **snapshot**.

> For example, the [Delta Sink for Apache Spark's Structured Streaming](https://github.com/delta-io/delta/blob/master/core/src/main/scala/org/apache/spark/sql/delta/sources/DeltaSink.scala) ensures exactly-once semantics when writing a stream into a table using the following process:
> 
> 1. Record in a write-ahead-log the data that will be written, along with a monotonically increasing identifier for this batch.
> 2. Check the current version of the transaction with appId = streamId in the target table. **If this value is greater than or equal to the batch being written**, then this data has already been added to the table and processing can skip to the next batch.
> 3. Write the data optimistically into the table.
> 4. Attempt to commit the transaction containing both the addition of the data written out and an updated appId version pair.
> 
> The semantics of the application-specific version are left up to the external system. **Delta only ensures that the latest version for a given appId is available in the table snapshot**. The Delta transaction protocol does not, for example, assume monotonicity of the version and it would be valid for the version to decrease, possibly representing a "rollback" of an earlier transaction.

fpj:
> Looks good, but I have a small request, can we please say where precisely in the delta protocol doc is the requirement around txn.version described in #274?

The requirement is coming from this section of the protocol spec: 

https://github.com/delta-io/delta/blob/master/PROTOCOL.md#transaction-identifiers

src/test/scala/io/qbeast/spark/utils/QbeastSparkTxnTest.scala
fpj: Thanks, @Jiaweihu08, but I think you are responding to my other question. In this thread, I'm asking for a clarification on the test case, could you have another look, please?

fpj: Ok, so where are we validating that a latter attempt is being ignored? My guess is that the `data.write` in lines 108-113 should be ignored, and if it isn't, then the count in line 114 is going to be incorrect and the test case will fail, is that it?

Jiaweihu08: Yes, as well as the one from 102-107.

fpj (approved):
+1, lgtm

PR #305, opened 2024-04-05
Versioning policy documentation

### LINKED ISSUE #285: Add Versioning Document
This issue is for creating a specific Document in qbeast-spark explaining how we update versions, and which is the semantic that we are using for each change. 

First proposal: 

- Use Semantic Versioning as soon we release 1.0.0 ([semver.org](http://semver.org/))
- In the meantime, after 0.6.0, we'll use:
    - 0.6.x for bugfix
    - 0.x.0 for feature release
    - For RCs (release candidates): 0.6.0-rc / 0.6.0-rcx
- We can still use the SNAPSHOT releases for internal testing. (0.6.0-SNAPSHOT)

cdelfosse:
## Description

Fixes #285 : explain how we update the project version

## Type of change

- **Documentation** update only

## Checklist:

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -147,6 +147,29 @@ sbt publishM2
 
 <br/>
 
+# Versioning
+
+Here are some rules on versioning the qbeast-spark project that must be applied before a new version is released.
+
+## post 0.6.0 release
+
+If the current version is 0.y.z, these rules must be applied for a new release:
+- 0.y.z+1 for a bugfix only release
+- 0.z+1.0 for a new feature(s) release
+- 0.y.z-rcN for release candidates, N being a positive integer
+
+## post 1.0.0 release
+
+We'll apply Semantic Versioning rules as defined at (semver.org)[https://semver.org].
+
+## Snapshots
+
+Snapshots can be made available for internal and test purposes.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -147,6 +147,29 @@ sbt publishM2
...
+## post 1.0.0 release
+
+We'll apply Semantic Versioning rules as defined at (semver.org)[https://semver.org].
+
+## Snapshots
+
+Snapshots can be made available for internal and test purposes.
+In this case the versioning to be applied is: <new-version-number>-SNAPSHOT-<short-commit-sha>
+
+Example: 0.6.2-SNAPSHOT-badfbadf
```
osopardo1, line 170: Until now, we were publishing `SNAPSHOT`s with the `sha` in the second position instead of the end.

Is it best to change it to this semantic?

osopardo1: Following the Reference Manual, seems like they always keep the last string for SNAPSHOT

https://www.scala-sbt.org/1.x/docs/Publishing.html

cdelfosse: We can keep the current semantic. I'll update the PR.

osopardo1: Thanks!

osopardo1 (approved):

Issue #294, opened 2024-03-26
Optimization of the Unindexed Files [Staging Area]


osopardo1:
Qbeast Spark supports reading files not indexed with Qbeast Metadata. There's different situations that can cause a table to have a hybrid state.

- Different set of writers. One writes with Qbeast, the others with Delta. All writers can commit new files to the transaction log. Those files written as Delta will not contain any Qbeast Metadata. 
- Old Table in Delta or Parquet converted to Qbeast. When we execute the Convert To Qbeast command, we are just adding a single metadata commit to the table, without rewriting or analyzing any of the existing files. 
- Deletes and Updates. If a Table receives an Update or Delete operation and uses the default Copy on Write strategy, it will create new files that are not indexed. 

The current behavior is to ignore the non-indexed files when reading and writing, thus disabling part of the Sampling capabilities and reducing the precision when estimating the index. Also, optimization of this "staging area", does not select that subset of files for any rearrangement operation.

This issue is to record and analyze which is the best storyline to follow when Optimizing the Non-Indexed files.

osopardo1:
I will update the issue with a better problem formulation. 
We still need to analyze which approaches we want to take, since Compaction, for example, is no longer used, and different techniques such as Index Without Rewrite were designed since this issue was open.

osopardo1:
Oh, at first it sounded weird to me. agree with Unindexed

osopardo1:
Opened PR: #440

Issue #282, opened 2024-03-13
Replication not enabled on Multiblock Format


osopardo1:
## Multiblock Format

The upcoming release of Qbeast Spark has[ new protocol updates](https://github.com/Qbeast-io/qbeast-spark/blob/main-1.0.0/docs/QbeastFormat1.0.0.md#block-metadata-before-the-version-100).

In this modification, we **change the layout of a cube spread in multiple files** into **a single file containing multiple cubes** (divided into blocks). This allows the `roll-up` operation to arrange several small cubes into a bigger file, helping queries to filter data more effectively. 

#### Original protocol metadata:

```json
"tags": {
  "state": "FLOODED",
  "cube": "w",
  "revision": "1",
  "minWeight": "2",
  "maxWeight": "3",
  "elementCount": "4" 
}


```

#### NEW protocol metadata:

```json
"tags": {
  "revision": "1",
  "blocks": "[
    {
      "cube": "w",
      "minWeight": 2,
      "maxWeight": 3,
      "replicated": false,
      "elementCount": 4
    },
    {
      "cube": "wg",
      "minWeight": 5,
      "maxWeight": 6,
      "replicated": false,
      "elementCount": 7
    },
  ]" 
}


```

But changes come with downsides.

## What is Replication?

Replication is the operation that optimizes the index for Sampling and Min-Max distribution.

In summary, it **reads data from the cubes that are overflowed** (containing much more records than their original capacity) **and spreads the information to their children**. 

![image](https://github.com/Qbeast-io/qbeast-spark/assets/22685017/58af7394-d0df-4fa9-bd12-efcb16f48001)

## The problem

When combining **Roll-Up** and **Replication**, the replicated (copied) data of a cube might end up in the same Parquet, thus making not possible the compatibility for reading the file from other underlying sources (delta and parquet) and even from the current qbeast implementation. 

## The situation

> We are removing Replication from the new version of qbeast-spark

It is a very specific feature, and we have to redesign it in a way that **doesn't affect compatibility with other Formats**. Right now, the effort of maintaining the operation is bigger than our development capacity. 

This issue is to resonate ways of writing and interacting with replicated data.

## Proposed solutions

- One solution might be to write replicated data in another folder inside the Table. The solution needs to be elaborated on and proposed in a document, this is only a high-level idea.

fpj:
[...]

PR #310, opened 2024-04-09
Issue 309: Update documentation for 0.6.0 release

### LINKED ISSUE #255: Update Developers Guide version
[Deverloper's Guide]( https://github.com/Qbeast-io/qbeast-spark/blob/main/CONTRIBUTING.md) docs should point to the new version of Qbeast-Spark instead of 0.4.0.

osopardo1:
## Description

Updates the documentation for 0.6.0 release. ( Also fixes #255 )

## Type of change

**Documentation** update (updates regarding the [Users docs](https://github.com/Qbeast-io/qbeast-spark/tree/main/docs) or the [Developer docs](https://docs.qbeast.io/), or similar).

## Checklist:

- [x] Update Spark Version on README.
- [x] Update Delta Version (and packages) on README.
- [x] Update Compatibility Matrix on README.
- [x] Update Spark & Delta Versions on CONTRIBUTING.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@
 
 **Qbeast Spark** is an extension for [**Data Lakehouses**](http://cidrdb.org/cidr2021/papers/cidr2021_paper17.pdf) that enables **multi-dimensional filtering** and **sampling** directly on the storage
 
-[![apache-spark](https://img.shields.io/badge/apache--spark-3.5.x-blue)](https://spark.apache.org/releases/spark-release-3-4-1.html) 
+[![apache-spark](https://img.shields.io/badge/apache--spark-3.5.x-blue)](https://spark.apache.org/releases/spark-release-3-5-0.html) 
 [![apache-hadoop](https://img.shields.io/badge/apache--hadoop-3.3.x-blue)](https://hadoop.apache.org/release/3.3.1.html)
 [![delta-core](https://img.shields.io/badge/delta--core-3.1.0-blue)](https://github.com/delta-io/delta/releases/tag/v2.4.0)
 [![codecov](https://codecov.io/gh/Qbeast-io/qbeast-spark/branch/main/graph/badge.svg?token=8WO7HGZ4MW)](https://codecov.io/gh/Qbeast-io/qbeast-spark)
--- a/README.md
+++ b/README.md
@@ -120,14 +120,29 @@ spark.sql(
 Use **`INSERT INTO`** to add records to the new table. It will update the index in a **dynamic** fashion when new data is inserted.
 
 ```scala
+val studentsDF = Seq((1, "Alice", 34), (2, "Bob", 36)).toDF("id", "name", "age")
 
+studentsDF.write.mode("overwrite").saveAsTable("visitor_students")
+
+// AS SELECT FROM
 spark.sql("INSERT INTO table student SELECT * FROM visitor_students")
 
+// VALUES
+spark.sql("INSERT INTO table student VALUES (3, 'Charlie', 37)")
+
+// SHOW
+spark.sql("SELECT * FROM student").show()
++---+-------+---+
+| id|   name|age|
++---+-------+---+
+|  1|  Alice| 34| 
+|  2|    Bob| 36|
+|  3|Charlie| 37|
++---+-------+---+
 ```
 
 ###  3. Load the dataset
 Load the newly indexed dataset.
-
 ```scala
 val qbeastDF =
   spark.
@@ -147,6 +162,8 @@ Go to the [Quickstart](./docs/Quickstart.md) or [notebook](docs/sampleopushdown_
 ### 5. Interact with the format
 
 Get **insights** or execute **operations** to the data using the `QbeastTable` interface!
+
+> WARNING: ISSUE #[282](https://github.com/Qbeast-io/qbeast-spark/issues/282) Replication not enabled on Multiblock Format (from v.0.6.0). We don't encourage to use this method since it can impact index consistency.
 
 ```scala
 import io.qbeast.spark.QbeastTable
--- a/README.md
+++ b/README.md
@@ -18,9 +18,9 @@
 
 **Qbeast Spark** is an extension for [**Data Lakehouses**](http://cidrdb.org/cidr2021/papers/cidr2021_paper17.pdf) that enables **multi-dimensional filtering** and **sampling** directly on the storage
 
-[![apache-spark](https://img.shields.io/badge/apache--spark-3.4.x-blue)](https://spark.apache.org/releases/spark-release-3-4-1.html) 
+[![apache-spark](https://img.shields.io/badge/apache--spark-3.5.x-blue)](https://spark.apache.org/releases/spark-release-3-4-1.html) 
 [![apache-hadoop](https://img.shields.io/badge/apache--hadoop-3.3.x-blue)](https://hadoop.apache.org/release/3.3.1.html)
-[![delta-core](https://img.shields.io/badge/delta--core-2.4.0-blue)](https://github.com/delta-io/delta/releases/tag/v2.4.0)
+[![delta-core](https://img.shields.io/badge/delta--core-3.1.0-blue)](https://github.com/delta-io/delta/releases/tag/v2.4.0)
 [![codecov](https://codecov.io/gh/Qbeast-io/qbeast-spark/branch/main/graph/badge.svg?token=8WO7HGZ4MW)](https://codecov.io/gh/Qbeast-io/qbeast-spark)
 
 </div>
@@ -62,16 +62,16 @@ You can run the qbeast-spark application locally on your computer, or using a Do
 You can find it in the [Packages section](https://github.com/orgs/Qbeast-io/packages?repo_name=qbeast-spark).
 
 ### Pre: Install **Spark**
-Download **Spark 3.4.1 with Hadoop 3.3.4**, unzip it, and create the `SPARK_HOME` environment variable:<br />
+Download **Spark 3.5.0 with Hadoop 3.3.4**, unzip it, and create the `SPARK_HOME` environment variable:<br />
 
 >:information_source: **Note**: You can use Hadoop 2.7 if desired, but you could have some troubles with different cloud providers' storage, read more about it [here](docs/CloudStorages.md).
 
 ```bash
-wget https://archive.apache.org/dist/spark/spark-3.4.1/spark-3.4.1-bin-hadoop3.tgz
+wget https://archive.apache.org/dist/spark/spark-3.5.0/spark-3.5.0-bin-hadoop3.tgz
 
-tar -xzvf spark-3.4.1-bin-hadoop3.tgz
+tar -xzvf spark-3.5.0-bin-hadoop3.tgz
 
-export SPARK_HOME=$PWD/spark-3.4.1-bin-hadoop3
+export SPARK_HOME=$PWD/spark-3.5.0-bin-hadoop3
  ```
 ### 1. Launch a spark-shell
 
@@ -79,7 +79,7 @@ export SPARK_HOME=$PWD/spark-3.4.1-bin-hadoop3
 
 ```bash
 $SPARK_HOME/bin/spark-shell \
---packages io.qbeast:qbeast-spark_2.12:0.5.0,io.delta:delta-core_2.12:2.4.0 \
+--packages io.qbeast:qbeast-spark_2.12:0.6.0,io.delta:delta-spark_2.12:3.1.0 \
 --conf spark.sql.extensions=io.qbeast.spark.internal.QbeastSparkSessionExtension \
 --conf spark.sql.catalog.spark_catalog=io.qbeast.spark.internal.sources.catalog.QbeastCatalog
 ```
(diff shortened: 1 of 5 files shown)


cdelfosse (approved):

Jiaweihu08 (approved):

osopardo1:
Added some fix regarding the `QbeastTable` package in `Quickstart.md` and the `README.md` INSERT INTO example.

fpj (approved):
+1, thanks for the update.

cdelfosse (approved):
LGTM

osopardo1:
I would update here the API for optimize.

README.md
```diff
@@ -155,31 +170,57 @@ val qbeastTable = QbeastTable.forPath(spark, tmpDir)
...
+
+### 6. Optimize the table
+
+**Optimize** is a very expensive operation that consist on **rewriting part of the files to accomplish an even distribution of the records withing files**.
+
+To minimize the write amplification of this command, **we execute it based on subsets of the table**, like `Revision ID's` or specific files.
+
+`Revisions` determine the area of the index. **The last Revision would contain the whole dataset area**, while previous Revisions would contain a _subset_ of the area written before. It can be useful to optimize by specific Revision ID's when we have an incremental appends scenario. 
+
+As an example, imagine you start writting a Dataset with the data from France, and the second time you append records from Germany. The first Revision would contain only the French countries, while the second would enlarge the space to also include Germany. If you are querying only French data, you would be interested in optimizing only the first Revision.
```
fpj, line 183: We need to refine this example, here is some suggested text, but don't take it blindly, please review and refine as needed:

> Say that the initial commit C1 inserting data into the table added data about cities in France, assuming a column called city name that we are indexing with Qbeast. With this commit, we add three files F1, F2 and F3. A second commit C2 adds data about German cities, so it expands the cardinality of the city name column. This commit adds another four files F4, F5, F6 and F7.
>
> The first commit C1 creates a revision R0, while C2 adds another revision R1. Revision R1 is necessary because it expands the cardinality of the city name dimension, and this expansion impacts the index as the range of values changes.
>
> Optimizing consequently can be performed over either revision as we elaborate below. Optimizing, for example, over R0 rewrites files F1, F2 and F3.

README.md
```diff
@@ -155,31 +170,57 @@ val qbeastTable = QbeastTable.forPath(spark, tmpDir)
 
 qbeastTable.getIndexMetrics()
 
-qbeastTable.analyze()
+```
+
+### 6. Optimize the table
+
+**Optimize** is a very expensive operation that consist on **rewriting part of the files to accomplish an even distribution of the records withing files**.
```
fpj, line 177: Could we please explain what optimizing a table entails? We said yesterday that it is about moving offsets to their legitimate cube/file and performing rollups of small files into their parents..

docs/QbeastFormat.md
```diff
@@ -18,12 +18,12 @@ Following each write transaction is the creation of a new log file. **Table-leve
 
 ## Qbeast-spark on Delta
 
-Qbeast format extends the Delta Lake Format, so the data is stored in Parquet
+Qbeast format **extends the Delta Lake Format**, so the data is stored in Parquet
 files and the Index Metadata is stored in the Delta log.
```
fpj, line 22: Either give an example or point to another location that has an example. I recall seeing the schema of an addFile entry somewhere in the documentation.

docs/QbeastFormat.md
```diff
@@ -63,15 +63,15 @@ Here you can see the changes on the `AddFile` **`tags`** information
...
-| `minWeight`    | The minimum weight of the block.                                                                                                             |
-| `maxWeight`    | The maximum weight of the block.                                                                                                             |
-| `elementCount` | The number of elements in the block.                                                                                                         |
-| `replicated`   | A flag that indicates if the block is replicated. Replication is **not available** from v0.6.0.                                                  |
+| Term           | Description                                                                                                    |
+|----------------|----------------------------------------------------------------------------------------------------------------|
+| `revision`     | **Locates a particular `min`-`max` range of the columns** in the space, and includes a Timestamp of the creation time. |
```
fpj, line 68: This is interesting, but I'm not sure I understand. How does it determine what min-max range is relevant and what if there are multiple files touching that min-max range? Is this adding a timestamp to tags in the addFile entry or somewhere else?

docs/QbeastFormat.md
```diff
@@ -63,15 +63,15 @@ Here you can see the changes on the `AddFile` **`tags`** information
...
+| `blocks`       | A **list of the blocks** that are stored inside the file. Can belong to different cubes.                           |
```
fpj, line 69: The concept of blocks is a bit weird. Blocks at least in storage are addressable, like I can fetch a block from a disk drive by providing a block number. I'm under the impression that we cannot really address a block directly, in which case, it only serves the purpose of indicating that a file contains data from a particular cube?

docs/QbeastFormat.md
```diff
@@ -63,15 +63,15 @@ Here you can see the changes on the `AddFile` **`tags`** information
...
+| `cube`         | The **serialized representation of the Cube**.                                                                     |
```
fpj, line 70: I don't understand what serialized means. Is it a hash ID or something similar?

docs/QbeastFormat.md
```diff
@@ -63,15 +63,15 @@ Here you can see the changes on the `AddFile` **`tags`** information
...
+| `minWeight`    | The minimum weight of the block.                                                                               |
+| `maxWeight`    | The maximum weight of the block.                                                                               |
+| `elementCount` | The number of elements in the block.                                                                           |
+| `replicated`   | A flag that indicates **if the block is replicated**. Replication is **not available** from v0.6.0.                |
```
fpj, line 74: What's going to happen to replication, do we know or is it up in the air?

docs/QbeastFormat.md
```diff
@@ -145,19 +145,19 @@ A closer look to the `qb.revision.$number`:
...
+| Term                            | Description                                                                     |
+|---------------------------------|---------------------------------------------------------------------------------|
+| `timestamp`                     | The time when the revision was created.                                         |
+| `tableID`                       | The identifier of the table that the revision belongs to.                       |
+| `desiredCubeSize`               | The cube size from the option `cubeSize`.                                       |
+| `columnTransformers`            | The metadata of the different columns indexed with the option `columnsToIndex`. |
+| `columnTransformers.columnName` | The **name of the column**.                                                     |
+| `columnTransformers.dataType`   | The **data type of the column**.                                                |
+| `transformations`               | Contains information about the **space mapping** of the data indexed by column. |
```
fpj, line 156: What are these transformations exactly?

fpj (changes requested):
Thanks for the PR, @osopardo1 . I have left a few comments, suggestions and questions.

README.md
```diff
@@ -18,9 +18,9 @@
 
 **Qbeast Spark** is an extension for [**Data Lakehouses**](http://cidrdb.org/cidr2021/papers/cidr2021_paper17.pdf) that enables **multi-dimensional filtering** and **sampling** directly on the storage
 
-[![apache-spark](https://img.shields.io/badge/apache--spark-3.4.x-blue)](https://spark.apache.org/releases/spark-release-3-4-1.html) 
+[![apache-spark](https://img.shields.io/badge/apache--spark-3.5.x-blue)](https://spark.apache.org/releases/spark-release-3-5-0.html)
```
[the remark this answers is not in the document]
osopardo1: Is the webpage that we use for the batches in the README.md. 

The ones in the image below:

image

README.md
osopardo1: I would only change that C1 creates revision R1 and C2 creates R2. Revision 0 should be used only for files that are not indexed / staging area. 

Now that I am writing it down, I think we should add that information too.

README.md
osopardo1: Do you think such specific information should be on the README? Or is it better to move all the explanation about optimize operation into the QbeastTable.md ?

docs/QbeastFormat.md
osopardo1: Is in the same document 👍

docs/QbeastFormat.md
osopardo1: Is it on the issue number #282 

Not clear resolution right now 😢

docs/QbeastFormat.md
osopardo1: I would change it for "the string representation of..."

docs/QbeastFormat.md
osopardo1: Transformations are the functions that map the **real values of a column indexed into the [0.0, 1.0) space.**

docs/QbeastFormat.md
osopardo1: Ideally, we should be able to retrieve only the blocks from a certain file that contains data that satisfies the query. The functionality is not implemented, though it requires some #175 support to be able to discard row groups from Parquet on the fly.

docs/QbeastFormat.md
osopardo1: The min-max is determined by the data itself. 
Every append has a range. An append A1 may or may not trigger a Revision Change (new Revision R2) if the ranges of the batch are larger than the previous revision (R1). 
The timestamp is added to the Revision schema. Each file entry in the log contains only the ID for the Revision. (1 or 2 in the case above)

docs/QbeastFormat.md
osopardo1: Updated the description!

README.md
osopardo1: I've moved part of the definition and example to the `QbeastTable` doc.

fpj: That's fine, maybe point here to the QbeastTable doc for completeness.

docs/QbeastFormat.md
fpj: I see, let's leave it as is then, maybe point to the issue #282 so that anyone interested can follow the conversation?

docs/QbeastFormat.md
```diff
@@ -16,28 +16,69 @@ Following each write transaction is the creation of a new log file. **Table-leve
...
 Let's take a look to the Index metadata to understand what is written in the Delta Log JSON files.
 
-### Index Metadata
+## Index Metadata
+
+On a high level:
 
-On a high level, the index consists of one or more `OTree` that contain `cubes` (or nodes), and each cube is made of `blocks` that contain the actual data written by the user.
-All records from the log are of **file-level** information.
+- the index consists of one or more `Otree` (search trees) that contain `cubes` (or nodes).
```
fpj, line 75: "... one or more OTree instrances...."

docs/QbeastFormat.md
fpj: Cool, why is is "transformations" plural, is it because there is going to be one for each column domain?

docs/QbeastFormat.md
fpj: Ok, I see you updated it with the issue number, thank you.

docs/QbeastFormat.md
```diff
@@ -182,9 +224,63 @@ ConvertToQbeastCommand(tableIdentifier, columnsToIndex, desiredCubeSize).run(spa
...
+
+```scala
+val revisions = qbeastTable.revisionsIDs() // Get all the Revision ID's available in the table.
+revisions.foreach(revision => 
+  qbeastTable.optimize(revision)
+)
+```
+> Note that **Revision ID number 0 is reserved for Stagin Area** (non-indexed files). This ensures compatibility with underlying table formats.
+
+## Compaction
```
fpj, line 283: I understand the general compaction concept, but can we elaborate on what it is doing for us specifically here?

docs/QbeastFormat.md
osopardo1: Correct 🥇

README.md
osopardo1: Yep, I've added two links: one in the Revision paragraph and other at the end of the section 👍

docs/QbeastFormat.md
osopardo1: Sure! I will also add that the behaviour is different than previous 0.6.0 version (now calls optimize() underneath) and will reference issue #294

docs/QbeastFormat.md
osopardo1: Ok, seems good to me

docs/QbeastFormat.md
osopardo1: Even the timestamp reference can be removed, it can be confusing to mention it when it's not the main characteristic of the revision, and the full schema is explained after.

What do you think of:

> A revision establishes the min-max range for each indexed column, amongst other metadata from a particular state of the index (creation timestamp, cube size...)

docs/QbeastFormat.md
```diff
@@ -72,7 +72,7 @@ Let's take a look to the Index metadata to understand what is written in the Del
 
 On a high level:
```
fpj, line 73: "At a high level:"

docs/QbeastFormat.md
```diff
@@ -231,8 +231,8 @@ By doing so, we also enable subsequent appends using either delta or qbeast.
 ## Optimization
 
 **Optimize** is an expensive operation that consist on:
```
fpj, line 233: **Optimize** is an expensive operation that consists of:

fpj (changes requested):
Really tiny typos, pls fix them quickly and I'll get this committed, it looks good otherwise.

docs/QbeastFormat.md
osopardo1: Solved in [3584051](https://github.com/Qbeast-io/qbeast-spark/pull/310/commits/358405183d8d9aa6a357277cc80526fbc92448b1)

fpj (approved):
+1, lgtm.

PR #319, opened 2024-04-23
Issue 321: Pre-commit hook

### LINKED ISSUE #321: Add Commit Hooks to write extra information within the same transaction
Table Formats encapsulate write actions into an Optimistic Transaction. Various processes could try to commit the info to the Transaction Log, but only one would succeed, making the others retry the operation.

When the user wants to write any extra metadata about the process or to do any checks on the data, it has to wait until the transaction is successfully over, thus leading to delays and possible inconsistencies. 

The idea is to add commit hooks that allow to **run custom code before or after calling the Delta Lake `txn.commit()` method**.

Jiaweihu08:
# Pre-commit Hooks
This PR introduces **pre-commit hooks** for the qbeast format, which enables the execution of custom code just before a write or optimization is committed. More details are described in Issue #321.

Fixes #321 and #323 

The hook should extend `io.qbeast.spark.delta.hook.PreCommitHook,` overriding the `run` method, which has access to the sequence of `Action`s created by the operation. The same method returns a `Map[String, String],` which will be used as `tags` for the transaction's `CommitInfo`.

## Hooks for Writes:
1. You can use more than one hook, as shown in the case below: `myHook1`, and `myHook2.`
2. For each hook you want to use, provide their class names with the option name: `qbeastPreCommitHook.<custom-hook-name>.`
3. Add an option with the name `qbeastPreCommitHook.<custom-hook-name>.arg` for the ones that take initiation arguments. Currently, only one String argument is allowed for each hook.

```scala
(df
  .write
  .format("qbeast")
  .option("qbeastPreCommitHook.myHook1", classOf[SimpleHook].getCanonicalName)
  .option("qbeastPreCommitHook.myHook2", classOf[StatefulHook].getCanonicalName)
  .option("qbeastPreCommitHook.myHook2.arg", myStringHookArg)
  .save(pathToTable)
```

## Hooks for Optimizations:

```scala
val qt = QbeastTable.forPath(spark, tablePath)
val options = Map(
  "qbeastPreCommitHook.myHook1" -> classOf[SimpleHook].getCanonicalName,
  "qbeastPreCommitHook.myHook2" -> classOf[StatefulHook].getCanonicalName,
  "qbeastPreCommitHook.myHook2.arg" -> "myStringHookArg"
)
qt.optimize(filesToOptimize, options)
```

# Example:
1. Create your custom hook
```scala
import io.qbeast.spark.delta.hook.PreCommitHook
import io.qbeast.spark.delta.hook.PreCommitHook.PreCommitHookOutput
import org.apache.spark.sql.delta.actions.Action

class SimpleHook extends PreCommitHook {

  override val name: String = "SimpleHook"

  override def run(actions: Seq[Action]): PreCommitHookOutput = {
    Map("clsName" -> "SimpleHook")
  }

}
```

2. Add the hook project jar to the spark-shell session
```bash
$SPARK_HOME350/bin/spark-shell \
--master "local[1]" \
--jars <path-to-qbeast-spark-jar>,<path-to-hook-jar> \
--packages io.delta:delta-spark_2.12:3.1.0 \
--conf spark.sql.catalog.spark_catalog=io.qbeast.spark.internal.sources.catalog.QbeastCatalog
```

3. Write data with a `SimpleHook`
```scala
import io.test.hook.SimpleHook
import spark.implicits._

val tmpDir = "/tmp/test"
val df = spark.sparkContext.range(0, 100).toDF()

(df
	.write
	.mode("append")
	.format("qbeast")
	.option("columnsToIndex", "value")
	.option("qbeastPreCommitHook.hook", classOf[SimpleHook].getCanonicalName)
	.save(tmpDir)
)
```

4. Check results
```bash
cat /tmp/test/_delta_log/00000000000000000000.json | jq
```

```bash
{
  "commitInfo": {
    ...,
    "tags": {
      "clsName": "SimpleHook"
    },
    ...
  }
}
```


fpj:
What's the issue number describing what issue this is resolving?

osopardo1:
> What's the issue number describing what issue this is resolving?

There's none. I will open one 👍

osopardo1 (commented):
We should review if Pre Commit Hooks are the best way of introducing changes to a different table... 

I think that:
- for registering metrics of the operation, we should use Post Commit Hooks. 
- for quality checks (schema, data...) we can leverage Pre Commit Hooks.


Plus few comments on the code:

osopardo1:
The code seems ready to review, but we will need to provide a better description of the feature, alongside logging and documentation.

src/main/scala/io/qbeast/spark/delta/hook/QbeastHook.scala
```diff
@@ -0,0 +1,66 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.qbeast.spark.delta.hook
+
+import io.qbeast.spark.internal.QbeastOptions
+import org.apache.spark.sql.delta.actions.Action
+
+/**
+ * A hook that can be run before a commit
```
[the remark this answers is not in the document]
osopardo1: So... no `PostCommitHook` for now. Starting with the Pre.

src/test/scala/io/qbeast/spark/utils/QbeastCompactionIntegrationTest.scala
```diff
@@ -239,7 +239,7 @@ class QbeastCompactionIntegrationTest extends QbeastIntegrationTestSpec with Log
       val qt = QbeastTable.forPath(spark, tmpDir)
       val m = qt.getIndexMetrics()
       val filePath = m.cubeStatuses.values.flatMap(_.blocks.map(_.file)).head.path
-      qt.optimize(Seq(filePath))
+      qt.optimize(Seq(filePath), Map.empty[String, String])
```
osopardo1, line 242: Should we leave an empty map as a default value for `options` in the method?

docs/AdvancedConfiguration.md
```diff
@@ -287,3 +287,36 @@ We can empty the staging area with a given write by setting the staging size to
 ```scala
 --conf spark.qbeast.index.stagingSizeInBytes=0
 ```
+
+## Pre-commit Hooks
```
osopardo1, line 291: I think would be useful to include this section in the `QbeastFormat.md` too

src/test/scala/io/qbeast/spark/delta/hook/PreCommitHookIntegrationTest.scala
```diff
@@ -0,0 +1,88 @@
...
+    testOutput
+  }
+
+}
+
+private object SimpleHook {
+  val testOutput: Map[String, String] = Map("clsName" -> "SimpleHook")
+}
+
+class PreCommitHookIntegrationTest extends QbeastIntegrationTestSpec {
```
osopardo1, line 27: Should we test as well the execution of several precommit hooks?

osopardo1 (commented):
In general looks very good!! Just a few things on documentation/options..

src/test/scala/io/qbeast/spark/utils/QbeastCompactionIntegrationTest.scala
Jiaweihu08: I was given the following error: "multiple overloaded alternatives of method optimize define default arguments."

docs/AdvancedConfiguration.md
Jiaweihu08: I'd say they aren't exactly format-related.

src/test/scala/io/qbeast/spark/delta/hook/PreCommitHookIntegrationTest.scala
Jiaweihu08: Done!

docs/AdvancedConfiguration.md
osopardo1: Yes, but it can affect the commited metadata, right?

Jiaweihu08: Yes, the `CommitInfo.` But it won't change any `metadata`.

src/test/scala/io/qbeast/spark/utils/QbeastCompactionIntegrationTest.scala
osopardo1: I think that is good for a WARN, but we should not overcomplicate the operation for the user...

docs/AdvancedConfiguration.md
Jiaweihu08: Added example hook output for `CommitInfo.tags.`

src/test/scala/io/qbeast/spark/utils/QbeastCompactionIntegrationTest.scala
Jiaweihu08: I'll address this in a PR for #337.

osopardo1 (approved):

PR #396, opened 2024-09-02
Issue 389: Documentation Updates should not trigger CI tests

### LINKED ISSUE #389: Documentation Updates should not trigger CI tests
Even if we change the Documentation files, the whole CI is triggered, which takes around 1h to complete. 

It is better to add a policy to ignore those tests when only `.md` files are modified.

jorgeMarin1:
## Description

Fixes #389. In this PR I will try to edit the CI that is executed in github actions everytime someone pushes some changes to a Pull Request in the qbeast-spark repository. If only markdown files have been modified, we will try to skip most of the steps of the worfklow.

Please review whenever you can @cdelfosse @osopardo1 

## Type of change

I have created a new step on the workflow. This step will check if the modified files are markdown files, if that's not the case a variable called `run tests` will be set to true so that the remaining steps can be executed by using if statements. If only .md files are modified, the `run tests`variable will remain unchanged and the next steps will be skipped.

## Checklist:

- [X] New feature / bug fix has been committed following the [Contribution guide](https://github.com/Qbeast-io/qbeast-spark/blob/main/CONTRIBUTING.md).

### DIFF
--- a/.github/workflows/test-artifact.yml
+++ b/.github/workflows/test-artifact.yml
@@ -14,8 +14,8 @@ jobs:
     - uses: actions/checkout@v2
       with:
         fetch-depth: 0
-    - name: Show modified files in the latest commit
-      run: git show --name-only --oneline HEAD
+    - name: Show modified files in this PR
+      run: git diff --name-only origin/main...HEAD
 
     - name: Show initial value of run_tests
       run: |
@@ -23,16 +23,15 @@ jobs:
     - name: Check if only .md files were changed
       id: check_md_files
       run: |
-          if git show --name-only --oneline HEAD | grep -qv '\.md$'; then
+          if git diff --name-only origin/main...HEAD | grep -qv '\.md$'; then
             echo "run_tests=true" >> $GITHUB_ENV
           else
             echo "Only markdown files were changed. Skipping steps."
           fi
-    - name: Show value of run_tests after checking .md files
+    - name: Show value of run_tests after checking changed files
       run: |
           echo "Value of run_tests after check: ${{ env.run_tests }}"
     - name: Set up JDK 1.8
-      if: env.run_tests == 'true'
       uses: actions/setup-java@v1
       with:
         java-version: 1.8
--- a/.github/workflows/test-artifact.yml
+++ b/.github/workflows/test-artifact.yml
@@ -14,8 +14,8 @@ jobs:
     - uses: actions/checkout@v2
       with:
         fetch-depth: 0
-    - name: Show modified files
-      run: git diff --name-only HEAD^ HEAD
+    - name: Show modified files in the latest commit
+      run: git diff-tree --no-commit-id --name-only -r HEAD
 
     - name: Show initial value of run_tests
       run: |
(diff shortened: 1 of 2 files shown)


osopardo1 (approved):
LGTM!!

Let's wait for @cdelfosse review and if everything is ok I will merge this 👍

jorgeMarin1:
Cesare said we should use a workflow that looks like this one: [USE frontend test](https://github.com/Qbeast-io/USE/blob/main/.github/workflows/frontend-tests-and-lint.yaml).

I'm not fully convinced about this. That workflow is executed when files are modified in a specific folder, whereas we have a workflow that runs when we change ANY file of the project and we only need to skip some of the steps of that workflow when only .md files have been changed.

It's true we could use the paths statement from the workflow that Cesare showed us if we had another workflow for changed markdown files, but if the objective is to reduce the runtime of the existing test I think we should stick to if statments. What do you guys think?

jorgeMarin1:
Quick summary of what I have added to the workflow:

- We use a new env. variable called `run_tests` to evaluate if we need to run all the steps or if we can skip most of them. It is initially set to false.
```
jobs:
  test-artifact:
    runs-on: ubuntu-latest
    env:
      GHPR_TOKEN: ${{ secrets.GHPR_TOKEN }}
      run_tests: 'false'
```
- We have a debugging step that shows the modified files in the created PR. It compares the files that have been changed compared to the main branch of the repository. In our case we only modified the ``test-artifact.yml`` file
```
Run git diff --name-only origin/main...HEAD
.github/workflows/test-artifact.yml
```
- We added a step that sets run_tests to true if we modified files that are not markdown. If we only changed .md files the variable run_tests remains as false and we skip the rest of the steps, except for the compiling step.
```
if git diff --name-only origin/main...HEAD | grep -qv '\.md$'; then
    echo "run_tests=true" >> $GITHUB_ENV
  else
    echo "Only markdown files were changed. Skipping steps."
  fi
```
- We have two debugging steps that show the value of run_tests before and after checking if we only modified .md files (Can be deleted if you think we don't need it).- We have two debugging steps that show the value of run_tests before and after checking if we only modified .md files (Can be deleted if you think we don't need it).

cdelfosse:
> Quick summary of what I have added to the workflow:
> 
> * We use a new env. variable called `run_tests` to evaluate if we need to run all the steps or if we can skip most of them. It is initially set to false.
> 
> ```
> jobs:
>   test-artifact:
>     runs-on: ubuntu-latest
>     env:
>       GHPR_TOKEN: ${{ secrets.GHPR_TOKEN }}
>       run_tests: 'false'
> ```
> 
> * We have a debugging step that shows the modified files in the created PR. It compares the files that have been changed compared to the main branch of the repository. In our case we only modified the `test-artifact.yml` file
> 
> ```
> Run git diff --name-only origin/main...HEAD
> .github/workflows/test-artifact.yml
> ```
> 
> * We added a step that sets run_tests to true if we modified files that are not markdown. If we only changed .md files the variable run_tests remains as false and we skip the rest of the steps, except for the compiling step.
> 
> ```
> if git diff --name-only origin/main...HEAD | grep -qv '\.md$'; then
>     echo "run_tests=true" >> $GITHUB_ENV
>   else
>     echo "Only markdown files were changed. Skipping steps."
>   fi
> ```
> 
> * We have two debugging steps that show the value of run_tests before and after checking if we only modified .md files (Can be deleted if you think we don't need it).- We have two debugging steps that show the value of run_tests before and after checking if we only modified .md files (Can be deleted if you think we don't need it).
> 
> From now on, if we need to modifiy the documentation, we should create a PR only to change the .md files and we shouldn't have to wait too much time...

Why not using [paths-ignore](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#example-excluding-paths)? It should work and is less complex.

cdelfosse (approved):
LGTM

PR #401, opened 2024-09-06
Issue #264: Update qviz

### LINKED ISSUE #264: Update qviz
## What went wrong?
Qviz needs update because with the introduction of virtual cube blocks, where a single parquet file can contain data from multiple cubes, it is not parsing the _delta_log properly.

Branch and commit id: `main-1.0.0`, [8bc87dc](https://github.com/Qbeast-io/qbeast-spark/commit/8bc87dc8cb839130f2d82478883d9875e1e4fc65)

jorgeMarin1:
## Description

Fixes #264 .

With the new qbeast format, when we try to run the OTree Index Visualization (qviz) it crashes, so we need to update it.

## Type of change

Bug fix.

## Checklist:

- [X] Add tests.

The OTree Visualization (qviz) was run locally on my PC.

Also, a pyspark shell was created in order to do some tests at the implemented fix. Then, a huge table of 4.000.000 rows was created, full of fake data. From this table, three Qbeast tables were created and read as Delta tables.

From these tables, we compared each of their total number of elements with the sumatory of the element_count of their respective cubes. Since these values matched, we concluded it worked as it should.


jorgeMarin1:
For future reference:

df = deltaTable.get_add_actions(True).to_pandas() returns a dataframe with these columns:

Index(['path', 'size_bytes', 'modification_time', 'data_change', 'num_records',
       'null_count.event_time', 'min.event_time', 'max.event_time',
       'null_count.event_type', 'min.event_type', 'max.event_type',
       'null_count.product_id', 'min.product_id', 'max.product_id',
       'null_count.category_id', 'min.category_id', 'max.category_id',
       'null_count.category_code', 'min.category_code', 'max.category_code',
       'null_count.brand', 'min.brand', 'max.brand', 'null_count.price',
       'min.price', 'max.price', 'null_count.user_id', 'min.user_id',
       'max.user_id', 'null_count.user_session', 'min.user_session',
       'max.user_session', 'tags.blocks', 'tags.revision'], dtype='object')

jorgeMarin1:
Blocks are obtained by using: `blocks = df["tags.blocks"]`

These blocks are in a panda Series and have the following structure:
`[{"cubeId":"wggg","minWeight":-103008559,"maxWeight":2147483647,"elementCount":9521,"replicated":false},{"cubeId":"wggQ","minWeight":-100165008,"maxWeight":2147483647,"elementCount":880,"replicated":false},{"cubeId":"wgg","minWeight":-1401341541,"maxWeight":-103459089,"elementCount":9759,"replicated":false},{"cubeId":"wggA","minWeight":-102673026,"maxWeight":2147483647,"elementCount":5982,"replicated":false},{"cubeId":"wggw","minWeight":-91591261,"maxWeight":2147483647,"elementCount":1242,"replicated":false}]`

jorgeMarin1:
This is the structure of the metadata for revision_id = 1 : 

Metadata:  {'revisionID': 1, 'timestamp': 1726134712571, 'tableID': '/Users/jorgemarin/Documents/qviz-bug/qbeast-spark/utils/visualizer/tests/resources/test_table_delta_multiples_writes/', 'desiredCubeSize': 10000, 'columnTransformers': [{'className': 'io.qbeast.core.transform.LinearTransformer', 'columnName': 'user_id', 'dataType': 'IntegerDataType'}, {'className': 'io.qbeast.core.transform.StringHistogramTransformer', 'columnName': 'brand', 'dataType': 'StringDataType'}], 'transformations': [{'className': 'io.qbeast.core.transform.LinearTransformation', 'minNumber': 274969076, 'maxNumber': 566347286, 'nullValue': 391116828, 'orderedDataType': 'IntegerDataType'}, {'className': 'io.qbeast.core.transform.StringHistogramTransformation', 'histogram': ['a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i', 'j', 'k', 'l', 'm', 'n', 'o', 'p', 'q', 'r', 's', 't', 'u', 'v', 'w', 'x', 'y', 'z']}]}

Jiaweihu08:
Can you add `ruff` for code formatting? @jorgeMarin1

Jiaweihu08:
Also, remove the redundant test tables. You can use the most complex one to cover all the test cases.

utils/visualizer/qviz/content_loader.py
```diff
@@ -2,250 +2,264 @@
...
-            path = entry['remove']['path']
-            add_files_dict.pop(path, None)
-
-    return list(add_files_dict.values())
-
-
-def addFiles_from_json_log_files(json_files: list[str], revision_id: str) -> (list[dict], set[str]):
+    try:
+        # Try to find the table in the provided path 
+        deltaTable = DeltaTable(table_path)
```
Jiaweihu08, line 110: Use snake case

utils/visualizer/qviz/content_loader.py
```diff
@@ -2,250 +2,264 @@
...
+        df['tags.blocks'] = df['tags.blocks'].apply(lambda x: json.loads(x) if isinstance(x, str) else x)
+        # Make sure that 'tags.blocks' is a list of dictionaries
+        exploded_df = df.explode('tags.blocks')
+        # Extract 'cubeId'
+        all_cube_ids = exploded_df['tags.blocks'].apply(lambda x: x['cubeId'] if isinstance(x, dict) and 'cubeId' in x else None).dropna().tolist()
+
+        for cube_id in all_cube_ids:
+            cube_encoding_size = len(cube_id)
+            if 0 < cube_encoding_size < symbol_count:
+                symbol_count = cube_encoding_size
```
Jiaweihu08, line 161: Why is `symbol_count` computed this way?

utils/visualizer/qviz/content_loader.py
```diff
@@ -2,250 +2,264 @@
...
-    Extract AddFiles and RemoveFiles from all json log files
-    :param json_files: list, a list of json log file paths
-    :param revision_id: target revisionID
-    :return: a list of AddFile and a set of RemoveFile paths
+    Extract metadata from a Delta table by checking if the revision ID is in that table.
+    If that ID exists in the table, a revision key (e.g. qbeast.revision.1) is created and the metadata is returned.
+    It also calculates the number of symbols used to representate the visualization.
+    :param delta_table: a created delta table.
+    :param revision_id: Configuration entry for the target RevisionID. e.g. 1
+    :return: A dictionary with the metadata related to the revisionId, an integer representing the number of simbols used fot the visualization representation.
```
Jiaweihu08, line 126: typo `fot`

utils/visualizer/qviz/content_loader.py
```diff
@@ -2,250 +2,264 @@
...
+
+    df = delta_table.get_add_actions(True).to_pandas()
+
+    df_filtered = df[['size_bytes', 'tags.blocks']]
+    # In this dictionary we store a cube for every cubeId. Each cube will be associated to his cube ID
+    cubes_dict = dict()
+
+    # Iterate throughout each row of the dataframe
+    for index, row in df_filtered.iterrows():
+        cube = row["tags.blocks"] # This returns a pandas series with the blocks of the cube in each of the iterated rows
```
Jiaweihu08, line 182: Make sure you are naming these variables correctly.
Here, the list of blocks should be called `blocks` instead.

utils/visualizer/qviz/content_loader.py
jorgeMarin1: This has been deleted, if we don't find the revision ID we will show a message saying the qbeast table doesn't exist.

utils/visualizer/qviz/content_loader.py
```diff
@@ -2,250 +2,261 @@
...
+def process_table(
+    table_path: str, revision_id: str
+) -> tuple[dict, dict, Cube, list[dict]]:
     """
-    Extract valid AddFiles from table's _delta_log
-    :param table_path: path to qbeast table
-    :param revision_id: target RevisionID
-    :return: a list of AddFile dictionaries and optionally the metadata dictionary
+    Process a delta table given a path to a Qbeast table and a revision ID.
+    Example json log file content:
```
Jiaweihu08, line 18: We don't need to show all this information here.
Keep only related information, maybe also something else you find relevant, and remove everything else.

utils/visualizer/qviz/content_loader.py
```diff
@@ -2,250 +2,261 @@
...
+        {"protocol":
+            {"minReaderVersion":1,
+            "minWriterVersion":2}}
+        {"add":
+            {"path":"8eb0fcd7-c5ca-4b4c-a967-bc10188be935.parquet",
+            "partitionValues":{},
+            "size":814021,
+            "modificationTime":1726134717043,
+            "dataChange":true,
+            "stats":"{
```
Jiaweihu08, line 50: `stats` and `partitionValues` can surely be removed.
Also, you don't need to keep too many blocks there.

utils/visualizer/qviz/content_loader.py
```diff
@@ -2,250 +2,261 @@
...
-
-def addFiles_from_json_log_files(json_files: list[str], revision_id: str) -> (list[dict], set[str]):
+    try:
+        # Try to find the table in the provided path
+        delta_table = DeltaTable(table_path)
+        return delta_table
+
+    except Exception as e:
+        # If the table doesn't exist, throw an exception
+        print(f"Failed when creating the delta table: {e}\n")
```
Jiaweihu08, line 118: Are you really throwing an exception here?

utils/visualizer/qviz/content_loader.py
```diff
@@ -2,250 +2,261 @@
...
+            symbol_count = (dimension_count + 5) // 6
+            print(symbol_count)
+            return metadata_key, symbol_count
+    else:
+        print(f"No metadata found for the given RevisionID. There is no Qbeast table.")
+        return None
 
-def load_single_log_file(file_path: str, revision_id: str) -> (list[dict], set[str]):
+# 3. Returns a dictionary, the keys are the cubeids and the values are the actual cubes
+def extract_cubes_from_blocks(delta_table: DeltaTable, symbol_count: int) -> dict:
```
Jiaweihu08, line 152: Try to find a better name for this method. You are extracting cubes from the table, and you are not providing blocks as inputs.

utils/visualizer/qviz/content_loader.py
```diff
@@ -2,250 +2,261 @@
...
+    # In this dictionary we store a cube for every cubeId. Each cube will be associated to his cube ID
+    cubes_dict = dict()
+
+    # Iterate throughout each row of the dataframe
+    for index, row in df_filtered.iterrows():
+        blocks_str = row[
+            "tags.blocks"
+        ]  # This returns a pandas series with the blocks of the cube in each of the iterated rows
+        # Check if the cube is a JSON chain & try to convert it into a list of dictionaries
+        if isinstance(blocks_str, str):
```
Jiaweihu08, line 172: You should do something if the `blocks_str` is not a string. Could you try to find a scenario where this is the case?

utils/visualizer/qviz/cube.py
```diff
@@ -5,7 +5,14 @@
 
 
 class Cube:
-    def __init__(self, cube_string: str, max_weight: int, element_count: int, size: int, depth: int):
+    def __init__(
+        self,
+        cube_string: str,
+        max_weight: int,
+        element_count: int,
+        size: int,
```
Jiaweihu08, line 13: You may need to modify this class because different blocks can belong to the same file.
Different blocks can belong to the same file. The `size` here will cause incorrect `sampled_size` and `total_size` in `SamplingInfo.`

utils/visualizer/qviz/content_loader.py
jorgeMarin1: I kept stats, because there we have the tags which store the blocks of the cubes

utils/visualizer/qviz/content_loader.py
jorgeMarin1: I've checked with three different tables and in all of them blocks_str is a string... maybe because of how the delta table is defined it will always be a string.

jorgeMarin1: About doing something if blocks_str is not a string... to avoid repeating code I will simply make sure before entering the if that it is a string

utils/visualizer/qviz/cube.py
jorgeMarin1: Why would it cause an incorrect sampled_size and total_size?

Right now when we extract the cubes from a delta table, we read each block and:

- **If we already have a cube with the block's cubeID** the size associated to that block is added to the cube's size.
- **If we don't have a cube with that cubeID** we create a new one and it's size is defined as the size associated to that block.

utils/visualizer/qviz/content_loader.py
osopardo1: yep, let's just complete the if else 😄

utils/visualizer/qviz/cube.py
osopardo1: I think you are right @jorgeMarin1 , if we update the cube parameters by processing each block separately, we are going to get the correct total size and sampled size. 

Let me just test it locally and see that the information is correct 👍

utils/visualizer/tests/content_loader_test.py
```diff
@@ -7,7 +7,7 @@
     addFiles_from_checkpoint_file,
     addFiles_from_json_log_files,
     extract_checkpoint_version,
-    process_table_delta_log
+    process_table_delta_log,
```
osopardo1, line 10: is this comma necessary?

jorgeMarin1: Right now I can't recall why I added that comma. Maybe I added it because I had some error when I ran a test.

However, correct me if I'm wrong, but I think this file `content_loader_test` is used to test the methods I deleted when I updated `content_loader.py`.

So maybe I should delete these files and change the name of the test file I created to something like `delta_table_process.py`

utils/visualizer/tests/drawing_elements_test.py
```diff
@@ -5,26 +5,58 @@
...
-                    "dataType": "IntegerDataType"
+                    "dataType": "IntegerDataType",
                 },
                 {
                     "className": "io.qbeast.core.transform.LinearTransformer",
                     "columnName": "price",
-                    "dataType": "DoubleDataType"
-                }
+                    "dataType": "DoubleDataType",
+                },
```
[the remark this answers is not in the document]
jorgeMarin1: Just like I mentioned before, these file is used to test methods we don't longer need, should I delete this file too?

utils/visualizer/pyproject.toml
```diff
@@ -25,6 +25,9 @@ Werkzeug = "2.2.3"
 zipp = "3.8.0"
 setuptools = "65.5.1"
 pyarrow = "~17.0.0"
+pandas = "2.1.1"
+numpy = "^1.24"
+deltalake = "^0.8.0"
```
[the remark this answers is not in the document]
jorgeMarin1: According to the [delta lake website](https://docs.delta.io/latest/releases.html) we need version `0.7.x and 0.8.x` of Delta Lake for version `3.0.x `of Apache Spark.

So shouldn't it be right?

utils/visualizer/tests/drawing_elements_test.py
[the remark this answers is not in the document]
osopardo1: uhuhhh, yes, let's delete everything that is unnecessary now

utils/visualizer/pyproject.toml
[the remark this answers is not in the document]
jorgeMarin1: Ok, fixed, thanks!

Issue #398, opened 2024-09-04
Qbeast Spark Refactor: Delta abstraction


osopardo1:
We noticed that some parts of the core/common code are still dependent of the Delta Format, losing its intention of sharing code between implementations.

One task to make the code more flexible is splitting the Qbeast Internals (such as the `RollUpWriter`) from their dependencies.

osopardo1:
Here are some more ideas about the changes:

- We should decouple the code in three layers:
    - **Engine** (spark): this is where all the heavy work happens: how we write and index the data, execute optimizations... It should include all the Datasource utilities for registering the table/dataset, although this part would eventually disappear to a more "optionable" way of writing with Qbeast (perhaps included in the third layer).
    - **Core**: this package should contain all the interfaces for refining the Layout in a format-agnostic manner. Instead of relying on `FileAction` or Delta's `Snapshot`, we should define Qbeast own API's for retrieving and managing metadata and layout. The main entry point can be the `QbeastSnapshot`. It is complicated to detach the core module from the Spark engine, since the DataFrame/Dataset API is very powerful to manipulate both data and metadata. 
    - **Format** (delta): this last layer should contain the implementations of `core` that are dependent of the Format. For example, the Metadata Commit Protocol and the encapsulation of the transaction. As well as all the retrieval methods to build the `QbeastSnapshot`.

osopardo1:
For the `RollupDataWritter` part:

- Leave DataWriter method and interface simple: just write a DataFrame in files according to the TableChanges and produce a sequence of IndexFile with metadata as a result.
- Move the `Rollup` code to the `OTreeSparkManager`, more precisely into the optimize method.

cugni:
I would organize the 3 modules slightly  differently: 
- **core** contains all the algorithms and various computations using a spark as an engine. We should use only Dataframes and SparkSession, but not internal Spark API. For instance, here I would put:
   -  the data model
   - The OTreeAnalyzer
   - RollupDataWriter
   - OpenTable API (QbeastSnapshot, etc.). 

- **Format** modules (Delta/Hudi/Iceberg): the adapters of any open table format. These modules should only depend on the core. 

- **Spark integration**: this module binds everything together, registers the data source and SQL extension, and provides any User interface. For instance:
    -  DataSources
    -  SQLextensions
    -  User interface (e.g., QbeastTable)
    -  Any "glue" code, that connects different modules. 
    - Anything that depends on Spark internals. 
    
```mermaid
[...]

PR #402, opened 2024-09-06
DataWriter refactory

cugni:
## Description
This PR is working toward the abstraction of Qbeast-spark from Delta code #398. It makes the RollupDataWrite more independent of the Open Table Format used and reorganizes the roles of accessing the data between the DataWriter and the QbeastSnapshot. This PR also solves a potential issue when optimizing a table with a changed schema.  

## Checklist:

- [x] Add tests.


build.sbt
```diff
@@ -102,7 +102,7 @@ ThisBuild / publishTo := {
 }
 
 // Sonatype settings
-ThisBuild / publishMavenStyle := true
+ThisBuild / publishMavenStyle.withRank(KeyRanks.Invisible) := true
```
osopardo1, line 105: what is this? not familiar...

osopardo1, line 105: what is this for? not familiar...

spark-warehouse/student/_delta_log/00000000000000000000.json
```diff
@@ -0,0 +1,3 @@
+{"commitInfo":{"timestamp":1725565486539,"operation":"WRITE","operationParameters":{"mode":"Overwrite","partitionBy":"[]"},"isolationLevel":"Serializable","isBlindAppend":false,"operationMetrics":{"numFiles":"1","numOutputRows":"0","numOutputBytes":"457"},"engineInfo":"Apache-Spark/3.5.0 Delta-Lake/3.1.0","txnId":"e3a74474-1ac4-4660-aa67-084f96dd16e2"}}
```
osopardo1, line 1: Should this `spark-warehouse/student` files be uploaded? Seems like some test wrote them and weren't deleted

src/main/scala/io/qbeast/core/model/QbeastSnapshot.scala
```diff
@@ -111,4 +112,6 @@ trait QbeastSnapshot {
    */
   def loadRevisionAt(timestamp: Long): Revision
 
+  def loadDataframeFromData(indexFile: Dataset[IndexFile]): DataFrame
```
osopardo1, line 115: I think it's best if we rename it as `loadDataFrameFromFiles`

src/main/scala/io/qbeast/spark/delta/writer/BlockWriter.scala
```diff
@@ -169,7 +169,7 @@ case class BlockWriter(
       .beginBlock()
       .setCubeId(cubeId)
       .setMaxWeight(maxWeight)
-      .setReplicated(state == State.ANNOUNCED || state == State.REPLICATED)
+      .setReplicated(state == CubeState.ANNOUNCED)
```
osopardo1, line 172: Maybe we can change the `setReplicated` method to `setAnnounced`

src/main/scala/io/qbeast/spark/table/IndexedTable.scala
```diff
@@ -496,10 +498,19 @@ private[table] class IndexedTableImpl(
...
+          val removeFiles =
+            indexFiles.map(IndexFiles.toRemoveFile(dataChange = false)).collect().toIndexedSeq
+
+          val data = snapshot.loadDataframeFromData(indexFiles)
+
+          val (dataExtended, tableChanges) =
+            DoublePassOTreeDataAnalyzer.analyzeOptimize(data, indexStatus)
+
+          val newFiles = dataWriter.write(tableID, schema, dataExtended, tableChanges)
+          dataExtended.unpersist()
```
osopardo1, line 512: Where do you `persist` the dataFrame? I would suggest to keep unpersist and persist operation in the same method

src/test/scala/io/qbeast/spark/delta/QbeastSnapshotTest.scala
```diff
@@ -176,4 +181,135 @@ class QbeastSnapshotTest extends QbeastIntegrationTestSpec {
         }
     }
 
+  "loadDataframeFromData" should " correct the a table from a list of files" in {
```
osopardo1, line 184: change name of the test if you decide to use `loadDataframeFromFiles`

osopardo1 (commented):
thanks for reducing the refactor PR size :)
few comments:

spark-warehouse/student/_delta_log/00000000000000000000.json
cugni: That's a good catch. To be honest, I don't understand why we created this folder during the project test. Can we move it to tmp and make it delete it automatically?

build.sbt
cugni: Remove the annoying warning that we got for years (This is the fix suggested by the warning).

src/main/scala/io/qbeast/spark/table/IndexedTable.scala
cugni: if we unpersist a dataframe before the action is called, it is useless, as the computation will run on a uncached dataset.

src/main/scala/io/qbeast/spark/delta/writer/BlockWriter.scala
cugni: I think that all the Replicated/announced parts need a full review. I prefer to leave this to a new PR, as it is widespread in many places.

spark-warehouse/student/_delta_log/00000000000000000000.json
osopardo1: It shouldn't be created... All the tests should use the `tmpWarehouse` proposed by the `withSparkAndTmpWarehouse`. 
Probably you are using somewhere the `withSparkAndTmpDir` and it uses the current directory as a default table warehouse.

src/main/scala/io/qbeast/spark/delta/writer/BlockWriter.scala
osopardo1: Fair enough, would open a ticket 👍

spark-warehouse/student/_delta_log/00000000000000000000.json
cugni: Then it is best to delete the withSparkAndTmpDir and keep only the withSparkAndTmpWarehouse.

src/main/scala/io/qbeast/spark/table/IndexedTable.scala
osopardo1: mm true. But it can be tricky to have a `persist` somewhere in the code that you might not be aware. Perhaps dividing more the code would be a solution. But we can leave it for a second step. 

Also, another question: the `persist` is used only for performance improvement, or to avoid that some recomputation on the data affects the final result?

cugni: In the optimization, we are calculating twice the cubed of each element:
1. A first time to compute the new block size with the updated weight, 
2. A second time to write the data. 

I saw that in the code, we are using profusely the broadcast, but we never unpersist them, which is not ideal as we are using memory that might be better used for other scopes. I would keep this code as it is and open an Issue when describing the problem to find a more general solution.

osopardo1:
FYI @cugni : I've opened a new issue with a more fine-grained scope than the Delta Refactor, and I've also changed the PR title with the Issue number linked.

src/main/scala/io/qbeast/spark/table/IndexedTable.scala
osopardo1: Seems good. Just add a line on the code referencing the new issue and the problem we will solve.

PR #411, opened 2024-09-12
Issue #417: Abstract the Qbeast Snapshot Module

JosepSampe:
## Description

This is a small initial part of issue #398.

The aim of this PR is to allow the `DeltaQbeastSnapshot` to accept the `tableID` as an input parameter instead of a Delta `snapshot`. This will make the top-level `QbeastSnapshot` class more abstract and enable better integration with other table formats.

Related tests have also been adapted to remove the strict dependency on `org.apache.spark.sql.delta.DeltaLog`. In the future, the tests will instantiate a `QbeastSnapshot` instead of a `DeltaQbeastSnapshot` to make them more generic.

## Type of change

Refactoring

## Checklist:

- [ ] Add tests.


src/main/scala/io/qbeast/core/model/QbeastSnapshot.scala
```diff
@@ -24,12 +26,26 @@ import org.apache.spark.sql.Dataset
  */
 trait QbeastSnapshot {
 
+  /**
+   * Returns the current snapshot object, representing the dataset state.
+   */
+  def snapshot: Any
```
cugni, line 32: I don't like this returning an any. Do we need to return the underlying snapshot type? I think it is best if we don't declare this method in the trait. 
On the hand, if we really have to do it, it is best to keep it as generic 
```
trait QbeastSnapstho[T] {
def snapshot:T
```

But this makes only sense if we are confident we have similar "snapshot" objects in all 3 open table formats.

src/main/scala/io/qbeast/core/model/QbeastSnapshot.scala
```diff
@@ -24,12 +26,26 @@ import org.apache.spark.sql.Dataset
...
 
+  /**
+   * Returns the total number of data files in the snapshot.
+   */
+  def allFilesCount: Long
+
+  /**
+   * Provides the schema of the dataset for this snapshot.
+   */
+  def schema: StructType
```
cugni, line 47: It is good to return StructType instead of general Schema types, as we have parametrized in the in the QbeastCoreContext[DATA, DataSchema, QbeastOptions, FileDescriptor]. Maybe is best to change also the QbeastCoreContext and set Data as Dataframe and DataSchema to StructType, for coherence. ..
Or we do the opposite, and you make this class parametrized

cugni (approved):
Just a few comments. I like that this PR also solve another issue, when we were reloading the deltaLog multiple times in the IndexTables, instead of only once for each QbeastStapshot

src/main/scala/io/qbeast/core/model/QbeastSnapshot.scala
JosepSampe: I will address this in an upcoming PR, as it will require more file changes, and I want to avoid this issue growing over time

build.sbt
```diff
@@ -45,6 +45,7 @@ ThisBuild / libraryDependencies ++= Seq(
   mockito % Test)
 
 Test / javaOptions ++= Seq("-Xmx10G", "-XX:+UseG1GC")
+Test / testOptions += Tests.Argument("-oD")
```
osopardo1, line 48: What is this?

src/test/scala/io/qbeast/spark/delta/QbeastSnapshotTest.scala
```diff
@@ -194,20 +187,16 @@ class QbeastSnapshotTest extends QbeastIntegrationTestSpec {
...
         .save(tmpDir)
 
-      val deltaLog = DeltaLog.forTable(spark, tmpDir)
       import spark.implicits._
 
-      val qbeastSnapshot = DeltaQbeastSnapshot(deltaLog.update())
+      val qbeastSnapshot = getQbeastSnapshot(tmpDir)
       val lastRev = qbeastSnapshot.loadLatestRevision
-      val files = qbeastSnapshot.loadRevisionFiles(lastRev.revisionID)
+      val filesToOptim = qbeastSnapshot.loadIndexFiles(lastRev.revisionID)
```
osopardo1, line 194: better to explicitly put the full name: `filesToOptimize`

src/test/scala/io/qbeast/spark/delta/QbeastSnapshotTest.scala
```diff
@@ -249,16 +238,11 @@ class QbeastSnapshotTest extends QbeastIntegrationTestSpec {
...
+      val qbeastSnapshot = getQbeastSnapshot(tmpDir)
 
-      val lastRev = qbeastSnapshot.loadLatestRevision
-      val files = qbeastSnapshot.loadAllRevisions
-        .map(rev => qbeastSnapshot.loadRevisionFiles(rev.revisionID))
-        .foldLeft(spark.emptyDataset[AddFile])(_ union _)
-      import spark.implicits._
-      val filesToOptimize: Dataset[IndexFile] =
-        files.map(IndexFiles.fromAddFile(lastRev.columnTransformers.size))
+      val filesToOptimize = qbeastSnapshot.loadAllRevisions
```
osopardo1, line 243: should we move the `filesToOptimize` code into a method? Make it more testable this way

src/test/scala/io/qbeast/spark/delta/QbeastSnapshotTest.scala
```diff
@@ -297,14 +281,9 @@ class QbeastSnapshotTest extends QbeastIntegrationTestSpec {
...
-        val deltaLog = DeltaLog.forTable(spark, tmpDir)
-        val qbeastSnapshot = DeltaQbeastSnapshot(deltaLog.update())
+        val qbeastSnapshot = getQbeastSnapshot(tmpDir)
         val lastRev = qbeastSnapshot.loadLatestRevision
-        val files = qbeastSnapshot.loadRevisionFiles(lastRev.revisionID)
-
-        import spark.implicits._
-        val filesToOptim =
-          files.map(IndexFiles.fromAddFile(lastRev.columnTransformers.size))
+        val filesToOptim = qbeastSnapshot.loadIndexFiles(lastRev.revisionID)
```
osopardo1, line 286: same here, `filesToOptimize`

src/main/scala/io/qbeast/spark/delta/DeltaQbeastSnapshot.scala
```diff
@@ -21,21 +21,25 @@ import io.qbeast.spark.utils.TagColumns
...
+ * @param tableID
+ *   the table ID
  */
-case class DeltaQbeastSnapshot(protected override val snapshot: Snapshot)
-    extends QbeastSnapshot
-    with DeltaStagingUtils {
+case class DeltaQbeastSnapshot(tableID: QTableID) extends QbeastSnapshot with DeltaStagingUtils {
+
+  override val snapshot: Snapshot =
+    DeltaLog.forTable(SparkSession.active, tableID.id).unsafeVolatileSnapshot
```
osopardo1, line 42: Perhaps is best to use the `update() `method to retrieve the `Snapshot` from Delta instead of the `unsafeVolatileSnapshot`

osopardo1 (changes requested):
Before merging and approving the PR, I would ask to add an issue with the scope and link the specific number instead of #398 🙏 

Also, some other comments to be resolved:

build.sbt
JosepSampe: This shows the total time spent on each individual test. Since tests take a long time to execute, I was curious to know which tests take more time and, potentially in the future, investigate if we can improve them.

src/test/scala/io/qbeast/spark/delta/QbeastSnapshotTest.scala
JosepSampe: For now, I would keep it like this since it is only called twice. I think we can consider, in the future, doing a pass through all the tests in a separate issue to see if we can improve them in some way.

build.sbt
osopardo1: Top 👍

osopardo1 (approved):

Issue #338, opened 2024-06-25
Implement HistogramTransformation


osopardo1:
After analyzing the efficiency of distribution functions for indexing (see issue #336 ), we can start implementing the HistogramTransformation.

The idea is to build it as another type of transformation, and eventually turn it into the default. 

The API can be something like:

```scala
df.write.format("qbeast").option("columnsToIndex", "id:histogram").save("/tmp/test-histogram")
```

And under the hood:

```scala

case class HistogramTransformation(hist: IndexedSeq[String]) extends Transformation {

  override def transform(value: Any): Double = ???

  /**
   * This method should determine if the new data will cause the creation of a new revision.
   *
   * @param newTransformation
   *   the new transformation created with statistics over the new data
   * @return
   *   true if the domain of the newTransformation is not fully contained in this one.
   */
  override def isSupersededBy(newTransformation: Transformation): Boolean = ???

  /**
   * Merges two transformations. The domain of the resulting transformation is the union of this
   *
   * @param other
   * @return
   *   a new Transformation that contains both this and other.
   */
  override def merge(other: Transformation): Transformation = ???
}

object HistogramTransformation {
  def apply(hist: IndexedSeq[String]): HistogramTransformation = new HistogramTransformation(hist)
}


```

We would take advantage of the first step of `OTreeDataAnalyzer` and compute an approximate histogram or quartiles of the columns specified. 

```scala
  /**
   * Analyze a specific group of columns of the dataframe and extract valuable statistics
   * @param data
   *   the data to analyze
   * @param columnTransformers
   *   the columns to analyze
   * @return
   */
  private[index] def getDataFrameStats(
      data: DataFrame,
      columnTransformers: IISeq[Transformer]): Row = {
    val columnStats = columnTransformers.map(_.stats)
    val columnsExpr = columnStats.flatMap(_.statsSqlPredicates)
    data.selectExpr(columnsExpr ++ Seq("count(1) AS count"): _*).first()
  }

```

osopardo1:
I am changing the title from Histogram to Approximate Quantile. Would be more adequate to use this method instead of the histogram to infer new statistics to the table and to transform the rows at writing time.

osopardo1:
Before implementing, I will prepare a document regarding the existing algorithms for the Approximate Quantiles. 

[...]

Issue #223, opened 2023-10-26
Refactor Revision (flow) and Metadata Changes 


osopardo1:
Right now, Revision Changes are being treated in three different parts:

1. `SparkRevisionFactory` -> creates a Revision with user configurations (`columnsToIndex`, `columnStats`...)
2. `OTreeAnalyzer` -> analyses the data and triggers a new Revision if supersedes the existing one, or if the user input is does not contain enough information (such as `columnStats` that initialises the `transformations`).
3. `MetadataWriter` -> commits the desired Revision Changes into the Table Metadata. 

Each one of the components works independently and without visibility about how the Revision is triggered.

That's why a lot of bugs appear in certain conditions (such as **appending to an empty table**) that affect only one of the processes. We need to:

- [ ] Analyze the current status.
- [ ] Design a better flow of information. (One option could be to pass some information around through `options` (a Map of Strings)
- [ ] Refactor the corresponding processes with new information treatment.

PR #413, opened 2024-09-13
Issue #416: Add CDFQuantile Transformers and Transformations

osopardo1:
## Description

Adds #416 (prior #338)

- Adds interfaces for `CDFQuantileTransformer`and `CDFQuantileTransformation`.
- Changes  `StringHistogramTransformation`for `CDFStringQuantileTransformer`.
- Changes `StringHistogramTransformation` for `CDFStringQuantileTransformation`.
- Adds `CDFNumericQuantileTransformer` and `CDFNumericQuantileTransformation`.
- Adds `ManualPlaceholderTransformation` and `ManualColumnStats` to control how the transformations are initialized. 
- **The quantiles SHOULD be added through `columnStats`**.
- **The `number of quantiles`** in the `QbeastUtils` method, would be 50 by default.
- **The` relative error` for quantiles calculation** in `QbeastUtils` method, would be 0.1 by default.
- For **triggering new Revision** (changing index parameters), we can use **a different set of quantiles in the `columnStats`** option.

### API

1. Compute the Quantiles of a column (strictly as Numeric or String) with `QbeastUtils.computeQuantilesForColumn` method:
```scala
  /**
   * Compute the quantiles for a given column
   * ...
   * @param df
   *   DataFrame
   * @param columnName
   *   Column name
   * @param numberOfQuantiles
   *   Number of Quantiles, default is 50
   * @param relativeError
   *   Relative Error, default is 0.1
   * @return
   */
  def computeQuantilesForColumn(
      df: DataFrame,
      columnName: String,
      numberOfQuantiles: Int = 50,
      relativeError: Double = 0.1)
```

```scala
val df = spark.range(0, 100).toDF("a")
val columnName = "a"
val columnQuantiles =
  QbeastUtils.computeQuantilesForColumn(df = df, columnName = columnName)
val columnQuantilesNumberOfQuantiles =   
  QbeastUtils.computeQuantilesForColumn(df = df, columnName = columnName, numberOfQuantiles = 100)
val columnQuantilesRelativeError =   
  QbeastUtils.computeQuantilesForColumn(df = df, columnName = columnName, relativeError = 0.3)
val columnQuantilesNumAndError =
  QbeastUtils.computeQuantilesForColumn(df = df, columnName = columnName, numberOfQuantiles = 100, relativeError = 0.3)

```

2. Index the data indicating the column type to index and the columnStats option.

```scala
df.write
  .mode("overwrite")
  .format("qbeast")
  .option("cubeSize", "30000")
  .option("columnsToIndex", s"$columnName:quantiles")
  .option("columnStats", s"""{"${columnName}_quantiles":$columnQuantiles}""")
  .save("/tmp/test-quantiles")
```

#### Example with default configuration:

```scala

import io.qbeast.spark.utils.QbeastUtils
import spark.implicits._

val df = spark.range(0, 100).toDF("a")
val columnName = "a"
val columnQuantiles =
  QbeastUtils.computeQuantilesForColumn(df, columnName)

df.write
  .mode("overwrite")
  .format("qbeast")
  .option("cubeSize", "30000")
  .option("columnsToIndex", s"$columnName:quantiles")
  .option("columnStats", s"""{"${columnName}_quantiles":$columnQuantiles}""")
  .save("/tmp/test-quantiles")
```
> Be mindful that **this interface is evolving and might change in the near future**.

#### Trigger updated quantiles

To trigger an update of the Revision with a new set of quantiles (when the data is changing distribution drastically), would be enough to add a different set of `columnStats`.

## Type of change

Developing a new feature.

## Checklist:

- [x] Add tests.

Unit test are in the package `io.qbeast.core.transform` under `QuantileTransformationTest` and `QuantileTransformerTest`.

For integration testing, we have:
-` io.qbeast.spark.index.model.transformer.QuantileTransformerIndexingTest` that does checks of the final index distribution (TODO)
-` io.qbeast.spark.index.SparkRevisionFactoryTest` for testing the correct creation of the transformations in a Revision.

Tested with the NYC Taxi Dataset. This is the output index metrics for an index using Quantiles vs the default linear indexing.

#### Metrics with Linear Transformation

```bash
OTree Index Metrics:
revisionId: 1
elementCount: 77966324
dimensionCount: 4
desiredCubeSize: 3248596
indexingColumns: PULocationID:linear,DOLocationID:linear,tpep_pickup_datetime:linear,tpep_dropoff_datetime:linear
height: 5 (3)
avgFanout: 5.2 (16.0)
cubeCount: 79
blockCount: 79
fileCount: 15
bytes: 1412854170
```

#### Metrics with Quantiles Transformation

```bash
OTree Index Metrics:
revisionId: 1
elementCount: 77966324
dimensionCount: 4
desiredCubeSize: 3248596
indexingColumns: PULocationID:quantiles,DOLocationID:quantiles,tpep_pickup_datetime:quantiles,tpep_dropoff_datetime:quantiles
height: 3 (3)
avgFanout: 12.0 (16.0)
cubeCount: 109
blockCount: 109
fileCount: 9
bytes: 1398833883 
```

We can notice that the `avgFanout` and the `height` generated by the Quantiles approach are more evenly than the ones written using a Linear Transfromation.


osopardo1:
Some comments TODOs:

- It seems like `QuantileTransformation` and `HistogramTransformation` maps the values in the same way. This is because the scope of the transformation is not clear: we should use the CDF as the abstraction for both of them. 
- The only difference between one and the other is the use of strings. And this only affects the way of calculating the bins.
- The bins should be computed during the analyze part, not in an external process.
- This PR & issue would only tackle the reorganization & renaming of the code.

This is the new proposed structure:

- `CDFTransformation` and `Transformer`: Abstract class to define the parameters to save of our transformation methodology.
- `CDFQuantileTransformation` and `Transformer`: both `StringHistogram` and `Quantile` Transformations would be grouped in a single class.
- For the future, we would add other types of CDF implementations, such as Histograms or Sketches.

src/main/scala/io/qbeast/core/model/QDataType.scala
```diff
@@ -72,31 +72,49 @@ trait OrderedDataType extends QDataType {
...
 
+  val defaultQuantiles: IndexedSeq[Any]
+
 }
 
 object DoubleDataType extends OrderedDataType {
   override def name: String = "DoubleDataType"
   override val ordering: Numeric[Any] = implicitly[Numeric[Double]].asInstanceOf[Numeric[Any]]
+
+  override val defaultQuantiles: IndexedSeq[Any] =
```
osopardo1, line 83: Not sure if we should support a `defaultQuantiles` value or should we throw an error when no quantiles are specified. What do you think @cugni ?

cugni: We should raise an exception if not provided, IMHO.

src/main/scala/io/qbeast/core/model/QDataType.scala
```diff
@@ -72,31 +72,49 @@ trait OrderedDataType extends QDataType {
...
+    BigDecimal(Double.MinValue).to(BigDecimal(Double.MaxValue), BigDecimal(1e307)).map(_.toDouble)
+
 }
 
 object IntegerDataType extends OrderedDataType {
   override def name: String = "IntegerDataType"
   override val ordering: Numeric[Any] = implicitly[Numeric[Int]].asInstanceOf[Numeric[Any]]
+  override val defaultQuantiles: IndexedSeq[Any] = Int.MinValue.to(Int.MaxValue)
```
cugni, line 91: Paola, be careful, this method is creating a list with 2^32 elements. When it gets serialized into the metadata, it would be huge.

src/main/scala/io/qbeast/core/model/QDataType.scala
```diff
@@ -106,11 +124,13 @@ object StringDataType extends QDataType {
...
   override def name: String = "TimestampDataType"
   override val ordering: Numeric[Any] = implicitly[Numeric[Long]].asInstanceOf[Numeric[Any]]
+  override val defaultQuantiles: IndexedSeq[Any] = Long.MinValue.to(Long.MaxValue)
 
 }
 
 object DateDataType extends OrderedDataType {
   override def name: String = "DateDataType"
   override val ordering: Numeric[Any] = implicitly[Numeric[Long]].asInstanceOf[Numeric[Any]]
+  override val defaultQuantiles: IndexedSeq[Any] = Long.MinValue.to(Long.MaxValue)
```
cugni, line 134: this even worse, 2^64

cugni (changes requested):
Please, remove the default interface quantile, as it could give problems, and how it is implemented now, it could use a lot of memory.

src/main/scala/io/qbeast/core/transform/CDFQuantilesTransformer.scala
```diff
@@ -64,8 +70,7 @@ trait CDFQuantilesTransformer extends Transformer {
    *
    * @return
    */
-  override def stats: ColumnStats = {
-    ColumnStats(s"${columnName}_quantiles" :: Nil, Nil)
-  }
+  override def stats: ColumnStats =
```
cugni, line 73: Wouldn't be better to just return an object NoColumnStats?

cugni (approved):
Thanks, great job! Thank you for putting a bit of order in all of this :)

src/main/scala/io/qbeast/core/transform/CDFQuantilesTransformer.scala
osopardo1: Right now, since the `SparkRevisionFactory` is tight to the `ColumnStats` to resolve the `Transformation`, we still need at least the name of the column (even if they are added through the `columnStats` options).

It would be super nice to take a look into the [Revision Flow issue](https://github.com/Qbeast-io/qbeast-spark/issues/223), but not on this PR

osopardo1:
Updating the tests to work with quantiles instead of histogram nomenclature 🐒

osopardo1:
TBD: should we include #422 in this PR?

osopardo1:
What I've added is:

- Flow to discard quantile transformations if no columnStats is provided. 
- I've had to modify a bit the SparkRevisionFactory including a check to see if any Transformer requires user metadata to be specified. 
- `CDFEmptyQuantileTransformation`. Since the `OTreeDataAnalyzer` is ALWAYS calling the `makeTransfromation` method, and we do not have a strategy to calculate quantiles during the Analyze for now, we would output this empty object.
- The `CDFEmptyQuantileTransformation` would NOT supersede any other CDF-based transformation. This ensures that, once a `CDFQuantileTransfromation` is initialized with a set of quantiles,  it would stay as the main transformation until the `columnStats` are specified again.

osopardo1:
Soooo about SparkRevisionFactory. Here's the suggested pipeline:

- User provides 'columnsToIndex' options and, optionally, 'cubeSize' and 'columnStats'. 
- For each column to index, a `Transformer` is created with the specified type (if no type is specifies, we would use the default depending on the type).
- For each `Transformer`, we look for it's Manual Column Stats.
  - If no column stats are provided, and manual stats are required, we use a `ManualPlaceholderTransformation`.
  - If no column stats are provided, and no manual stats are required, we use an `EmptyTransformation`.
  - If manual column stats are provided, we create a `Transformation` with boundaries. This Transformation can be superseded if the OTreeDataAnalyzer detects a change in the stats values.


  /**
   * Creates a new revision
   *
   * The Revision is created with the provided QbeastOptions: cubeSize, columnsToIndex and
   * columnStats
   *
   *   - For each column to index, a Transformer is created.
   *   - For each Transformer, we look for manual column stats.
   *   - If no column stats are provided, and manual stats are required, we use a
   *     ManualPlaceholderTransformation.
   *   - If manual column stats are provided, we create a Transformation with boundaries.
   *   - If no column stats are provided, and no manual stats are required, we use an
   *     EmptyTransformation.
   * @param qtableID
   *   the table identifier
   * @param schema
   *   the schema
   * @param options
   *   the options
   * @return
   */
  override def createNewRevision(
      qtableID: QTableID,
      schema: StructType,
      options: QbeastOptions): Revision

cugni (approved):
Looks good, but there are many changes, thus it is hard to review.
I was afraid the new tests were going to slow down the CI, but it doesn't seem so.

docs/AdvancedConfiguration.md
```diff
@@ -167,35 +167,36 @@ val columnStats =
...
   .save(targetPath))
 ```
-This is only necessary for the first write, if not otherwise made explicit, all subsequent appends will reuse the same histogram.
-Any new custom histogram provided during `appends` forces the creation of a new `Revision`.
+This is only necessary for the first write, if not otherwise made explicit, all subsequent appends will reuse the same quantile calculation.
+Any new custom quantiles provided during `appends` forces the creation of a new `Revision`.
 
-A default **String** histogram("a" t0 "z") will be used if the use of histogram is stated(`stringColName:string_hist`) with no histogram in `columnStats`.
-The default histogram can not supersede an existing `StringHashTransformation`.
+- A default set of Quantiles (for String is "a" t0 "z", and for Numeric is percentiles based on the `[Number.MinValue, Number.MaxValue]` range) will be used if the use of histogram is stated(`stringColName:string_hist`) with no histogram in `columnStats`.
```
Jiaweihu08, line 198: Is this still true? Do we still use `string_hist` for string columns?

docs/AdvancedConfiguration.md
```diff
@@ -167,35 +167,36 @@ val columnStats =
...
+- The default Quantiles can not supersede an existing `StringHashTransformation`.
```
Jiaweihu08, line 199: This line can also be removed, as it no longer makes sense.

src/main/scala/io/qbeast/core/transform/ManualPlaceholderTransformation.scala
```diff
@@ -0,0 +1,45 @@
...
+
+  override def transform(value: Any): Double = {
+    throw new UnsupportedOperationException(
+      "ManualPlaceholderTransformation does not support transform. " +
+        s"Please provide the valid transformation of $columnName through option 'columnStats'")
+  }
+
+  override def isSupersededBy(newTransformation: Transformation): Boolean = {
+    throw new UnsupportedOperationException(
+      "ManualPlaceholderTransformation does not support superseeded by. " +
```
Jiaweihu08, line 35: "superseded"

src/main/scala/io/qbeast/spark/utils/QbeastUtils.scala
```diff
@@ -72,10 +59,76 @@ object QbeastUtils extends Logging {
...
+    approxQuantile
+  }
+
+  /**
+   * Compute the quantiles for a given column
+   *
+   * Since computing the quantiles can be expensive, this method is used outside the indexing
+   * process.
+   *
+   * It outputs the quantiles of the column as format [bin1, bin2, bin3, ...] Number of bins by
```
Jiaweihu08, line 85: The numbers returned are not bins themselves but rather the boundaries of the bins.

src/main/scala/io/qbeast/spark/index/SparkRevisionFactory.scala
```diff
@@ -38,39 +62,55 @@ object SparkRevisionFactory extends RevisionFactory[StructType, QbeastOptions] {
...
-      case None => Revision.firstRevision(qtableID, desiredCubeSize, transformers)
-      case Some(stats) =>
-        val columnStats = stats.first()
-        var shouldCreateNewSpace = true
-        val transformations = {
-          val builder = Vector.newBuilder[Transformation]
-          builder.sizeHint(transformers.size)
+    // Check if the columns to index are present in the schema
+    var shouldCreateNewSpace = true
+    val manualDefinedColumnStats = options.stats.isDefined
```
Jiaweihu08, line 67: `manuallyDefinedColumnStats` or `customColumnStats`?

docs/AdvancedConfiguration.md
osopardo1: You are right, I would remove both bullet points, as they are no longer used (we do not use any default sequence to initialize the quantiles)

docs/AdvancedConfiguration.md
```diff
@@ -167,36 +167,51 @@ val columnStats =
      |"date_max":"${formatter.format(maxTimestamp)}" }""".stripMargin
 ```
 
-## String indexing via Histograms
+## Indexing via Quantile Based CDF
```
Jiaweihu08, line 170: Maybe we should also explicitly state that the feature is still experimental.

src/main/scala/io/qbeast/spark/utils/QbeastUtils.scala
```diff
@@ -72,10 +59,77 @@ object QbeastUtils extends Logging {
...
+    quantiles.map(string => s"'$string'")
+
+  }
+
+  private def computeQuantilesForNumericColumn(
+      df: DataFrame,
+      columnName: String,
+      numberOfQuantiles: Int,
+      relativeError: Double): Array[Double] = {
+    val probabilities = (1 to numberOfQuantiles).map(_ / numberOfQuantiles.toDouble).toArray
```
Jiaweihu08, line 72: If `numberOfQuantiles` is 2, the probabilities will be `Array(0.5, 1.0).`
If `numberOfQuantiles` is 10, the probabilities will be `Array(0.1, 0.2, ..., 1.0).`
In other words, we are always leaving the minimum value out, but always including the maximum value.

docs/AdvancedConfiguration.md
osopardo1: done 👍

src/main/scala/io/qbeast/spark/utils/QbeastUtils.scala
```diff
@@ -28,39 +30,24 @@ import org.apache.spark.sql.DataFrame
...
-   * "column:histogram").option("columnStats",histogram).save()
+   * Compute the quantiles for a given column of type String
    *
    * @param df
+   *   the DataFrame
    * @param columnName
+   *   the name of the column
    */
-  def computeHistogramForColumn(df: DataFrame, columnName: String, numBins: Int = 50): String = {
+  private def computeQuantilesForStringColumn(
```
Jiaweihu08, line 40: We are using two approaches to compute the quantiles for string and numeric columns; this may need to be revised.— `MultiDimClusteringFunctions.range_partition_id` probably works for all data types.

src/main/scala/io/qbeast/spark/utils/QbeastUtils.scala
osopardo1: Should we remove as well the maximum value?

osopardo1: Discussed:
1. We include 0.0 and 1.0
2. Should be > 1
3. Do not remove duplicates on String resolution

src/main/scala/io/qbeast/spark/utils/QbeastUtils.scala
osopardo1: preferred to use a Spark method embedded in the df for numeric.

src/main/scala/io/qbeast/spark/utils/QbeastUtils.scala
```diff
@@ -28,54 +30,104 @@ import org.apache.spark.sql.DataFrame
...
     val binStarts = "__bin_starts"
     val stringPartitionColumn =
-      MultiDimClusteringFunctions.range_partition_id(col(columnName), numBins)
+      MultiDimClusteringFunctions.range_partition_id(col(columnName), numberOfQuantiles)
 
-    val histogram = df
+    val quantiles = df
       .select(columnName)
-      .distinct()
-      .na
```
Jiaweihu08, line 66: Is it ok to drop "null"s?

src/test/scala/io/qbeast/spark/index/model/transformer/CDFNumericQuantilesIndexingTest.scala
```diff
@@ -0,0 +1,116 @@
...
+
+class CDFNumericQuantilesIndexingTest
+    extends QbeastIntegrationTestSpec
+    with CDFQuantilesTestUtils {
+
+  "Quantiles index" should "allow append without specifying columnStats" in withSparkAndTmpDir(
+    (spark, tmpDir) => {
+
+      import spark.implicits._
+      val df = 1.to(10).toDF("int_col")
```
Jiaweihu08, line 15: Generate DataFrames with `val df = spark.range(1, 10).toDF(colName)` can is more efficient.

src/main/scala/io/qbeast/spark/utils/QbeastUtils.scala
osopardo1: good catch, I deleted unnecessary code lines 🤡 

I think that to mimic the behaviour of `approxQuantiles` for number, we should leave null and nans out of the calculation

src/main/scala/io/qbeast/core/transform/CDFQuantilesTransformation.scala
```diff
@@ -0,0 +1,129 @@
...
+    // If the value is null, we return 0
+    if (value == null) return 0d
+    // Otherwise, we search for the value in the quantiles
+    quantiles.search(mapValue(value)) match {
+      // First case when the index is found
+      case Found(foundIndex) => foundIndex.toDouble / (quantiles.length - 1)
+      // When the index is not found, we return the relative position of the insertion point
+      case InsertionPoint(insertionPoint) =>
+        if (insertionPoint == 0) 0d
+        else if (insertionPoint == quantiles.length + 1) 1d
```
Jiaweihu08, line 84: This was already wrong in `StringHistogramTransformation`, but the largest `InsertionPoint` is `quantiles.length`, so it can never be `quantiles.length + 1.`

src/main/scala/io/qbeast/core/transform/CDFStringQuantilesTransformation.scala
```diff
@@ -0,0 +1,116 @@
...
+}
+
+class CDFStringQuantilesTransformationDeserializer
+    extends StdDeserializer[CDFStringQuantilesTransformation](
+      classOf[CDFStringQuantilesTransformation]) {
+
+  override def deserialize(
+      p: JsonParser,
+      ctxt: DeserializationContext): CDFStringQuantilesTransformation = {
+    val histogramBuilder = IndexedSeq.newBuilder[String]
```
Jiaweihu08, line 105: Let's make sure we use a more consistent name, `quantilesBuilder` may be more accurate.

Jiaweihu08 (changes requested):
A few more comments.

src/main/scala/io/qbeast/core/transform/CDFQuantilesTransformation.scala
osopardo1: resolved!

src/main/scala/io/qbeast/core/transform/CDFStringQuantilesTransformation.scala
osopardo1: done 👍

Jiaweihu08 (approved):

Issue #418, opened 2024-09-16
Qbeast Spark Refactor: Create Missing Interfaces and Abstractions


JosepSampe:
Review the main Qbeast components, such as `StagingDataManager`, `MetadataManager`, `PreCommitHook`, and `RollupDataWriter`, etc. Identify and create any missing interfaces and abstractions to improve modularity and maintainability.

PR #421, opened 2024-09-16
Issue #418: Abstract PreCommitHook and StagingDataManager

JosepSampe:
## Description

This is a partial PR for #418 (related to #398), which introduces the missing interfaces for the PreCommitHook and StagingDataManager.

Additionally, this PR refactors the code to use IndexFile instead of delta.AddFile/RemoveFile, making the implementation more generic and flexible.

## Type of change

Code refactoring

## Checklist:

- [ ] Add tests.


src/main/scala/io/qbeast/context/QbeastContext.scala
```diff
@@ -75,7 +75,7 @@ trait QbeastContext {
  */
 object QbeastContext
     extends QbeastContext
-    with QbeastCoreContext[DataFrame, StructType, QbeastOptions, FileAction] {
+    with QbeastCoreContext[DataFrame, StructType, QbeastOptions, IndexFile] {
```
cugni, line 78: If we use the IndexFile for all internal data management, we don't need to have it parametrized, we can simplify the QbeastCoreContext assuming that we always use IndexFiles.

src/main/scala/io/qbeast/core/model/IndexFile.scala
```diff
@@ -34,7 +34,8 @@ case class IndexFile(
     size: Long,
     modificationTime: Long,
     revisionId: RevisionID,
-    blocks: IISeq[Block])
+    blocks: IISeq[Block],
+    stats: String = null)
```
cugni, line 38: Two things. First, it is best to avoid using null; use Options instead.
Second, I'm not sure about using stats as a String, as I'm not sure that the way we serialize Stats into a string is the same in Delta/Iceberg or Hudi.

If this is not the case (the 3 formats use different ways to store metadata), I think it would be best to create our own stats representation (some case classes that follow our naming convention) and then define specific adapters for each format.

src/main/scala/io/qbeast/core/model/MetadataManager.scala
```diff
@@ -62,7 +62,8 @@ trait MetadataManager[DataSchema, FileDescriptor, QbeastOptions] {
       tableID: QTableID,
       schema: DataSchema,
       options: QbeastOptions,
-      append: Boolean)(writer: => (TableChanges, IISeq[FileDescriptor])): Unit
+      append: Boolean)(
+      writer: => (TableChanges, IISeq[FileDescriptor], IISeq[FileDescriptor])): Unit
```
cugni, line 66: Why are we returning 2 lists of file descriptors? Can you update the method docs to describe it?

src/main/scala/io/qbeast/spark/delta/IndexFiles.scala
```diff
@@ -89,9 +92,18 @@ object IndexFiles {
       size = indexFile.size,
       modificationTime = indexFile.modificationTime,
       dataChange = dataChange,
+      stats = indexFile.stats,
       tags = tags)
   }
 
+  def fromRemoveFile(removeFile: RemoveFile): IndexFile = {
```
cugni, line 99: I'm not sure it is a good idea to map AddFiles and RemoveFiles to IndexFile. Why would we want to consider IndexFiles, which we use to compute the index status, the RemoveFiles? If we don't want to use the DeleteFiles, I think it is best to create our own hierarchy (e.g., trait QbeastFile, extended by IndexFile and DeletedFile).

src/main/scala/io/qbeast/core/model/MetadataManager.scala
JosepSampe: Fixed

src/main/scala/io/qbeast/context/QbeastContext.scala
JosepSampe: Fixed

src/main/scala/io/qbeast/spark/delta/IndexFiles.scala
JosepSampe: I created the trait QbeastFile, extended by IndexFile and DeletedFile classes

src/main/scala/io/qbeast/core/model/IndexFile.scala
JosepSampe: I fixed this (avoid using null), but for now I keep it as a string.
I will create a new QbeastStats representation in a separate PR

src/main/scala/io/qbeast/spark/delta/writer/BlockWriter.scala
```diff
@@ -20,7 +20,7 @@ import io.qbeast.core.model.IndexFileBuilder
 import io.qbeast.core.model.IndexFileBuilder.BlockBuilder
 import io.qbeast.core.model.TableChanges
 import io.qbeast.core.model.Weight
-import io.qbeast.spark.delta.IndexFiles
+import io.qbeast.spark.delta.QbeastFileUtils
```
cugni, line 23: Why do we need UtilsFiles? Can't we use QbeastFile as an object to provide utility methods to construct?

src/main/scala/io/qbeast/core/model/QbeastFile.scala
```diff
@@ -0,0 +1,31 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.qbeast.core.model
+
+/**
+ * QBeast File base interface
+ */
+trait QbeastFile extends Serializable {
```
cugni, line 21: I think QbeastFile would be a good candidate for using sealed traits, which gives some benefits at compiling time and pattern matching.

src/main/scala/io/qbeast/core/model/DeleteFileBuilder.scala
```diff
@@ -0,0 +1,76 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.qbeast.core.model
+
+/**
+ * Builder for creating IndexFile instances.
+ */
+final class DeleteFileBuilder {
```
cugni, line 21: Do we need a FileBuilder? It is not really a pattern in Scala, and we are using it for the IndexFiles only because we don't have all the block information when we create the file at the beginning. But if it is not the case with DeleteFiles (meaning, when we make a DeleteFile, we have path, size, and deletionTimestamp), I would avoid it.

src/main/scala/io/qbeast/spark/delta/writer/BlockWriter.scala
JosepSampe: `QbeastFileUtils` contains `delta` imports, that's why I didn't include the logic in the QbeastFile object that is placed in the `core`

cugni: ok!

src/main/scala/io/qbeast/core/model/QbeastFile.scala
JosepSampe: Changed

cugni (approved):

PR #436, opened 2024-10-09
Issue #398: Fix small overhead added during the refactoirng

JosepSampe:
## Description

- This is part of the result of the refactoring #398 validation.

With the current refactoring:
cpm_last_vs_ref

With the fix included in this PR:
cmp_fixed_qs_night

### DIFF
--- a/delta/src/main/scala/io/qbeast/spark/delta/DeltaMetadataManager.scala
+++ b/delta/src/main/scala/io/qbeast/spark/delta/DeltaMetadataManager.scala
@@ -28,8 +28,6 @@ import org.apache.spark.sql.SparkSession
  */
 object DeltaMetadataManager extends MetadataManager {
 
-  private val deltaLogCache = scala.collection.mutable.Map[QTableID, DeltaLog]()
-
   override def updateWithTransaction(
       tableID: QTableID,
       schema: StructType,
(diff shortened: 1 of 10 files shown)


delta/src/main/scala/io/qbeast/spark/delta/DeltaMetadataManager.scala
```diff
@@ -28,6 +28,8 @@ import org.apache.spark.sql.SparkSession
  */
 object DeltaMetadataManager extends MetadataManager {
 
+  private val deltaLogCache = scala.collection.mutable.Map[QTableID, DeltaLog]()
```
osopardo1, line 31: Could this structure affect the memory? If so, how much? (I guess the worst case scenario is to manage a big Catalog of tables in the same Spark Session...)

osopardo1 (commented):
Looks good, but I am afraid that the cache can cause other problems...

delta/src/main/scala/io/qbeast/spark/delta/DeltaMetadataManager.scala
JosepSampe: I removed the `deltaLogCache` as actually it was not producing the overhead

Issue #438, opened 2024-10-21
Deprecate Staging Area Manager


osopardo1:
In the current code base, we use a `StagingDataManager` component to write new data while efficiently indexing in batch.

This `StagingDataManager` was **added to avoid creating many small files when appending a small number of records**.

Every time an appends takes place, the `StagingDataManager`:
1. Loads all the Unindexed Files. (tagged as Revision ID = 0L)
2. Checks if the sum of sizes of the Unindexed Files had reached the `stagingAreaSizeInBytes`.
3. If so, it indexes all the Unindexed Files alongside the appended data. 
4. If not:
     - If the `stagingAreaSizeInBytes` is set > 0L -> append data as delta.
     - If the `stagingAreaSizeInBytes` is <= 0L -> append data as beast. 

But this `StagingAreaManager` implementation covers all the use cases that generate Unindexed Files, including:
- Small writes in Qbeast. -> This case has been solved with the RollUp feature. The RollUp merges the small cubes into their parents. In the worst-case scenario, it would create a single file with the size of the appended data.
- External Writers that append, update, or delete data of a Qbeast Table.
Large Tables that we Convert To Qbeast -> This use case can be problematic. If the `stagingAreaSizeInBytes` is set to greater than 0L, the code will try to index all the existing data at once on the next append. 

The Small Writes in Qbeast use case is the literal definition of Staging Area (a part of the table that we postpone its processing for efficiency). Since this use case is already solved by RollUp, the purpose of the Staging Area is defeated. 

This issue is to mark as Deprecated the `StagingAreaManager`, since is no longer a useful component for Qbeast Spark.

The indexing of Unindexed Files should be treated entirely by the Optimize command (issue number #294).

osopardo1:
We remove it directly on #487 . It did not show any usage and was an interfering component for any Multi-Format support.

PR #440, opened 2024-10-22
Issue #294: Optimization of Unindexed Files [Staging Area]

osopardo1:
## Description

Adds #294 

## Type of change

New Feature. The Unindexed Files of a Qbeast Table were only optimizable from the `StagingDataManager` component. After thinking about structure and use cases, we noticed that the Staging Area has lost its original purpose (check issue #438) and that we should treat Indexed and Unindexed File separately from the Append execution. 

For that reason, we extend the interface of optimization to enable the processing of the unindexed files too. 

#### API

```scala
import io.qbeast.spark.QbeastTable

val qbeastTable = QbeastTable.forPath(spark, "/path")
qbeastTable.optimize(revisionID = 0L, fraction = <fraction_to_optimize>)
```
- `revisionID = 0L` is the selected Revision ID for Unindexed Files.
- `fraction`: Any number from 0.0 to 1.0 that we want to optimize. By default is 1.0. If it's a recently converted table with Qbeast, and contains a lot of legacy data, we suggest reducing the fraction to optimize and doing the operation in batches. 

> WARNING: each time we execute optimize() for the Unindexed Files it would calculate the bytes to optimize from the current state. If some files had already been indexed, they would not be part of the second iteration. 

#### Implementation

1. Load the `QbeastSnapshot` of the Table.
2. Read the list of Unindexed Files from the `QbeastSnapshot`.
3. Select files til `fraction * totalBytes` threshold is reached.
4. Apply indexing and roll-up to the Data.
5. Write the data in files.
6. In the same transaction: (this step should be done internally as Qbeast Spark, since we have more control of Add Files, Delete Files, and transaction open/close. )
    1. mark the old files as Deleted.
    2. Add the new file entries.

## Checklist:

- [x] Add tests.

This has been tested locally with: `QbeastOptimizationIntegrationTest`.

I've added four cases:
1. Optimization of a Converted Table. (All data is unindexed).
2. Optimization of a Hybrid Table after Append. (Some data is unindexed after an external append).
3. Optimization of a fraction of a Hybrid Table. (Do not optimize all the Unindexed Files at once). 

**Test Configuration**:
* Spark Version: 3.5.0
* Hadoop Version: 3.3.4
* Cluster or local? Local

### DIFF
--- a/docs/QbeastTable.md
+++ b/docs/QbeastTable.md
@@ -31,11 +31,11 @@ qbeatsTable.lastRevisionID() // the last Revision identifier
 Through `QbeastTable` you can also execute the `Optimize` operation. This command is used to **rearrange the files in the table** according to the dictates of the index to render queries more efficient. 
 
 #### Paramaters of Manual Optimization
-| Parameter    | Description                                                                                               | Default            |
-|--------------|-----------------------------------------------------------------------------------------------------------|--------------------|
-| `revisionID` | The revision number you want to optimize.                                                                 | Latest revision    |
-| `fraction`   | The fraction of the data of the specified revision you want to optimize.                                  | None specified     |
-| `options`    | A map of options for optimization. You can specify `userMetadata` and configurations for `PreCommitHook`. | None specified     |
+| Parameter    | Description                                                                                               | Default         |
+|--------------|-----------------------------------------------------------------------------------------------------------|-----------------|
+| `revisionID` | The revision number you want to optimize.                                                                 | Latest revision |
+| `fraction`   | The fraction of the data of the specified revision you want to optimize.                                  | 1.0             |
+| `options`    | A map of options for optimization. You can specify `userMetadata` and configurations for `PreCommitHook`. | None specified  |
 
 
 #### Examples of Manual Optimization
--- a/docs/QbeastTable.md
+++ b/docs/QbeastTable.md
@@ -26,13 +26,19 @@ qbeatsTable.lastRevisionID() // the last Revision identifier
 ```
 
 ## Table Operations
+
+### Optimization
 Through `QbeastTable` you can also execute the `Optimize` operation. This command is used to **rearrange the files in the table** according to the dictates of the index to render queries more efficient. 
 
-These are a few different ways of executing the `optimize` operation, with the input parameters being:
-1. `revisionID`: The revision number you want to optimize, the default is the latest revision.
-2. `fraction`: The fraction of the data of the specified revision you want to optimize.
-3. `options`: A map of options of the optimization. You can specify the `userMetadata` as well as configurations for `io.qbeast.spark.delta.hook.PreCommitHook.`
+#### Paramaters of Manual Optimization
+| Parameter    | Description                                                                                               | Default            |
+|--------------|-----------------------------------------------------------------------------------------------------------|--------------------|
+| `revisionID` | The revision number you want to optimize.                                                                 | Latest revision    |
+| `fraction`   | The fraction of the data of the specified revision you want to optimize.                                  | None specified     |
+| `options`    | A map of options for optimization. You can specify `userMetadata` and configurations for `PreCommitHook`. | None specified     |
 
+
+#### Examples of Manual Optimization
 ```scala
 // Optimizing 10% of the data from Revision number 2, and stores some user metadata
 qbeastTable.optmize(2L, 0.1, Map["userMetadata" -> "user-metadata-for-optimization"])
@@ -45,6 +51,25 @@ qbeastTable.optimize()
 
 // Optimizes the specific files
 qbeastTable.optimize(Seq("file1", "file2"))
+```
+
+### Optimization of Unindexed Files
+
+There are some use cases in which a Table could have several **Unindexed Files**.
+- **Staging Data**: Enabling the Staging Area gives the possibility to **ingest data without indexing it**. Since very small appends could produce overhead during the write process, the new data would be commited to the table without reorganization. Every time the staging are size is reached, the data is indexed using the latest state of the Table. 
+- **Table Converted To Qbeast**: An existing `parquet` or `delta` Table can be converted to a `qbeast` Table through the `ConvertToQbeastCommand`. Since the table can be very big, the conversion only adds a metadata commit to the Log, indicating that from that point onwards the appends would be indexed with Qbeast.
+- **External Table Writers**: External writers can write data to the table in the underlying format (delta, hudi or iceberg)
+
+All the sets of Unindexed Files are mapped to a revision number 0. For manually indexing these files, you can use the `optimize` method with the `revisionId` parameter set to 0.
+
+#### Examples of Manual Optimization of Unindexed Files
+```scala
+qbeastTable.optimize(revisionId = 0L)
+
+// If the table is very large, 
+// we recommend to use the fraction configuration 
+// to decide the percentage of unindexed ddta to optimize
+qbeastTable.optimize(revisionId = 0L, fraction = 0.5)
 ```
 
 ## Index Metrics
--- a/docs/QbeastTable.md
+++ b/docs/QbeastTable.md
@@ -41,7 +41,7 @@ Through `QbeastTable` you can also execute the `Optimize` operation. This comman
 #### Examples of Manual Optimization
 ```scala
 // Optimizing 10% of the data from Revision number 2, and stores some user metadata
-qbeastTable.optmize(2L, 0.1, Map["userMetadata" -> "user-metadata-for-optimization"])
+qbeastTable.optimize(2L, 0.1, Map["userMetadata" -> "user-metadata-for-optimization"])
 
 // Optimizing all data from a given Revision
 qbeastTable.optimize(2L)
@@ -56,11 +56,11 @@ qbeastTable.optimize(Seq("file1", "file2"))
 ### Optimization of Unindexed Files
 
 There are some use cases in which a Table could have several **Unindexed Files**.
-- **Staging Data**: Enabling the Staging Area gives the possibility to **ingest data without indexing it**. Since very small appends could produce overhead during the write process, the new data would be commited to the table without reorganization. Every time the staging are size is reached, the data is indexed using the latest state of the Table. 
-- **Table Converted To Qbeast**: An existing `parquet` or `delta` Table can be converted to a `qbeast` Table through the `ConvertToQbeastCommand`. Since the table can be very big, the conversion only adds a metadata commit to the Log, indicating that from that point onwards the appends would be indexed with Qbeast.
+- **Staging Data**: Enabling the Staging Area gives the possibility to **ingest data without indexing it**. Since very small appends could produce overhead during the write process, the new data would be commited to the table without reorganization. Every time the staging area size is reached, the data could be indexed using the latest state of the Table. 
+- **Table Converted To Qbeast**: An existing `parquet` or `delta` Table can be converted to a `qbeast` Table through the `ConvertToQbeastCommand`. Since the table can be very big, the conversion only adds a metadata commit to the Log, indicating that from that point onwards the appends could be indexed with Qbeast.
 - **External Table Writers**: External writers can write data to the table in the underlying format (delta, hudi or iceberg)
 
-All the sets of Unindexed Files are mapped to a revision number 0. For manually indexing these files, you can use the `optimize` method with the `revisionId` parameter set to 0.
+All the Unindexed Files are mapped to a revision number 0. For manually indexing these files, you can use the `optimize` method with the `revisionId` parameter set to 0.
 
 #### Examples of Manual Optimization of Unindexed Files
 ```scala
--- a/src/main/scala/io/qbeast/table/IndexedTable.scala
+++ b/src/main/scala/io/qbeast/table/IndexedTable.scala
@@ -133,23 +133,21 @@ trait IndexedTable {
   def optimize(revisionID: RevisionID, fraction: Double, options: Map[String, String]): Unit
 
   /**
-   * Optimizes the given table for a given revision
-   * @param revisionID
-   *   the identifier of revision to optimize
-   * @param options
-   *   Optimization options where user metadata and pre-commit hooks are specified.
-   */
-  def optimize(revisionID: RevisionID, options: Map[String, String]): Unit
-
-  /**
    * Optimizes the table by optimizing the data stored in the specified index files.
    *
-   * @param files
+   * @param indexFiles
    *   the index files to optimize
    * @param options
    *   Optimization options where user metadata and pre-commit hooks are specified.
    */
-  def optimize(files: Seq[String], options: Map[String, String]): Unit
+  def optimizeIndexFiles(indexFiles: Seq[String], options: Map[String, String]): Unit
+
+  /**
+   * Optimizes the table by optimizing the data stored in the specified unindexed files.
+   * @param unindexedFiles
+   * @param options
+   */
+  def optimizeUnindexedFiles(unindexedFiles: Seq[String], options: Map[String, String]): Unit
 }
 
 /**
@@ -509,10 +507,32 @@ private[table] class IndexedTableImpl(
       fraction: Double,
       options: Map[String, String]): Unit = {
     assert(fraction > 0d && fraction <= 1d)
-    val indexFiles = snapshot.loadIndexFiles(revisionID)
-    import indexFiles.sparkSession.implicits._
-    val files = indexFiles.transform(filterSamplingFiles(fraction)).map(_.path).collect()
-    optimize(files, options)
+    log.info(s"Selecting files to Optimize for Revision $revisionID")
+    // Load the Index Files for the given revision
+    val revisionFilesDF = snapshot.loadIndexFiles(revisionID)
+    import revisionFilesDF.sparkSession.implicits._
+    // Filter the Index Files by the fraction
+    if (isStaging(revisionID)) { // If the revision is Staging, we should INDEX the staged data up to the fraction
+      val bytesToOptimize = revisionFilesDF.map(_.size).collect().sum * fraction
+      logInfo(s"Total bytes to optimize in the Staging Area: $bytesToOptimize")
+      val revisionFiles = revisionFilesDF.collect()
+      var filesToOptimize = Seq.empty[IndexFile]
+      var selectedBytesToOptimize = 0L
+      for (file <- revisionFiles) {
+        if (selectedBytesToOptimize < bytesToOptimize) {
+          filesToOptimize = filesToOptimize :+ file
+          selectedBytesToOptimize += file.size
+        }
+      }
+      logInfo(s"Total number of files to optimize in the Staging Area: ${filesToOptimize.size}")
+      val filesToOptimizeNames = filesToOptimize.map(_.path)
+      optimizeUnindexedFiles(filesToOptimizeNames, options)
+    } else { // If the revision is not Staging, we should optimize the index files up to the fraction
+      val filesToOptimize = revisionFilesDF.transform(filterSamplingFiles(fraction))
+      val filesToOptimizeNames = filesToOptimize.map(_.path).collect()
+      logInfo(s"Total Number of Files To Optimize: ${filesToOptimizeNames.size}")
+      optimizeIndexFiles(filesToOptimizeNames, options)
+    }
   }
 
   private[table] def filterSamplingFiles(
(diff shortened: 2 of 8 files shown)


src/main/scala/io/qbeast/spark/QbeastTable.scala
```diff
@@ -43,17 +43,35 @@ class QbeastTable private (
...
   private def indexedTable: IndexedTable = indexedTableFactory.getIndexedTable(tableID)
 
+  /**
+   * Checks if the revision is available in the table.
+   *
+   * If the revision is not available, or is not the staging revision, an exception is thrown.
+   * @param revisionID
+   *   the revision to check
+   */
   private def checkRevisionAvailable(revisionID: RevisionID): Unit = {
```
Jiaweihu08, line 73: `qbeastSnapshot.existsRevision` will return false if `revisionID` is the staging revision?

src/main/scala/io/qbeast/spark/QbeastTable.scala
```diff
@@ -43,17 +43,35 @@ class QbeastTable private (
...
-    if (!qbeastSnapshot.existsRevision(revisionID)) {
+    if (revisionID != stagingID && !qbeastSnapshot.existsRevision(revisionID)) {
```
Jiaweihu08, line 74: Use the StagingUtil strait to asses if a revisionID is the staging revision.

src/main/scala/io/qbeast/spark/table/IndexedTable.scala
```diff
@@ -498,10 +496,32 @@ private[table] class IndexedTableImpl(
       fraction: Double,
       options: Map[String, String]): Unit = {
     assert(fraction > 0d && fraction <= 1d)
-    val indexFiles = snapshot.loadIndexFiles(revisionID)
-    import indexFiles.sparkSession.implicits._
-    val files = indexFiles.transform(filterSamplingFiles(fraction)).map(_.path).collect()
-    optimize(files, options)
+    log.info(s"Selecting files to Optimize for Revision $revisionID")
+    // Load the Index Files for the given revision
+    val revisionFilesDF = snapshot.loadIndexFiles(revisionID)
```
Jiaweihu08, line 501: It's a DS rather than a DF.

src/main/scala/io/qbeast/spark/table/IndexedTable.scala
```diff
@@ -498,10 +496,32 @@ private[table] class IndexedTableImpl(
...
+    import revisionFilesDF.sparkSession.implicits._
+    // Filter the Index Files by the fraction
+    if (isStaging(revisionID)) { // If the revision is Staging, we should INDEX the staged data up to the fraction
+      val bytesToOptimize = revisionFilesDF.map(_.size).collect().sum * fraction
```
Jiaweihu08, line 505: Use spark sql to find the value rather than collecting a sequence of Longs: `revisionFilesDS.agg(sum("size")).as[Long].first()`

Jiaweihu08: Or you can compute the total bytes and then the `bytesToOptimize` using `f` and log both values after `revisionFiles` are collected.

src/main/scala/io/qbeast/spark/table/IndexedTable.scala
```diff
@@ -498,10 +496,32 @@ private[table] class IndexedTableImpl(
...
+      logInfo(s"Total bytes to optimize in the Staging Area: $bytesToOptimize")
+      val revisionFiles = revisionFilesDF.collect()
+      var filesToOptimize = Seq.empty[IndexFile]
+      var selectedBytesToOptimize = 0L
+      for (file <- revisionFiles) {
```
Jiaweihu08, line 510: Use `foldLeft` with a Sequence Builder for `filesToOptimize`

src/main/scala/io/qbeast/spark/table/IndexedTable.scala
```diff
@@ -513,34 +533,65 @@ private[table] class IndexedTableImpl(
...
+    // 3. In a transaction, update the table with the new data
+    metadataManager.updateWithTransaction(
+      tableID,
+      schema,
+      optimizationOptions(options),
+      append = true) {
+      // Index the data with IndexManager
+      val (data, tableChanges) = indexManager.index(filesDF, indexStatus)
+      // Write the data with DataWriter
+      val newFiles = dataWriter.write(tableID, schema, data, tableChanges)
```
Jiaweihu08, line 557: We have to make sure that the `newFiles` have `dataChange = false.`
Can you confirm that's the case here?

src/main/scala/io/qbeast/spark/table/IndexedTable.scala
```diff
@@ -513,34 +533,65 @@ private[table] class IndexedTableImpl(
...
+      (tableChanges, newFiles ++ removeFiles)
+    }
+  }
 
-  override def optimize(files: Seq[String], options: Map[String, String]): Unit = {
+  override def optimizeIndexFiles(files: Seq[String], options: Map[String, String]): Unit = {
     val paths = files.toSet
     val schema = metadataManager.loadCurrentSchema(tableID)
+    // For each Revision, we should optimize the matching files
     snapshot.loadAllRevisions.foreach { revision =>
```
Jiaweihu08, line 570: Add a `filterNot` here to discard the staging revision.

src/main/scala/io/qbeast/spark/table/IndexedTable.scala
```diff
@@ -513,34 +533,65 @@ private[table] class IndexedTableImpl(
...
+      unindexedFiles: Seq[String],
+      options: Map[String, String]): Unit = {
+    if (unindexedFiles.isEmpty) return // Nothing to optimize
+    // 1. Load the files from the Staging ID (Unindexed)
+    val files =
+      snapshot.loadIndexFiles(stagingID).filter(f => unindexedFiles.contains(f.path))
+    import files.sparkSession.implicits._
+    // 2. Load the Dataframe, the index status and the schema
+    val filesDF = snapshot.loadDataframeFromIndexFiles(files)
+    val indexStatus = snapshot.loadIndexStatus(stagingID)
```
Jiaweihu08, line 546: This `IndexStatus` will be used to index that data, and it has to be that of the latest revision.

Jiaweihu08 (changes requested):
Make sure the staging data is added to the latest revision

src/main/scala/io/qbeast/spark/table/IndexedTable.scala
osopardo1: ahhh, nice catch

src/main/scala/io/qbeast/spark/table/IndexedTable.scala
osopardo1: true 🤡

src/main/scala/io/qbeast/spark/QbeastTable.scala
osopardo1: mmm, the `revisionID` value is available when extending the interface also. 

Should we keep it private?

osopardo1: i'm ok with using the `isStaging` method, though

src/main/scala/io/qbeast/spark/table/IndexedTable.scala
osopardo1: thank god you review it.. I will add tests as well

src/main/scala/io/qbeast/spark/QbeastTable.scala
osopardo1: Deleting this part

Jiaweihu08 (changes requested):

src/test/scala/io/qbeast/spark/utils/QbeastOptimizeIntegrationTest.scala
```diff
@@ -107,4 +109,182 @@ class QbeastOptimizeIntegrationTest extends QbeastIntegrationTestSpec {
...
+      val firstSnapshot = deltaLog.update()
+      val firstUnindexedFiles = firstSnapshot.allFiles.where("tags is null")
+      firstUnindexedFiles.count() shouldBe firstSnapshot.allFiles.count()
+      // Optimize the Table
+      val qt = QbeastTable.forPath(spark, tmpDir)
+      qt.optimize(0L)
+
+      // After optimization, all files from the Legacy Table should be indexed
+      val snapshot = deltaLog.update()
+      val unindexedFiles = snapshot.allFiles.where("tags is null") // no tags
```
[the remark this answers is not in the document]
osopardo1: sure!

src/test/scala/io/qbeast/spark/utils/QbeastOptimizeIntegrationTest.scala
```diff
@@ -107,4 +109,182 @@ class QbeastOptimizeIntegrationTest extends QbeastIntegrationTestSpec {
...
+    val qt = QbeastTable.forPath(spark, tmpDir)
+    qt.optimize(0L)
+
+    // After optimization, all files from the Hybrid Table should be indexed
+    val snapshot = deltaLog.update()
+    val unindexedFiles = snapshot.allFiles.where("tags is null") // no tags
+    unindexedFiles shouldBe empty
+  }
+
+  it should "optimize a table with Deletes" in withQbeastContextSparkAndTmpDir {
```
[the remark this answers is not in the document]
osopardo1: When using Delete and Update with Copy on Write, it will create new files with no tags. We should re-index them as part of the unindexed files, right?

src/main/scala/io/qbeast/spark/table/IndexedTable.scala
```diff
@@ -513,37 +534,66 @@ private[table] class IndexedTableImpl(
...
-
-          val newFiles = dataWriter.write(tableID, schema, dataExtended, tableChanges)
-          dataExtended.unpersist()
-          (tableChanges, newFiles ++ removeFiles)
-        }
+        val data = snapshot.loadDataframeFromIndexFiles(indexFiles)
+        // 3. In the same transaction
+        metadataManager
+          .updateWithTransaction(tableID, schema, optimizationOptions(options), append = true) {
+            //
```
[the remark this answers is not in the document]
osopardo1: unnecessary hehe

src/main/scala/io/qbeast/spark/table/IndexedTable.scala
```diff
@@ -498,10 +497,32 @@ private[table] class IndexedTableImpl(
...
-    import indexFiles.sparkSession.implicits._
-    val files = indexFiles.transform(filterSamplingFiles(fraction)).map(_.path).collect()
-    optimize(files, options)
+    log.info(s"Selecting files to Optimize for Revision $revisionID")
+    // Load the Index Files for the given revision
+    val revisionFilesDS = snapshot.loadIndexFiles(revisionID)
+    import revisionFilesDS.sparkSession.implicits._
+    // Filter the Index Files by the fraction
+    if (isStaging(revisionID)) { // If the revision is Staging, we should INDEX the staged data up to the fraction
+      val bytesToOptimize = revisionFilesDS.agg(sum("size")).as[Long].first() * fraction
```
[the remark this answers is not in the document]
osopardo1: yea, but you say to revert it to calculate the bytesToOptimize... do you prefer everything in the same collect() then?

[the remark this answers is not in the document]
osopardo1: I've been trying to think of a way of doing the fold left using SQL, but nothing rings the bell.... do you have other ideas?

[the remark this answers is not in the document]
osopardo1: In the meantime, I would do everything using a single collect

src/main/scala/io/qbeast/spark/table/IndexedTable.scala
Jiaweihu08: This needs to be fixed. Data writer return by default files with `dataChange = true`

src/test/scala/io/qbeast/spark/utils/QbeastOptimizeIntegrationTest.scala
```diff
@@ -107,4 +113,225 @@ class QbeastOptimizeIntegrationTest extends QbeastIntegrationTestSpec {
...
+        .save(tmpDir)
+
+      // Delete data with DeltaTable
+      val deltaTable = DeltaTable.forPath(spark, tmpDir)
+      deltaTable.delete("id > 1 and id < 5")
+
+      // Check that the number of unindexed files is not 0
+      val deltaLog = DeltaLog.forTable(spark, tmpDir)
+      val firstSnapshot = deltaLog.update()
+      val firstUnindexedFiles = firstSnapshot.allFiles.where("tags is null")
```
Jiaweihu08, line 237: Use load staging files instead of specifying `tags is null`

src/main/scala/io/qbeast/spark/table/IndexedTable.scala
```diff
@@ -132,23 +132,21 @@ trait IndexedTable {
...
-   *
-   * @param files
-   *   the index files to optimize
+   * Optimizes the table by optimizing the data stored in the specified unindexed files.
+   * @param unindexedFiles
    * @param options
-   *   Optimization options where user metadata and pre-commit hooks are specified.
    */
-  def optimize(files: Seq[String], options: Map[String, String]): Unit
+  def optimizeUnindexedFiles(unindexedFiles: Seq[String], options: Map[String, String]): Unit
```
Jiaweihu08, line 149: Another concern is the terminology; UNindexed files sound more like files that were indexed at one point but unindexed. What about `nonIndexedFiles?`

osopardo1: sure

osopardo1: In all cases I think they are Unindexed:
- For appends and legacy: the data is never indexed with Qbeast
- For updates and deletes: the content of the new file might have been indexed in the past, but now it's a different file, so it's never indexed with Qbeast

I know that the reasoning might conflict logically in point 2, but... is it convincing?

src/main/scala/io/qbeast/spark/table/IndexedTable.scala
osopardo1: done!

src/main/scala/io/qbeast/spark/table/IndexedTable.scala
osopardo1: after discussion, we decided that, since updates and deletes are outside of clarification for Unindexed, we should keep the implementation with the current naming. 

Also, Unindexed means Non-Indexed, so there should be no problem with terminology.

osopardo1:
Leaving this review and merge after #446

docs/QbeastTable.md
```diff
@@ -26,13 +26,19 @@ qbeatsTable.lastRevisionID() // the last Revision identifier
...
+|--------------|-----------------------------------------------------------------------------------------------------------|--------------------|
+| `revisionID` | The revision number you want to optimize.                                                                 | Latest revision    |
+| `fraction`   | The fraction of the data of the specified revision you want to optimize.                                  | None specified     |
+| `options`    | A map of options for optimization. You can specify `userMetadata` and configurations for `PreCommitHook`. | None specified     |
+
 
+#### Examples of Manual Optimization
 ```scala
 // Optimizing 10% of the data from Revision number 2, and stores some user metadata
 qbeastTable.optmize(2L, 0.1, Map["userMetadata" -> "user-metadata-for-optimization"])
```
Jiaweihu08, line 44: `qbeastTable.optmize(2L, 0.1, Map("userMetadata" -> "user-metadata-for-optimization"))`
I think it was a mistake I made in a previous PR. Use `Map()` instead of `Map[]`

docs/QbeastTable.md
```diff
@@ -26,13 +26,19 @@ qbeatsTable.lastRevisionID() // the last Revision identifier
...
 
-These are a few different ways of executing the `optimize` operation, with the input parameters being:
-1. `revisionID`: The revision number you want to optimize, the default is the latest revision.
-2. `fraction`: The fraction of the data of the specified revision you want to optimize.
-3. `options`: A map of options of the optimization. You can specify the `userMetadata` as well as configurations for `io.qbeast.spark.delta.hook.PreCommitHook.`
+#### Paramaters of Manual Optimization
+| Parameter    | Description                                                                                               | Default            |
+|--------------|-----------------------------------------------------------------------------------------------------------|--------------------|
+| `revisionID` | The revision number you want to optimize.                                                                 | Latest revision    |
+| `fraction`   | The fraction of the data of the specified revision you want to optimize.                                  | None specified     |
```
Jiaweihu08, line 37: The default, I think, is 100%.

docs/QbeastTable.md
```diff
@@ -47,6 +53,25 @@ qbeastTable.optimize()
 qbeastTable.optimize(Seq("file1", "file2"))
 ```
 
+### Optimization of Unindexed Files
+
+There are some use cases in which a Table could have several **Unindexed Files**.
+- **Staging Data**: Enabling the Staging Area gives the possibility to **ingest data without indexing it**. Since very small appends could produce overhead during the write process, the new data would be commited to the table without reorganization. Every time the staging are size is reached, the data is indexed using the latest state of the Table. 
+- **Table Converted To Qbeast**: An existing `parquet` or `delta` Table can be converted to a `qbeast` Table through the `ConvertToQbeastCommand`. Since the table can be very big, the conversion only adds a metadata commit to the Log, indicating that from that point onwards the appends would be indexed with Qbeast.
```
Jiaweihu08, line 60: **could** be indexed with Qbeast.**

docs/QbeastTable.md
```diff
@@ -47,6 +53,25 @@ qbeastTable.optimize()
...
+- **External Table Writers**: External writers can write data to the table in the underlying format (delta, hudi or iceberg)
+
+All the sets of Unindexed Files are mapped to a revision number 0. For manually indexing these files, you can use the `optimize` method with the `revisionId` parameter set to 0.
```
Jiaweihu08, line 63: All ~~the sets of~~ Unindexed Files are mapped to a revision number 0. For manually indexing these files, you can use the `optimize` method with the `revisionId` parameter set to 0.

src/main/scala/io/qbeast/table/IndexedTable.scala
```diff
@@ -524,45 +569,92 @@ private[table] class IndexedTableImpl(
...
   }
 
-  override def optimize(revisionID: RevisionID, options: Map[String, String]): Unit =
-    optimize(revisionID, 1.0, options)
+  override def optimizeUnindexedFiles(
+      unindexedFiles: Seq[String],
+      options: Map[String, String]): Unit = {
+    if (unindexedFiles.isEmpty) return // Nothing to optimize
+    // 1. Load the files from the Staging ID (Unindexed)
+    val files =
```
Jiaweihu08, line 577: Convert `unindexedFiles` to a Set first to reduce time complexity.

Jiaweihu08 (commented):
One more question: Are we checking if the input fractions are valid values—f in `(0, 1.0]?`

osopardo1:
> One more question: Are we checking if the input fractions are valid values—f in `(0, 1.0]?`

Yes, there's an assertion on IndexedTable `optimize` method:

```scala
  override def optimize(
      revisionID: RevisionID,
      fraction: Double,
      options: Map[String, String]): Unit = {
    assert(fraction > 0d && fraction <= 1d)
    log.info(s"Selecting Files to Optimize for Revision $revisionID")
    // Filter the Index Files by the fraction
    if (isStaging(revisionID)) { // If the revision is Staging, we should INDEX the staged data up to the fraction
      optimizeUnindexedFiles(selectUnindexedFilesToOptimize(fraction), options)
    } else { // If the revision is not Staging, we should optimize the index files up to the fraction
      optimizeIndexedFiles(selectIndexedFilesToOptimize(revisionID, fraction), options)
    }
  }
```

src/main/scala/io/qbeast/table/IndexedTable.scala
osopardo1: done in 8cab2b8

docs/QbeastTable.md
osopardo1: done in 8cab2b8

docs/QbeastTable.md
osopardo1: done in 8cab2b8

docs/QbeastTable.md
osopardo1: done in 8cab2b8

docs/QbeastTable.md
osopardo1: correct! done in 70042784

Jiaweihu08 (approved):

Issue #441, opened 2024-10-22
Handle ConcurrentAppendException in qbeast-spark


JosepSampe:
Since PR [Issue #405: DataWriter refactory #402](https://github.com/Qbeast-io/qbeast-spark/pull/402), a ConcurrentAppendException started appearing during a write() and a optimize()

```
io.delta.exceptions.ConcurrentAppendException: Files were added to the root of the table by a concurrent update. Please try the operation again.
Conflicting commit: {"timestamp":1728639379364,"operation":"WRITE","operationParameters":{"mode":Append},"readVersion":4,"isolationLevel":"Serializable","isBlindAppend":true,"operationMetrics":{"numFiles":"3","numOutputRows":"160724","numOutputBytes":"120799894"},"engineInfo":"Apache-Spark/3.5.0 Delta-Lake/3.1.0","txnId":"27fd66c8-5027-4e2e-b5e7-5611887b2f73"}
Refer to https://docs.delta.io/latest/concurrency-control.html for more details.
| at org.apache.spark.sql.delta.DeltaErrorsBase.concurrentAppendException(DeltaErrors.scala:2300)
| at org.apache.spark.sql.delta.DeltaErrorsBase.concurrentAppendException$(DeltaErrors.scala:2291)
| at org.apache.spark.sql.delta.DeltaErrors$.concurrentAppendException(DeltaErrors.scala:3203)
| at org.apache.spark.sql.delta.ConflictChecker.$anonfun$checkForAddedFilesThatShouldHaveBeenReadByCurrentTxn$1(ConflictChecker.scala:305)
| at scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.java:23)
| at org.apache.spark.sql.delta.ConflictChecker.recordTime(ConflictChecker.scala:499) 
| at org.apache.spark.sql.delta.ConflictChecker.checkForAddedFilesThatShouldHaveBeenReadByCurrentTxn(ConflictChecker.scala:276)
| at org.apache.spark.sql.delta.ConflictChecker.checkConflicts(ConflictChecker.scala:142)
| at org.apache.spark.sql.delta.OptimisticTransactionImpl.checkForConflictsAgainstVersion(OptimisticTransaction.scala:1882)
| at org.apache.spark.sql.delta.OptimisticTransactionImpl.checkForConflictsAgainstVersion$(OptimisticTransaction.scala:1872)
[...]

PR #442, opened 2024-10-22
Issue #441: Fix data change on optimize

JosepSampe:
## Description
Fixes issue #441 by setting `dataChange = False` during `optimize()`

## Type of change

Bug Fix

### DIFF
--- a/core/src/main/scala/io/qbeast/core/model/IndexFileBuilder.scala
+++ b/core/src/main/scala/io/qbeast/core/model/IndexFileBuilder.scala
@@ -23,6 +23,7 @@ import scala.collection.immutable
 final class IndexFileBuilder {
   private var path: Option[String] = None
   private var size: Long = 0L
+  private var dataChange: Boolean = true
   private var modificationTime: Long = 0L
   private var revisionId: RevisionID = 0L
   private val blocks = immutable.Seq.newBuilder[VolatileBlock]
@@ -64,6 +65,19 @@ final class IndexFileBuilder {
    */
   def setStats(stats: Option[QbeastStats]): IndexFileBuilder = {
     this.stats = stats
+    this
+  }
+
+  /**
+   * Sets the data change.
+   *
+   * @param dataChange
+   *   whether this index file represents a data change
+   * @return
+   *   this instance
+   */
+  def setDataChange(dataChange: Boolean): IndexFileBuilder = {
+    this.dataChange = dataChange
     this
   }
 
@@ -117,6 +131,7 @@ final class IndexFileBuilder {
     IndexFile(
       filePath,
       size,
+      dataChange,
       modificationTime,
       revisionId,
       blocks.result().map(_.toBlock(filePath)),
--- a/delta/src/main/scala/io/qbeast/spark/delta/DeltaMetadataWriter.scala
+++ b/delta/src/main/scala/io/qbeast/spark/delta/DeltaMetadataWriter.scala
@@ -176,8 +176,8 @@ private[delta] case class DeltaMetadataWriter(
 
       // Execute write
       val (tableChanges, indexFiles, deleteFiles) = writer
-      val addFiles = indexFiles.map(DeltaQbeastFileUtils.toAddFile(dataChange = true))
-      val removeFiles = deleteFiles.map(DeltaQbeastFileUtils.toRemoveFile(dataChange = false))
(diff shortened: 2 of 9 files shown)


core/src/main/scala/io/qbeast/core/model/IndexFileBuilder.scala
```diff
@@ -23,6 +23,7 @@ import scala.collection.immutable
 final class IndexFileBuilder {
   private var path: Option[String] = None
   private var size: Long = 0L
+  private var dataChange: Boolean = true
```
osopardo1, line 26: Why `dataChange` should be a part of the `IndexFile` interface? Is it not a particular case for Delta?

osopardo1: Maybe is good to add it, but before introducing it here, we should think if we will maintain it through other implementations or not

delta/src/main/scala/io/qbeast/spark/delta/DeltaMetadataWriter.scala
```diff
@@ -228,7 +228,8 @@ private[delta] case class DeltaMetadataWriter(
       .collect()
       .map(DeltaQbeastFileUtils.fromAddFile(dimensionCount))
       .flatMap(_.tryReplicateBlocks(deltaReplicatedSet))
-      .map(DeltaQbeastFileUtils.toAddFile(dataChange = false))
+      .map(DeltaQbeastFileUtils.toAddFile)
+      .map(addFile => addFile.copy(dataChange = false))
```
osopardo1, line 232: you can do this operation in one single map:

```
.map(file => {
val addFile = DeltaQbeastFileUtils.toAddFile(file)
addFile.copy(dataChange = false)
}
```

osopardo1: But here it's confusing because according to the interface, I would expect that the `IndexFile` would have the `dataChange` flag already parsed in `toAddFile`

src/main/scala/io/qbeast/table/IndexedTable.scala
```diff
@@ -515,14 +515,17 @@ private[table] class IndexedTableImpl(
               DeleteFile(
                 path = indexFile.path,
                 size = indexFile.size,
+                dataChange = false,
```
osopardo1, line 518: DeleteFiles always have the dataChange to false?

JosepSampe: The default dataChange is always `true`. in this case, as it is in the `optimize()`, it must be `false`

delta/src/main/scala/io/qbeast/spark/delta/DeltaMetadataWriter.scala
JosepSampe: I agree this can be done in the same map. 

On the other hand, I couldn't test this particular method. I'm not sure which test is using it, or if it's even being used, since the method is related to 'replication,' and as far as I know, replication is not utilized. In any case, I forced it to be `false` because the default `dataChange` is always `true`.

core/src/main/scala/io/qbeast/core/model/IndexFileBuilder.scala
JosepSampe: I've been considering two different options:

- The first option (current PR) is to include the `dataChange` flag in the `QbeastFile`, allowing it to be propagated from one place to another. This approach works with the current `metadataManager` implementation, where both `write()` and `optimize()` call the same `metadataManager.updateWithTransaction()` method.

- The second option I've been testing involves extending the `metadataManager` to create a new method (e.g., `metadataManager.updateWithTransactionOptimize()`). This would differentiate between `write()` calling `metadataManager.updateWithTransactionWrite()`, where `dataChange` must be true, and optimize() calling `metadataManager.updateWithTransactionOptimize()`,, where dataChange must be false.

What do you think? Do you see an alternative approach?

osopardo1: Mmm, interesting. It is true that, right now, the data change is property of the Metadata Manager (or the underlying format), and is true that varies from operations... But I don't want to add more methods without setting a proper evaluation with this flag. 

I think it's best if we leave it in the `QbeastFile` for now. So the IndexedTable can manage it depending on the operation.

Issue #429, opened 2024-09-27
Time Travel for Qbeast-Spark!


Jiaweihu08:
Add time-traveling feature to qbeast-spark:

```scala
// Time travel using version
val df = spark.read.format("qbeast").option("versionAsOf", "1984").load(tablePath)
val df = spark.sql("SELECT * FROM myCatalog.tableName VERSION AS OF 1984")

// Time travel using timestamp
val df = spark.read.format("qbeast").option("timestampAsOf", "1844-10-15").load(tablePath)
val df = spark.sql("SELECT * FROM myCatalog.tableName TIMESTAMP AS OF '1844-10-15'")
```

PR #446, opened 2024-10-24
Issue #398: Qbeast Spark Refactor -- Delta abstraction

JosepSampe:
## Description

Merge Refactoring Branch into Main #447 
Final part of #398


core/src/main/scala/io/qbeast/core/model/PreCommitHook.scala
```diff
@@ -46,7 +45,7 @@ trait PreCommitHook {
    * @return
    *   The output of the hook as a `PreCommitHookOutput`.
    */
-  def run(actions: Seq[Action]): PreCommitHookOutput
+  def run(actions: Seq[QbeastFile]): PreCommitHookOutput
```
osopardo1, line 48: The purpose of the `PreCommitHook` should be to inject any block of code to any action available to the table. It is true that we only manage write actions, so it's fair to assume the input would be a list of `QbeastFile`. But in the future, I would expect this to be extended to different operations (change schema, update, delete...)

If it's not for this PR, maybe open an issue and discuss about it. Wdyt @Jiaweihu08 ?

core/src/main/scala/org/apache/spark/qbeast/config.scala
```diff
@@ -58,19 +58,31 @@ package object config {
...
+    ConfigBuilder("spark.qbeast.tableFormat")
+      .version("0.2.0")
+      .stringConf
+      .createWithDefault("delta")
+
+  def DEFAULT_NUMBER_OF_RETRIES: Int = SparkSession.active.sparkContext.conf
     .get(defaultNumberOfRetries)
 
-  def DEFAULT_CUBE_SIZE: Int = QbeastContext.config
+  def DEFAULT_CUBE_SIZE: Int = SparkSession.active.sparkContext.conf
```
osopardo1, line 70: should we define a `getActiveSession` method?

delta/src/main/scala/io/qbeast/spark/delta/DeltaMetadataWriter.scala
```diff
@@ -149,11 +152,13 @@ private[delta] case class DeltaMetadataWriter(
...
-  private def runPreCommitHooks(actions: Seq[Action]): PreCommitHookOutput = {
-    preCommitHooks.foldLeft(Map.empty[String, String]) { (acc, hook) => acc ++ hook.run(actions) }
+  private def runPreCommitHooks(actions: Seq[QbeastFile]): PreCommitHookOutput = {
+    preCommitHooks.foldLeft(Map.empty[String, String]) { (acc, hook) =>
+      acc ++ hook.run(actions)
+    }
   }
 
-  def writeWithTransaction(writer: => (TableChanges, Seq[FileAction])): Unit = {
+  def writeWithTransaction(writer: => (TableChanges, Seq[IndexFile], Seq[DeleteFile])): Unit = {
```
osopardo1, line 161: Again, maybe not for this PR, but could we create our qbeast `FileAction` abstraction for the action committed on the transaction? Or is it really useful to have Index and Deleted files in separated variables?

osopardo1: That way, we can have a single map for both sequences and in the future we could add different actions to the list. 

Up to discussion. what do you think?

delta/src/main/scala/io/qbeast/spark/delta/DeltaQbeastSnapshot.scala
```diff
@@ -16,26 +16,31 @@
...
  *
- * @param snapshot
- *   the internal Delta Lakes log snapshot
+ * @param tableID
+ *   the table ID
  */
-case class DeltaQbeastSnapshot(protected override val snapshot: Snapshot)
-    extends QbeastSnapshot
-    with DeltaStagingUtils {
+case class DeltaQbeastSnapshot(tableID: QTableID) extends QbeastSnapshot with DeltaStagingUtils {
```
osopardo1, line 40: Having the `Snapshot` attached to the `TableId` would give you access all the time to the most updated snapshot, which conflicts with the terminology. The concept of Snapshot should be an isolated abstraction of the state of a table for a given moment. 

The most immediate consequence is that removes the possibility of adding time travel (see issue #429 ). 


Again, not for this PR. But agree to open an issue and document it, if we agree. Possibily it's not necessary to build it asap.

src/main/scala/io/qbeast/table/IndexedTable.scala
```diff
@@ -470,17 +473,24 @@ private[table] class IndexedTableImpl(
       options: QbeastOptions,
       append: Boolean): Unit = {
     logTrace(s"Begin: Writing data to table $tableID")
-    val stagingDataManager: StagingDataManager = new StagingDataManager(tableID)
+    val stagingDataManager: StagingDataManager = stagingDataManagerFactory.getManager(tableID)
```
osopardo1, line 476: shouldn't the staging data manager be global for the `IndexedTable`?

src/test/scala/io/qbeast/QbeastIntegrationTestSpec.scala
```diff
@@ -204,8 +206,13 @@ trait QbeastIntegrationTestSpec extends AnyFlatSpec with Matchers with DatasetCo
    * @tparam T
    * @return
    */
-  def withOTreeAlgorithm[T](code: IndexManager[DataFrame] => T): T = {
+  def withOTreeAlgorithm[T](code: IndexManager => T): T = {
```
osopardo1, line 209: let's rename it with `withSparkOTreeManager` haha............. 🤡

osopardo1 (commented):
Few comments left! Some of them can be solved here, others would be best to leave it for another discussion. 
Let me know what do you think 🙏 

The rest looks goooood

core/src/main/scala/io/qbeast/core/model/PreCommitHook.scala
Jiaweihu08: There's indeed a narrowing of scope from delta `Action` to `QbeastFile.`
We need an issue to discuss this and expand the classes in qbeast for other available actions.

osopardo1: Would open one 👍

src/main/scala/io/qbeast/table/IndexedTable.scala
osopardo1: resolution: do not change on this PR

delta/src/main/scala/io/qbeast/spark/delta/DeltaQbeastSnapshot.scala
osopardo1: opened issue #448

delta/src/main/scala/io/qbeast/spark/delta/DeltaMetadataWriter.scala
osopardo1: Opened issue #449

osopardo1 (approved):

core/src/main/scala/org/apache/spark/qbeast/config.scala
```diff
@@ -58,19 +58,31 @@ package object config {
       .intConf
       .createWithDefault(3)
 
-  def DEFAULT_NUMBER_OF_RETRIES: Int = QbeastContext.config
+  private[config] val tableFormat: ConfigEntry[String] =
+    ConfigBuilder("spark.qbeast.tableFormat")
+      .version("0.2.0")
```
Jiaweihu08, line 63: The version here should be 0.8.0?

JosepSampe: Updated

osopardo1:
Just one small thing that we notice!!

We should NOT update the README documentation. The only document that should be updated with developments is `Contributing.md` (developer docs), so maintainers and devs can test with the proper configuration.

But for regular users, copying and pasting from the docs would not work with the new packages and the latest version 0.7.0.

CONTRIBUTING.md
```diff
@@ -193,8 +193,8 @@ sbt assembly
 $SPARK_HOME/bin/spark-shell \
 --jars ./target/scala-2.12/qbeast-spark-assembly-0.6.0.jar \
```
osopardo1, line 194: here we should also change the version of assembly to 0.8.0-SNAPSHOT

osopardo1: In fact, I would suggest to update the whole example with this new version, so we can start developing from Refactoring

Issue #463, opened 2024-11-12
Allow custom filename for storing rows during write


JosepSampe:
Currently, the filename `.parquet` where the rows are stored is generated using a UUID (UUID.parquet) independently of the table format. However, certain table formats might require a custom `filename` pattern to function correctly. 

We need to extend the DataWriter to support custom filename patterns, allowing each table format to provide its own specific naming convention.

Acceptance Criteria:

- Extend DataWriter to allow each table format to specify a custom filename pattern.
- Ensure backward compatibility by defaulting to the UUID-based naming if no custom pattern is provided.

Issue #460, opened 2024-11-08
Allow enable/disable of RollUp feature


osopardo1:
The features or components that we add, even if they are expected to be the default behaviour, should have a mechanism to enable and disable them internally. 

This issue was created with the Rollup feature in mind but can be extended to others if we find it necessary.

PR #464, opened 2024-11-12
Issue #463: Support custom filename patterns in DataWriter

JosepSampe:
## Description

Fix for #463 

This PR allows table formats to specify custom filenames through a new configuration option in DataWriter. If a custom filename pattern is not provided, the default UUID.parquet format will be used, ensuring backward compatibility.

This PR updates the getRollupCubeIdUDF function to return a UUID instead of an Array[Byte]. It was found that the column added by this UDF is used solely to determine the file writer and filename. By using a UUID directly, we simplify the logic, as each unique RollupCubeID now maps to a unique UUID for filename generation.

## Type of change
- refactoring

## Checklist:

- [X] New feature / bug fix has been committed following the [Contribution guide](https://github.com/Qbeast-io/qbeast-spark/blob/main/CONTRIBUTING.md).
- [X] Add logging to the code following the [Contribution guide](https://github.com/Qbeast-io/qbeast-spark/blob/main/CONTRIBUTING.md).
- [X] Add comments to the code (make it easier for the community!).
- [X] Change the documentation.
- [X] Add tests.
- [X] Your branch is updated to the main branch (dependent changes have been merged).

### DIFF
--- a/core/src/main/scala/io/qbeast/spark/index/QbeastColumns.scala
+++ b/core/src/main/scala/io/qbeast/spark/index/QbeastColumns.scala
@@ -52,7 +52,7 @@ object QbeastColumns {
   /**
    * Cube to rollup column name.
    */
-  val cubeToRollupColumnName = "_qbeastCubeToRollup"
+  val fileUUIDColumnName = "_qbeastFileUUID"
 
   /**
    * Cube to rollup file name column name.
@@ -65,7 +65,7 @@ object QbeastColumns {
     stateColumnName,
     revisionColumnName,
     cubeToReplicateColumnName,
-    cubeToRollupColumnName,
+    fileUUIDColumnName,
     filenameColumnName)
 
   /**
@@ -94,7 +94,7 @@ object QbeastColumns {
       stateColumnIndex = columnIndexes.getOrElse(stateColumnName, -1),
       revisionColumnIndex = columnIndexes.getOrElse(revisionColumnName, -1),
       cubeToReplicateColumnIndex = columnIndexes.getOrElse(cubeToReplicateColumnName, -1),
-      cubeToRollupColumnIndex = columnIndexes.getOrElse(cubeToRollupColumnName, -1),
+      fileUUIDColumnIndex = columnIndexes.getOrElse(fileUUIDColumnName, -1),
       filenameColumnIndex = columnIndexes.getOrElse(filenameColumnName, -1))
   }
 
@@ -134,8 +134,8 @@ object QbeastColumns {
  *   the revision column index or -1 if it is missing
  * @param cubeToReplicateColumnIndex
  *   the cube to replicate column index or -1 if it is missing
- * @param cubeToRollupColumnIndex
- *   the cube to rollup column index or -1 if it is missing
+ * @param fileUUIDColumnIndex
+ *   target file UUID column index or -1 if it is missing
  * @param filenameColumnIndex
  *   the cube to rollup file name column index or -1 if it is missing
  */
@@ -145,7 +145,7 @@ case class QbeastColumns(
     stateColumnIndex: Int,
     revisionColumnIndex: Int,
     cubeToReplicateColumnIndex: Int,
-    cubeToRollupColumnIndex: Int,
+    fileUUIDColumnIndex: Int,
     filenameColumnIndex: Int) {
 
   /**
@@ -162,7 +162,7 @@ case class QbeastColumns(
     columnIndex == stateColumnIndex ||
     columnIndex == revisionColumnIndex ||
     columnIndex == cubeToReplicateColumnIndex ||
-    columnIndex == cubeToRollupColumnIndex ||
+    columnIndex == fileUUIDColumnIndex ||
     columnIndex == filenameColumnIndex
   }
 
@@ -212,7 +212,7 @@ case class QbeastColumns(
    * @return
    *   the cube to rollup column exists
    */
-  def hasCubeToRollupColumn: Boolean = cubeToRollupColumnIndex >= 0
+  def hasFileUUIDColumn: Boolean = fileUUIDColumnIndex >= 0
 
   /**
    * Returns whether the filename column exists.
--- a/core/src/main/scala/io/qbeast/spark/index/QbeastColumns.scala
+++ b/core/src/main/scala/io/qbeast/spark/index/QbeastColumns.scala
@@ -54,13 +54,19 @@ object QbeastColumns {
    */
   val cubeToRollupColumnName = "_qbeastCubeToRollup"
 
+  /**
+   * Cube to rollup file name column name.
+   */
+  val filenameColumnName = "_qbeastFilename"
+
   val columnNames: Set[String] = Set(
     weightColumnName,
     cubeColumnName,
     stateColumnName,
     revisionColumnName,
     cubeToReplicateColumnName,
-    cubeToRollupColumnName)
+    cubeToRollupColumnName,
+    filenameColumnName)
 
   /**
    * Creates an instance for a given data frame.
@@ -88,7 +94,8 @@ object QbeastColumns {
       stateColumnIndex = columnIndexes.getOrElse(stateColumnName, -1),
       revisionColumnIndex = columnIndexes.getOrElse(revisionColumnName, -1),
       cubeToReplicateColumnIndex = columnIndexes.getOrElse(cubeToReplicateColumnName, -1),
-      cubeToRollupColumnIndex = columnIndexes.getOrElse(cubeToRollupColumnName, -1))
+      cubeToRollupColumnIndex = columnIndexes.getOrElse(cubeToRollupColumnName, -1),
+      filenameColumnIndex = columnIndexes.getOrElse(filenameColumnName, -1))
   }
 
   /**
@@ -129,6 +136,8 @@ object QbeastColumns {
  *   the cube to replicate column index or -1 if it is missing
  * @param cubeToRollupColumnIndex
  *   the cube to rollup column index or -1 if it is missing
+ * @param filenameColumnIndex
+ *   the cube to rollup file name column index or -1 if it is missing
  */
 case class QbeastColumns(
     weightColumnIndex: Int,
@@ -136,7 +145,8 @@ case class QbeastColumns(
     stateColumnIndex: Int,
     revisionColumnIndex: Int,
     cubeToReplicateColumnIndex: Int,
-    cubeToRollupColumnIndex: Int) {
+    cubeToRollupColumnIndex: Int,
+    filenameColumnIndex: Int) {
 
   /**
    * Returns whether a given column is one of the Qbeast columns.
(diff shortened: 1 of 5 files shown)


core/src/main/scala/io/qbeast/spark/writer/RollupDataWriter.scala
```diff
@@ -76,11 +76,12 @@ trait RollupDataWriter extends DataWriter {
...
       getIndexFileWriterFactory(tableId, schema, extendedData, revisionId, trackers)
     extendedRows => {
-      val writers = mutable.Map.empty[CubeId, IndexFileWriter]
+      val writers = mutable.Map.empty[String, IndexFileWriter]
       extendedRows.foreach { extendedRow =>
-        val (row, weight, cubeId, rollupCubeId) = extract(extendedRow)
+        val (row, weight, cubeId, filename) = extract(extendedRow)
         val cubeMaxWeight = getCubeMaxWeight(cubeId)
-        val writer = writers.getOrElseUpdate(rollupCubeId, writerFactory.createIndexFileWriter())
+        val writer =
```
osopardo1, line 83: careful with the space!

delta/src/main/scala/io/qbeast/spark/delta/DeltaRollupDataWriter.scala
```diff
@@ -40,23 +39,24 @@ import java.net.URI
...
     val revision = tableChanges.updatedRevision
     val dimensionCount = revision.transformations.length
 
     val statsTrackers = StatsTracker.getStatsTrackers
     val fileStatsTracker = getFileStatsTracker(tableId, data)
     val trackers = statsTrackers ++ fileStatsTracker
 
-    val filesAndStats = internalWrite(tableId, schema, data, tableChanges, trackers)
+    val extendedData = extendDataWithCubeToRollup(data, tableChanges)
+    val filesAndStats = doWrite(tableId, schema, extendedData, tableChanges, trackers)
```
osopardo1, line 58: where's the internalWrite call?

core/src/main/scala/io/qbeast/spark/index/QbeastColumns.scala
```diff
@@ -202,4 +213,12 @@ case class QbeastColumns(
...
    */
   def hasCubeToRollupColumn: Boolean = cubeToRollupColumnIndex >= 0
+
+  /**
+   * Returns whether the filename column exists.
+   *
+   * @return
+   *   the filename column exists
+   */
+  def hasFilenameColumn: Boolean = filenameColumnIndex >= 0
```
osopardo1, line 223: There's no other option for this feature than using a column in the dataframe? Couldn't it be calculated on the fly and returned in the output file stats? 

If there's no other way around it, should we add another component that takes care of dividing the DataFrame into files based on its logic? In that way, we can accommodate different strategies to choose and distribute files.

osopardo1: For example:

`OTreeManager` -> `FileManager` -> `DataWriter`

JosepSampe: > There's no other option for this feature than using a column in the dataframe?

In this case, yes. The file name must be present in all the rows of the DataFrame before storing it to the Parquet file when using Hudi. For Delta, this is unnecessary.

> If there's no other way around it, should we add another component that takes care of dividing the DataFrame into files based on its logic?

The logic that handles this is in the `RollupDataWriter`. If the column is not set, it will work as before, without additional overhead.

delta/src/main/scala/io/qbeast/spark/delta/DeltaRollupDataWriter.scala
JosepSampe: Has been renamed to `doWrite()`, in the `RollupDataWriter`

core/src/main/scala/io/qbeast/spark/index/QbeastColumns.scala
osopardo1: > The logic that handles this is in the RollupDataWriter

Yeaaa, but I think we should have an interface between the RollUp and the final DataWriter. Maybe for another PR, if we agree 🐒 


What I am not convinced from a logical perspective, is that we use the `getRollupCubeIdUDF` to return something that is not a CubeId... 
Also, the `fileName` contains the string for the `cubeId` somewhere? Would it be a good practise to save it as cubeID + timestamp or cubeID + UUID ?

JosepSampe: > What I am not convinced from a logical perspective, is that we use the getRollupCubeIdUDF to return something that is not a CubeId...

In practice, the UUID is a unique identifier for the Cube. That's why I included it here. Another option I considered was creating a separate column, `cubeToRollupUUID`, but then the original `cubeToRollup `column would go unused, as its only purpose is to create a different data writer for each diferent `cubeToRollup`, which could be done with the `cubeToRollupUUID`.

> Also, the fileName contains the string for the cubeId somewhere? Would it be a good practise to save it as cubeID + timestamp or cubeID + UUID ?

As of today the filename does not contain the cube ID. I'm not certain about Delta, but Hudi is very strict with its filename format, so including it is not possible. The approach I'm following now is to make the filename exactly match the format expected by each underlying table format. In Delta, it's `UUID.parquet`; in Hudi, it's `UUID_TOKEN_TIMESTAMP.parquet`; and in Iceberg, it appears to be `UUID.parquet`.

osopardo1: >  the UUID is a unique identifier for the Cube..

mmm, I think it's important not to merge the concepts of Cube Identifier with File Identifier. The CubeID is not (or should not be) a UUID. UUID is unique, while the Cube ID can have the same value in multiple use cases. 

A different data writer should be created for a different File. In this case (the default), each file belongs to the rollup cube, but for example, if we want to turn RollUp off (issue #460 ), this casuistic would not be applied. 

> The approach I'm following now is to make the filename exactly match the format expected by each underlying table format.

Then sounds good to add a UUID!! 

Just make sure is referenced as `fileUUID` and it does not interfere with the Cube ID concept. Even if the mapping is 1:1, they are created through an entirely different process and have different meanings. And allows the code to be more flexible.

JosepSampe: So, in your opinion, how would you proceed?

- Keep the `rollupCubeID` column and add another one for the `cubeUUID`?
- Remove the `rollupCubeID` column and create a more generic column, such as `partitionUUID`, which identifies the destination partition (parquet file) for the data? In the case of a rollup, `cubeToRollup` would populate the `partitionUUID`, but if the rollup is deactivated, a different logic would populate the `partitionUUID` for each row.
- Other ideas?

osopardo1: After small discussion, we agreed to:
- using `fileUUID` as a file identifier, independently of `rollupCubeID`.
- For the DataWriter, the only important column name should be fileID, which would create a writer per group, 
- In other PR's, we should think about decoupling the `RollUp` from the `DataWriter` code.

osopardo1 (approved):
LGTM; let me confirm with @Jiaweihu08 that the `cubeToRollup` is only used for file writing purposes, and I would merge 👍

core/src/main/scala/io/qbeast/spark/writer/RollupDataWriter.scala
```diff
@@ -138,21 +140,33 @@ trait RollupDataWriter extends DataWriter {
...
-    var rollupCubeId = rollup.get(cubeId)
-    var parentCubeId = cubeId.parent
-    while (rollupCubeId.isEmpty) {
-      parentCubeId match {
-        case Some(value) =>
-          rollupCubeId = rollup.get(value)
-          parentCubeId = value.parent
-        case None => rollupCubeId = Some(cubeId)
+  private def computeCubeUUIDs(rollup: Map[CubeId, CubeId]): Map[CubeId, String] = {
+    val uuidCache = mutable.Map[CubeId, String]()
```
Jiaweihu08, line 144: This can be simplified as follows:
```
rollup.values.toSeq.distinct.map(c => c -> UUID.randomUUID().toString).toMap
```

core/src/main/scala/io/qbeast/spark/writer/RollupDataWriter.scala
```diff
@@ -138,21 +140,33 @@ trait RollupDataWriter extends DataWriter {
...
+      if (!uuidCache.contains(cubeToRollupId)) {
+        uuidCache(cubeToRollupId) = UUID.randomUUID().toString
       }
     }
-    rollupCubeId.get.bytes
-  })
+    uuidCache.toMap
+  }
+
+  private def getFileUUIDUDF(
```
Jiaweihu08, line 153: Using two maps for a UDF can add significant overhead.
You can merge `rollup` and `rollupCubeUUIDs` : `rollup + rollupCubeUUIDs: Map[CubeId, String]`

core/src/main/scala/io/qbeast/spark/writer/RollupDataWriter.scala
```diff
@@ -100,9 +101,12 @@ trait RollupDataWriter extends DataWriter {
       val weight = Weight(extendedRow.getInt(qbeastColumns.weightColumnIndex))
       val cubeIdBytes = extendedRow.getBinary(qbeastColumns.cubeColumnIndex)
       val cubeId = revision.createCubeId(cubeIdBytes)
-      val rollupCubeIdBytes = extendedRow.getBinary(qbeastColumns.cubeToRollupColumnIndex)
-      val rollupCubeId = revision.createCubeId(rollupCubeIdBytes)
-      (row, weight, cubeId, rollupCubeId)
+      val filename = {
```
Jiaweihu08, line 104: The column with which the data was partitioned (line 58) and the value we organize index file writers with (line 84) must be the same. You can use an internal variable to sync them.

Jiaweihu08 (changes requested):

core/src/main/scala/io/qbeast/spark/writer/RollupDataWriter.scala
JosepSampe: Technically it is the same:

In line 58 I use:
```
.repartition(col(QbeastColumns.fileUUIDColumnName))
```
In line 84 we use filename, which is extracted from:
```
  val filename = {
    if (qbeastColumns.hasFilenameColumn)
      extendedRow.getString(qbeastColumns.filenameColumnIndex)
    else extendedRow.getString(qbeastColumns.fileUUIDColumnIndex) + ".parquet"
  }
```
Both contains the same UUID, the difference is that a table format can provide a custom filename (containing this UUID) or not. In case of delta the `qbeastColumns.filenameColumnIndex` column is not set, so it will use the `qbeastColumns.fileUUIDColumnIndex`. In case of hudi `qbeastColumns.filenameColumnIndex` is set, so it will use it (contains the UUID in `qbeastColumns.fileUUIDColumnName` + custom filename convention).

core/src/main/scala/io/qbeast/spark/writer/RollupDataWriter.scala
JosepSampe: Fixed

core/src/main/scala/io/qbeast/spark/writer/RollupDataWriter.scala
JosepSampe: Fixed

Jiaweihu08 (changes requested):

Jiaweihu08 (approved):

PR #471, opened 2024-11-15
Issue #465: Allow MetadataManager to pass the commit timestamp to DataWriter

### LINKED ISSUE #465: Allow MetadataManager to pass commitTimestamp to DataWriter
Currently, the commitTimestamp is created inside `MetadataManager`, but there is no mechanism to pass it to `DataWriter`. Some table formats require this timestamp to generate filenames and populate additional metadata columns during writes. This enhancement will allow the MetadataManager to provide the commitTimestamp to the `DataWriter`, enabling format-specific handling of filenames and extra columns.

Acceptance Criteria:

- Implement a way for `MetadataManager` to transfer commitTimestamp to `DataWriter`.
- Ensure `DataWriter `can use the commitTimestamp for filename generation and additional columns as required by specific table formats.
- Maintain backward compatibility for cases where commitTimestamp is not needed by the `DataWriter`.
- Have at least one test case that reproduces the issue and shows that the code changes fix it.

JosepSampe:
## Description
Allows to pass the timestamp generated in the MetadataManager to the DataWriter

Fixes #465 

## Type of change

feature addition

## Checklist:

- [x] Add tests.

### DIFF
--- a/delta/src/main/scala/io/qbeast/spark/delta/DeltaMetadataWriter.scala
+++ b/delta/src/main/scala/io/qbeast/spark/delta/DeltaMetadataWriter.scala
@@ -177,8 +177,8 @@ private[delta] case class DeltaMetadataWriter(
       registerStatsTrackers(statsTrackers)
 
       // Execute write
-      val commitStartTime = txn.txnStartTimeNs.toString
-      val (tableChanges, indexFiles, deleteFiles) = writer(commitStartTime)
+      val transactionStartTime = txn.txnStartTimeNs.toString
+      val (tableChanges, indexFiles, deleteFiles) = writer(transactionStartTime)
       val addFiles = indexFiles.map(DeltaQbeastFileUtils.toAddFile)
       val removeFiles = deleteFiles.map(DeltaQbeastFileUtils.toRemoveFile)
 
--- a/delta/src/main/scala/io/qbeast/spark/delta/DeltaMetadataWriter.scala
+++ b/delta/src/main/scala/io/qbeast/spark/delta/DeltaMetadataWriter.scala
@@ -177,7 +177,8 @@ private[delta] case class DeltaMetadataWriter(
       registerStatsTrackers(statsTrackers)
 
       // Execute write
-      val (tableChanges, indexFiles, deleteFiles) = writer(txn.txnStartTimeNs.toString)
+      val commitStartTime = txn.txnStartTimeNs.toString
+      val (tableChanges, indexFiles, deleteFiles) = writer(commitStartTime)
       val addFiles = indexFiles.map(DeltaQbeastFileUtils.toAddFile)
       val removeFiles = deleteFiles.map(DeltaQbeastFileUtils.toRemoveFile)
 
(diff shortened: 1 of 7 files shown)


delta/src/main/scala/io/qbeast/spark/delta/DeltaMetadataWriter.scala
```diff
@@ -176,7 +177,7 @@ private[delta] case class DeltaMetadataWriter(
       registerStatsTrackers(statsTrackers)
 
       // Execute write
-      val (tableChanges, indexFiles, deleteFiles) = writer
+      val (tableChanges, indexFiles, deleteFiles) = writer(txn.txnStartTimeNs.toString)
```
osopardo1, line 180: I need a bit more of understanding here. Why the commitTime is the start transaction time?

osopardo1: And isn't it better if we manipulate it as Long?

osopardo1: Nevertheless, instead of `commitTime`, we should use a more verbose naming (such as `startCommitTime`) to avoid confusion in the future.
 Also, do you think it should be saved somewhere? I know it's Hudi-specific, but should we include it in Qbeast protocol too?

Jiaweihu08: Or `txnStartTime,` as it is created when the transaction starts?

Jiaweihu08: Even before any parquet file is written

osopardo1: I would suggest using something more general than `txn` ... although the three formats are transaction-based, the abbreviation is widely used in Delta Lake only I believe.

Maybe `transactionStartTime` if you all agree 😆 😝 🙄  `commitStartTime` feels ok too

Github should allow polls for this.........

JosepSampe: In hudi they call it `instantTime`, so i proceed with `commitStartTime` 😆

Jiaweihu08: IMHO, `transactionStartTime` does a better job of conveying the nature of this parameter.

The thing is that the time is not exactly the commit start time.
It is created when we start indexing and analyzing the data.
Then, it is written, which takes a huge chunk of time, and then, finally, it starts with the commit.

osopardo1: That is a good point actually..

JosepSampe: Changed

osopardo1 (approved):

Jiaweihu08 (approved):

Issue #475, opened 2024-11-19
Enhance Metadata Manager to Differentiate Optimize Operations


JosepSampe:
Currently, the `MetadataManager` accepts a variable called `append`, which can be `true` or `false`. This enables differentiation between `append` and `overwrite` operations. However, it does not account for `optimize` operations.

While there exists a `tableChanges.isOptimizeOperation` variable, it appears to always return `false`, which prevents the `MetadataWriter` from correctly identifying `optimize` operations. In any case, the `tableChanges.isOptimizeOperation` is produced after the `write()` operation, but in `Hudi`, it is required to know in advance that the operation is an `optimize` operation before the `write()` process begins.

PR #476, opened 2024-11-19
Issue #475: Add Qbeast WriteMode to distinguish between write operations

JosepSampe:
## Description

Introduce Qbeast WriteMode to distinguish between Append, Overwrite, and Optimize operations #475 

## Type of change

New feature

## Checklist:

- [x] Add tests.

### DIFF
--- a/core/src/main/scala/io/qbeast/core/model/WriteMode.scala
+++ b/core/src/main/scala/io/qbeast/core/model/WriteMode.scala
@@ -1,0 +1,27 @@
+/*
+ * Copyright 2021 Qbeast Analytics, S.L.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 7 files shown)


core/src/main/scala/io/qbeast/core/model/WriteMode.scala
```diff
@@ -0,0 +1,27 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.qbeast.core.model
+
+/**
+ * Names of possible write modes
+ */
+
+object WriteMode {
```
osopardo1, line 22: can we use this object to map between other Mode classes? (for Delta, Spark...)

JosepSampe: Not sure I'm following. What do you mean exactly?

osopardo1: I think that in different parts of the codebase we reference the SaveModes from Spark (QbeastDataSource) and from Delta (DeltaMetadataManager)

Should we add to this WriteMode object the mapping between them?

A method like:

`.fromSparkSaveMode(...)
`
or do you think it is best to keep the `WriteMode` as an enumeration an leave the logic to the external components (as it is right now)?

JosepSampe: In principle, the only place I see a potential replacement of `SaveMode`->`WriteMode` is in the files I modified in this PR, primarily in the `DeltaMetadataManager/Writer`. While I do see `SaveMode` referenced in other files, I think it is not interchangeable.

Jiaweihu08 (approved):
LGTM

Issue #489, opened 2024-11-28
Extend QbeastSnapshot Interface to Expose Relevant Information


JosepSampe:
The current `QbeastSnapshot` interface provides several abstractions for accessing table-related metadata and configurations. However, to improve its usability and decouple it from format-specific details, it is required to extend the interface with two new methods:

- `loadConfiguration()`: This method will load and return the table configuration, including the `TBLPROPERTIES`.
- `loadLastCommitTags()`: This method will retrieve the tags added by the `preCommitHooks` execution for the last commit.

PR #492, opened 2024-11-28
Issue #489: Extend QbeastSnapshot Interface to Expose Relevant Information

JosepSampe:
## Description

Add `loadConfiguration` in the QbeastSnapshot #489 
Moved a few tests to the delta folder

## Type of change

New feature

## Checklist:

- [x] Add tests.

### DIFF
--- a/core/src/main/scala/io/qbeast/core/model/QbeastSnapshot.scala
+++ b/core/src/main/scala/io/qbeast/core/model/QbeastSnapshot.scala
@@ -56,6 +56,18 @@ trait QbeastSnapshot {
   def loadDescription: String
 
   /**
+   * The current table configuration.
+   * @return
+   */
+  def loadConfiguration: Map[String, String]
+
+  /**
+   * The last commit tags.
+   * @return
+   */
+  def loadLastCommitTags: Map[String, String]
+
+  /**
    * The current table properties of the snapshot.
    * @return
    */
--- a/core/src/main/scala/io/qbeast/core/model/QbeastSnapshot.scala
+++ b/core/src/main/scala/io/qbeast/core/model/QbeastSnapshot.scala
@@ -62,10 +62,10 @@ trait QbeastSnapshot {
   def loadConfiguration: Map[String, String]
 
   /**
-   * The last commit tags.
+   * The last commit tags added by the PreCommitHooks.
    * @return
    */
-  def loadLastCommitTags: Map[String, String]
+  def loadLatestPreCommitHookInfo: Map[String, String]
 
   /**
    * The current table properties of the snapshot.
--- a/core/src/main/scala/io/qbeast/core/model/QbeastSnapshot.scala
+++ b/core/src/main/scala/io/qbeast/core/model/QbeastSnapshot.scala
@@ -42,7 +42,7 @@ trait QbeastSnapshot {
   /**
    * Returns the total number of data files in the snapshot.
    */
-  def allFilesCount: Long
+  def numOfFiles: Long
 
   /**
    * Provides the schema of the dataset for this snapshot.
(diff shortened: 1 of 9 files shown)


delta/src/main/scala/io/qbeast/spark/delta/DeltaQbeastSnapshot.scala
```diff
@@ -46,8 +49,9 @@ case class DeltaQbeastSnapshot(tableID: QTableID) extends QbeastSnapshot with De
 
   override val basePath: Path = new Path(tableID.id)
 
-  override val snapshot: Snapshot =
-    DeltaLog.forTable(SparkSession.active, tableID.id).update()
+  private val deltaLog: DeltaLog = DeltaLog.forTable(SparkSession.active, tableID.id)
```
osopardo1, line 52: To be honest, I'm not super happy with the Snapshot management in `QbeastSnapshot`... We are always loading the newest version, while the `QbeastSnapshot` should have a 1:1 relation with an underlying Snapshot of the Table.

QbeastTable can have access to different QbeastSnapshots. But a `QbeastSnapshot` must be isolated. Always access to the same set of files and metadata. 

I know we changed that during the Refactor because we were anyways initializing the object with the most updated one, but I think we should look at it carefully and fix it before adding more changes to it...

what do you think? feels relevant? or would you like to fix it after?

core/src/main/scala/io/qbeast/core/model/QbeastSnapshot.scala
```diff
@@ -55,6 +55,18 @@ trait QbeastSnapshot {
...
+   * The current table configuration.
+   * @return
+   */
+  def loadConfiguration: Map[String, String]
+
+  /**
+   * The last commit tags.
+   * @return
+   */
+  def loadLastCommitTags: Map[String, String]
```
osopardo1, line 68: What are these last commit tags, and why are they important for the QbeastSnapshot protocol?

osopardo1: Did you thought about other useful info we can include in this PR? (For example, last commit timestamp..)

osopardo1 (commented):
It's nice to retrieve more info from `QbeastSnapshot`, but we should at least think about more possible info and decide about what to do with the Snapshot Management.

core/src/main/scala/io/qbeast/core/model/QbeastSnapshot.scala
JosepSampe: > What are these last commit tags

These are the tags added by the PreCommmitHooks

> Did you thought about other useful info we can include in this PR? (For example, last commit timestamp..)

For now i'm adding what I need for hudi.

delta/src/main/scala/io/qbeast/spark/delta/DeltaQbeastSnapshot.scala
JosepSampe: From my point of view, currently I'm only adding things that I need for hudi. 
I think we should create a separate issue for this and address it in a different PR.

core/src/main/scala/io/qbeast/core/model/QbeastSnapshot.scala
osopardo1: > These are the tags added by the PreCommmitHooks

Could you rename it to `loadLatestPreCommitHookInfo` ? It feels more descriptive.

JosepSampe: Done

osopardo1 (approved):

PR #510, opened 2024-12-11
Issue #509: Remove QbeastStats and use the original JSON string from DeltaStatsTracker

### LINKED ISSUE #509: Qbeast Optimize breaks columns stats format when the min (or max) value is a string, but the value is a number.
## What went wrong?
I've indexed a large dataset (500 GB) with Qbeast, where one of the columns is named "title." After several incremental appends, I've run a full Qbeast optimize, but the JSON commit log is corrupted, as it happens that some rows have a number as a value, and it gets encoded as int instead of a string. 
So the last _delta_log file looks like this:
```json
{
  "numRecords": 50148,
  "minValues": {
    "title": 1997
  },
  "maxValues": {
    "title": "Veilchen",
  },
  "nullCount": {
    "text": 0,
  }
}
```
Where the title is a string, but in this file, it happens that the min value equals the number 1997, and for some reason, is then encoded as `1997` instead of `"1997"`
## How to reproduce?

It happened when indexing this dataset: https://huggingface.co/datasets/Upstash/wikipedia-2024-06-bge-m3. I'll try to create a smaller code snippet to recreate the issue in a smaller environment. 

I reproduced the error with both Apache Spark and deltalake. Hter the stack error in Spark:
```
Caused by: org.apache.spark.sql.catalyst.util.BadRecordException: com.fasterxml.jackson.core.JsonParseException: Unexpected character ('1' (code 49)): was expecting comma to separate Object entries
 at [Source: (byte[])"{"add":{"path":"355e1315-857f-4cee-bbac-4b4210c235ce.parquet","partitionValues":{},"size":299950993,"modificationTime":1733747878484,"dataChange":false,"stats":"{\"numRecords\":68001,\"minValues\":{\"url\":\"https://en.wikipedia.org/wiki/19\",\"text\":\"\\t//the array that will be sorted\",\"weight_partition\":0.0,\"id\":2428414.23,\"title\":"1927",\"text_size\":100},\"maxValues\":{\"url\":\"https://en.wikipedia.org/wiki/An?\",\"text\":\"? ,  Australasia (ANZ) was a com?\",\"weight_partiti"[truncated 701 bytes]; line: 1, column: 347]

```
```scala
// your-code-here
```

### 3. Spark version:
3.5.3

SLURM
### 6. Stack trace:

osopardo1:
## Description

Fixes #509 

## Type of change

Bug Fix.

## Checklist:

- [x] Add tests.

Tested unitary in `QbeastDeltaStatsTests` and integration in `QbeastDeltaStatsIntegrationTests`. We ensure the map is preserved for strings and numeric types.


JosepSampe (approved):
looks good

cugni (approved):
Looks great to me, thanks!

Issue #479, opened 2024-11-20
Error control + enforce columnStats when indexing non-deterministic or source-changing DataFrames 


osopardo1:
As a first solution for #466, we need to force users to add the `columnStats` when indexing Tables with the following characteristics:

- Underlying data source changes constantly.
- DataFrame contains non-deterministic columns to index.
- DataFrame contains non-deterministic predicates. 

ColumnStats would infer the data's min/max values before the DataFrame Analysis, which can produce inconsistent results when loading the DataFrame twice for Indexing in any of the above use cases.

osopardo1:
Update the title of the issue to treat here only the Non-Deterministic case.

Issue #414, opened 2024-09-16
Error when indexing a table with BIGINT 


osopardo1:
## What went wrong?
When creating a table using BIGINT on a date column and inserting a set of 10 rows, a `ScalaMatch` error appears. 

We would need to investigate the flow for the BIGINT type. Is weird, because when using CREATE TABLE(id LONG) it maps the type to BIGINT in the SHOW CREATE TABLE statement. But the indexing works ok 🤔 

## How to reproduce?

Different steps about how to reproduce the problem.

### 1. Code that triggered the bug, or steps to reproduce:

```scala
spark.sql("CREATE TABLE table(date BIGINT) USING qbeast OPTIONS('columnsToIndex'='date')")
spark.sql("INSERT INTO table SELECT temp_date FROM temp_view")
```

### 2. Branch and commit id:

main 

### 3. Spark version:
On the spark shell run `spark.version`.

3.5.0

### 4. Hadoop version:
On the spark shell run `org.apache.hadoop.util.VersionInfo.getVersion()`.

3.3.4

### 5. How are you running Spark?
Are you running Spark inside a container? Are you launching the app on a remote K8s cluster? Or are you just running the tests in a local computer?

Local 
### 6. Stack trace:
Trace of the log/error messages.

```shell

java.lang.illegalargumentexception: requirement failed
        at scala.Predef$.require(Predef.scala:268)
        at io.qbeast.core.model.CubeId$.containers(CubeId.scala:88)
        at io.qbeast.core.model.CubeDomainsBuilder.computeWeightsAndSizes(CubeDomainsBuilder.scala:132)
        at io.qbeast.core.model.CubeDomainsBuilder.resultInternal(CubeDomainsBuilder.scala:118)
        at io.qbeast.core.model.CubeDomainsBuilder.result(CubeDomainsBuilder.scala:114)
        at io.qbeast.spark.index.DoublePassOTreeDataAnalyzer$.$anonfun$computePartitionCubeDomains$6(OTreeDataAnalyzer.scala:177)
        at org.apache.spark.sql.execution.MapPartitionsExec.$anonfun$doExecute$3(objects.scala:198)

```

osopardo1:
I am not able to reproduce the error using a synthetic dataset in main. 

I either need:
- The original data that was written.
- Test with previous version (0.7.0). Will try this before asking for the data.

osopardo1:
Cannot reproduce with version 0.7.0 :/
[...]

PR #512, opened 2024-12-11
Issue #479: Error Control for Non-Deterministic Source Queries

osopardo1:
# Issue Description
Qbeast-io/qbeast-spark#414 revealed a significant limitation in the writing process for Spark.

The current approach requires multiple traversals over the `DataFrame,` including the following steps:
1. **Collecting statistics from the data:** Calculating stats such as min/max values for the columns to index and the count of elements.
2. **Estimating the index:** Determining how the data should be indexed.
3. **Indexing rows:** Assign rows to cubes
4. **Writing data:** Group rows by cubes and write them to files.

This repeated loading of the `DataFrame` introduces potential inconsistencies if the input data source is:
1. modified by another process or
2. built using non-deterministic functions that produce different results across executions

If either of these situations occurs, the results from one of the writing steps described above will not be reliable for the next. For instance, if a change occurs between steps 1 and 2 or steps 1 and 3, we may be unable to correctly position the row within the space. If a change to the source occurs after step 2, the estimated index may not be 100% correct for the changed data.

# Proposed Solution
To address part of this issue, we propose introducing a new agent, the `SparkPlanAnalyzer`, to improve error handling for non-deterministic queries before processing the data. 

## Non-Deterministic Queries Unsupported

To properly define the types of Queries that aren't supported anymore, let's list the Non-Deterministic query plans that Spark can encounter:

- **`LIMIT`** -> Applying a LIMIT clause twice (if the Source is not SORTED) would lead to different results.
- **`SAMPLE`** -> The Spark Sample Command uses Random Sampling to extract the percentage of rows specified in the operation. Unless it is a Qbeast Table, we cannot ensure the determinism of the results.
- **`FILTERING`** **with a Non-Deterministic Column** -> Using `rand()` or other types of non-deterministic predicates would lead to different results depending on the execution.
- **Indexing a Non-Deterministic Column** -> Calculating statistics over a non-deterministic column would provoke a mismatch between Transformations computed in step 1 and the results from steps 2 and 3. No error should be raised if all indexing columns are deterministic

## User Workaround
This approach provides users with 4 options:

- Modify the query: Ensure that the query (for the indexing columns) is deterministic to avoid inconsistencies.
- Materialize the query results before indexing. Instead of indexing the DF directly, save the data with another format before indexing.
- Add `columnStats` using the `.option` method: By providing column statistics directly, **users can mitigate these problems**. However, **this does not guarantee that the final written values will match those produced by the initial query.**
- Use more flexible Transformer Types: Changing the default `LinearTransformation` for numeric columns to `Quantiles` would map all values outside the min/max range to the extremes of the space. 

## Type of change

New feature.

## Checklist:

- [x] Add tests.

This is tested with `QbeastInputSourcesTest.` Added tests for error control when using:
- LIMIT
- FILTER BY Non-Deterministic Columns
- SAMPLE
- Use of Non-deterministic Expressions on projection.


osopardo1:
This PR is blocked until we complete the #520  . Since the determinism of the DataFrame is part of the Revision Flow/DataFrame Analysis, we aim to have a proper logic component before introducing new changes.

core/src/main/scala/io/qbeast/spark/index/SparkPlanAnalyzer.scala
```diff
@@ -0,0 +1,109 @@
...
+   * @param plan
+   *   the logical plan
+   * @return
+   */
+  private def isLogicalPlanDeterministic(plan: LogicalPlan): Boolean = {
+    // Recursively traverse the logical plan to find non-deterministic operations
+    val isCurrentOperationDeterministic = plan match {
+      case LocalLimit(_, _: Sort) => true // LocalLimit with Sort is deterministic
+      case LocalLimit(_, _) => false
+      case Sample(_, _, _, _, _) => false
```
Jiaweihu08, line 34: Doesn't this prevent a qbeast sample from being indexed with qbeast?
```
val qbeastDf = spark.read.format("qbeast").load(tablePath)
(qbeastDf
  .sample(0.1)
  .write
  .mode("append")
  .format("qbeast")
  .option("columnsToIndex", "col_1,col_2")
  .save(outputPath)
)
```

core/src/main/scala/io/qbeast/spark/index/SparkPlanAnalyzer.scala
```diff
@@ -0,0 +1,109 @@
...
+import org.apache.spark.sql.catalyst.plans.logical.LocalLimit
+import org.apache.spark.sql.catalyst.plans.logical.LogicalPlan
+import org.apache.spark.sql.catalyst.plans.logical.Project
+import org.apache.spark.sql.catalyst.plans.logical.Sample
+import org.apache.spark.sql.catalyst.plans.logical.Sort
+import org.apache.spark.sql.DataFrame
+
+import scala.collection.convert.ImplicitConversions.`collection asJava`
+
+object SparkPlanAnalyzer {
```
Jiaweihu08, line 15: Better use a `trait` and have `DataAnalyzer` extend it.

core/src/main/scala/io/qbeast/spark/index/SparkPlanAnalyzer.scala
```diff
@@ -0,0 +1,109 @@
...
+   *   - The logical plan of the DataFrame is checked for determinism
+   *   - The columns to index are checked for determinism
+   *
+   * @param dataFrame
+   *   the DataFrame to analyze
+   * @param columnsToIndex
+   *   the columns to index
+   * @return
+   */
+  def analyzeDataFrameDeterminism(dataFrame: DataFrame, columnsToIndex: Seq[String]): Boolean = {
```
Jiaweihu08, line 93: I want to propose the following:
```
def assertDataFrameDeterminism(dataFrame: DataFrame, revision: Revision): Boolean
```
This should be the first thing we do when entering `DataAnalyzer.analyze,` it should throw an error inside the method if the conditions aren't met.

core/src/main/scala/io/qbeast/spark/index/SparkPlanAnalyzer.scala
osopardo1: You are right, although I had my concern at first..
 
Currently, the sample with Qbeast is deterministic, because we use the hash of the columns to reproduce the same weight. But is that premise going to be true in the future? Or should the Sample with Qbeast be deterministic by all means?

core/src/main/scala/io/qbeast/spark/index/SparkPlanAnalyzer.scala
osopardo1: Two things here:

1. the api makes sense, in that way we can have a broader set of Revision characteristics to tests.
2. However I am wondering about the second statement...
> This should be the first thing we do when entering DataAnalyzer.analyze. 

What if the user changes the columns to index in the future? Wouldn't we need to asses that first with the `SparkRevisionFactory`?

core/src/main/scala/io/qbeast/spark/index/SparkPlanAnalyzer.scala
osopardo1: sure!

core/src/main/scala/io/qbeast/spark/index/SparkPlanAnalyzer.scala
Jiaweihu08: We can reverse the order if column changes ever come to be, but for now, I think it'd be smart to avoid computing the dataframe statistics if the process is to be aborted.

core/src/main/scala/io/qbeast/spark/index/SparkPlanAnalyzer.scala
Jiaweihu08: Can you detect a qbeast sample?

osopardo1: Yep, that is possible for sure

core/src/main/scala/io/qbeast/spark/index/OTreeDataAnalyzer.scala
```diff
@@ -216,6 +217,21 @@ object DoublePassOTreeDataAnalyzer
       indexStatus: IndexStatus,
       options: QbeastOptions): (DataFrame, TableChanges) = {
     logTrace(s"Begin: Analyzing the input data with existing revision: ${indexStatus.revision}")
+
+    // Check if the DataFrame is deterministic
+    logDebug(s"Checking the determinism of the input data")
+    val isSourceDeterministic =
+      analyzeDataFrameDeterminism(dataFrame, indexStatus.revision)
+    // TODO: we need to add columnStats control before the assert
```
Jiaweihu08, line 225: ColumnStats can't help the case if we consider the fact that it is not only the columns stats computed from the input dataframe will change, but the estimated cube domain sizes (consequently cube max weights) will also be wrong. 

In that case, the data can still be written but not as desired.

If we are ok with this, we can abort the process only when there is a `LinearTransformer` as other `Transformations` don't suffer from changing data.

core/src/main/scala/io/qbeast/spark/index/OTreeDataAnalyzer.scala
```diff
@@ -216,6 +217,21 @@ object DoublePassOTreeDataAnalyzer
       indexStatus: IndexStatus,
       options: QbeastOptions): (DataFrame, TableChanges) = {
     logTrace(s"Begin: Analyzing the input data with existing revision: ${indexStatus.revision}")
+
+    // Check if the DataFrame is deterministic
+    logDebug(s"Checking the determinism of the input data")
+    val isSourceDeterministic =
+      analyzeDataFrameDeterminism(dataFrame, indexStatus.revision)
```
Jiaweihu08, line 224: Use `revisionToUse` instead of the existing revision in case we want to allow changing indexing columns

core/src/main/scala/io/qbeast/spark/index/OTreeDataAnalyzer.scala
```diff
@@ -216,6 +217,21 @@ object DoublePassOTreeDataAnalyzer
       indexStatus: IndexStatus,
       options: QbeastOptions): (DataFrame, TableChanges) = {
     logTrace(s"Begin: Analyzing the input data with existing revision: ${indexStatus.revision}")
+
+    // Check if the DataFrame is deterministic
+    logDebug(s"Checking the determinism of the input data")
```
Jiaweihu08, line 222: Create a separate method for this check?

Jiaweihu08 (changes requested):

core/src/main/scala/io/qbeast/spark/index/OTreeDataAnalyzer.scala
osopardo1: already created in `analyzeDataFrameDeterminism`

core/src/main/scala/io/qbeast/spark/index/OTreeDataAnalyzer.scala
osopardo1: This part is a bit challenging. While I could use `revisionToUse`, it would defeat the purpose of avoiding the computation of min/max when we already know the source is non-deterministic.

Alternatively, it would be ideal to allow users to provide their own `columnStats` and enforce their use instead of relying on DataFrame analysis. This approach would give users more control and help bypass unnecessary computations when the source characteristics are already known.

What is your take with the current code status?

core/src/main/scala/io/qbeast/spark/index/OTreeDataAnalyzer.scala
osopardo1: Now I see this comment, somehow related with the above. 

> Use revisionToUse instead of the existing revision in case we want to allow changing indexing columns

In addition to the reasoning, I think we can have a conflict there. 

Imagine the user inputs its own `columnStats`, and the source is non-deterministic (for `LinearTransformation`, as you mentioned). The DF analysis might output a min/max outside of `columnStats` boundaries. If that's the case, should we detect it and throw an error? Or should we proceed with indexing?

src/test/scala/io/qbeast/spark/utils/QbeastInputSourcesTest.scala
```diff
@@ -0,0 +1,81 @@
+package io.qbeast.spark.utils
+
+import io.qbeast.QbeastIntegrationTestSpec
+
+class QbeastInputSourcesTest extends QbeastIntegrationTestSpec {
```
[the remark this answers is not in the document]
osopardo1: done in [d351eb3](https://github.com/Qbeast-io/qbeast-spark/pull/512/commits/d351eb377a1fd48b0eda0195d18e513e6fe6b7b0)

osopardo1:
After some discussion, we agreed to:

- Throw a Warning instead of an Error when indexing Non-Deterministic Columns or Queries. This Warning should advise the user that the query is non-deterministic, and that she/he can expect a failure during the process (if the data falls outside the Revision boundaries). It should also contain the two possible solutions:
    - Add columnStats
    - Materialize the data in a previous step 
- Group the Transformers that might be impacted by the non-determinism (in this case, just the LinearTransformer) and only check those. 
- Error control on LinearTrasnformation.transform -> values outside [0.0,1.0) would cause the indexing process to stop. The error should also contain the two possible solutions: `columnStats` and Materialization.

core/src/main/scala/io/qbeast/spark/index/RowUtils.scala
```diff
@@ -44,7 +44,17 @@ object RowUtils {
...
+      val transformation = t.transform(v)
+      assert(
+        transformation >= 0.0 && transformation <= 1.0,
+        s"Value $v is out of bounds of the Transformation. This can happen if you are indexing a Dataframe that:" +
+          s"1. It's source is constantly changing." +
+          s"2. The query that produces the DataFrame is not deterministic." +
+          s"To avoid this problem, we suggest to: " +
+          s"1. Change the transformer type to quantiles." +
+          s"2. Add columnStats with a greater space range to avoid indexing errors." +
+          s"3. save the DF as delta and Convert it To Qbeast in a second step")
```
Jiaweihu08, line 56: "3. Materialize the input data before indexing." *

core/src/main/scala/io/qbeast/spark/index/SparkPlanAnalyzer.scala
```diff
@@ -0,0 +1,120 @@
...
+  /**
+   * Extracts the select expressions from the logical plan for the provided column name
+   *
+   * @param logicalPlan
+   *   the logical plan
+   * @param columnName
+   *   the column name
+   * @return
+   */
+  private[index] def collectSelectExpressions(
```
Jiaweihu08, line 59: Does this catch the case where an indexing column `b` is built using a non-deterministic column `a` and `a` is not indexed?

core/src/main/scala/io/qbeast/core/transform/Transformer.scala
```diff
@@ -156,6 +156,12 @@ trait Transformer extends Serializable {
 
   def spec: String = s"$columnName:${transformerType.transformerSimpleName}"
 
+  /**
+   * Returns if the transformation should have deterministic values at indexing time
+   * @return
+   */
+  def shouldBeDeterministic: Boolean = false
```
Jiaweihu08, line 163: We don't really need this, do we?

core/src/main/scala/io/qbeast/spark/index/OTreeDataAnalyzer.scala
```diff
@@ -225,6 +227,28 @@ object DoublePassOTreeDataAnalyzer
     }
     logDebug(s"revisionToUse=$revisionToUse")
 
+    // Check if the DataFrame and the Columns To Index are deterministic
```
Jiaweihu08, line 230: Create a method for `DoublePassOTreeDataAnalyzer` that handles all these checks.

core/src/main/scala/io/qbeast/spark/index/SparkPlanAnalyzer.scala
```diff
@@ -0,0 +1,120 @@
...
+   *
+   * @param dataFrame
+   *   the DataFrame to analyze
+   * @param columnsToAnalyze
+   *   the columns to analyze for determinism
+   * @return
+   */
+  def analyzeDataFrameDeterminism(
+      dataFrame: DataFrame,
+      columnsToAnalyze: Seq[String]): Boolean = {
```
Jiaweihu08, line 101: Instead of using `columnsToAnalyze: Seq[String],` pass the `Reivison` object. Filter the `Transformers` that have to be deterministic and check the determinism of their columns. `areColumnsToAnalyzeDeterministic` should be `true` if there's no such `Transformer` in the `Revision.`

We probably don't need to create `shouldBeDeterministic` for the Transformer trait for now.

Use something like 
```scala
revision.columnTransformers.filter {
  case t: LinearTransformer => true
  case _ => false
}
  .exist(isColumnDeterministic(_.columnName))
```

src/test/scala/io/qbeast/spark/index/SparkPlanAnalyzerTest.scala
```diff
@@ -0,0 +1,99 @@
...
+
+  object SparkPlanAnalyzerTesting extends SparkPlanAnalyzer
+
+  private lazy val nonDeterministicUDF = org.apache.spark.sql.functions
+    .udf(() => {
+      scala.util.Random.nextInt()
+    })
+    .asNondeterministic()
+
+  "SparkPlanAnalyzer" should "detect underterminism in LIMIT" in withSparkAndTmpDir {
```
Jiaweihu08, line 15: "indeterminism"?

Jiaweihu08: Let's use a single word for this description: non-deterministic, e.g., non-deterministic query plan, non-deterministic column

Jiaweihu08 (changes requested):

core/src/main/scala/io/qbeast/spark/index/SparkPlanAnalyzer.scala
osopardo1: I was thinking about this, but I wanted to narrow the scope of the class to avoid including anything from Qbeast. See, this SparkPlanAnalyzer should only be in charge of Analyzing the SparkPlan with the internals of Spark.

osopardo1: On the shouldBeDeterministic, I think it's a better approach to push the acknowledge of determinism to the Transformer itself. 
Otherwise, each time you build a new Transformer, you must remember each time to add it (or not) to this other part of the code.

src/test/scala/io/qbeast/spark/index/SparkPlanAnalyzerTest.scala
osopardo1: true

core/src/main/scala/io/qbeast/spark/index/OTreeDataAnalyzer.scala
osopardo1: sure

core/src/main/scala/io/qbeast/spark/index/SparkPlanAnalyzer.scala
Jiaweihu08: Consider renaming it to something like "isBounded" to better reflect the nature of the default Transformation it will create, i.e., will all values be mapped to `[0, 1]` regardless of the min/max or other parameters used to create the Transformation? Using `shouldBeDeterministic` ties the property too much to this issue.

osopardo1: Mm, I like it. It can work for other constraints in the future.

osopardo1: Have a silly question, though. What is bounded? the Transformer or the Transformation?

core/src/main/scala/io/qbeast/spark/index/SparkPlanAnalyzer.scala
osopardo1: It should, but let me double chect it

core/src/main/scala/io/qbeast/spark/index/SparkPlanAnalyzer.scala
Jiaweihu08: They can be in the Transformations because we can access them after computing the revision changes.

As for how to define this property, the thing that concerns us is whether the `transform` method can or cannot output values outside the range [0, 1]. Then, a `Transformation` is bounded only when `transform` as a function has a domain of [0,1] at max.

osopardo1:
UPDATE:

We weren't confident enough with the solution, and we recently found out that Delta Lake has a [similar issue when doing Merge INTO](https://docs.delta.io/latest/delta-update.html#upsert-into-a-table-using-merge).

They solved it by materializing the data involved in the operation before executing the second pass. We can introduce that change in a second step. I would advocate to:

1. Add the logic to detect Undeterministic Sources.
2. Add the logic to materialize the data under demand. (This includes configuration of the Storage Level, enforcement..)

osopardo1:
Closing it due to inactivity. I would update the documentation to proceed in case of error.

